PR #5, opened 2020-06-10
Match related objects between AWS account and operator crs

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,40 @@ osdctl account check-secrets
 osdctl account check-secrets <Account CR Name>
 ```
 
+### Match AWS Account with AWS Account Operator related resources
+
+1. Get AWS Account Operator related resources
+
+```bash
+# Get Account Name by AWS Account ID, output to json
+osdctl account get account -i <Account ID> -o json
+
+# Get Account Claim CR by AWS Account ID, output to yaml
+osdctl account get account-claim -i <Account ID> -o yaml
+
+# Get Legal Entity information by AWS Account ID
+osdctl account get legal-entity -i <Account ID>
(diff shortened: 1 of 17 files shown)


clcollins:
@yeya24 This is looking really good.  I love they way this works, and it'll save a lot of time for us!  The JSON and YAML output of the full object is fantastic, too.

Regarding:

```
osd-utils-cli get account -i <Account ID>
osd-utils-cli get account-claim <Account ID>
osd-utils-cli get aws-account -a <Account CR Name>
```

Can you be more consistent about where flags and positional arguments are used?  Pick one and do it that way, just for easy of use for the rest of the team.

I'm excited for this feature! We do this kind of stuff ALL THE TIME, so it will be a huge timesaver!

yeya24:
> @yeya24 This is looking really good. I love they way this works, and it'll save a lot of time for us! The JSON and YAML output of the full object is fantastic, too.
> 
> Regarding:
> 
> ```
> osd-utils-cli get account -i <Account ID>
> osd-utils-cli get account-claim <Account ID>
> osd-utils-cli get aws-account -a <Account CR Name>
> ```
> 
> Can you be more consistent about where flags and positional arguments are used? Pick one and do it that way, just for easy of use for the rest of the team.
> 
> I'm excited for this feature! We do this kind of stuff ALL THE TIME, so it will be a huge timesaver!

Thanks. I updated the pr to use the flag way. You can use `-i` or `--account-id` to specify the account id.

rogbas:
@rogbas please build this version and validate it

PR #33, opened 2020-06-26
Use go generate instead of bespoke mockgen

2uasimojo:
This way you can just stuff more //go:generate directives where needed
rather than maintaining a bespoke script.


Makefile
```diff
@@ -43,7 +43,7 @@ docs: build
 	@git diff --exit-code -- ./docs/command/
 
 mockgen: ensure-mockgen
-	./hack/mockgen.sh
+	$(GO) generate ./...
```
yeya24, line 46: Would you minding add `${BUILDFLAGS}` after `go generate`? I think this might fix the CI error.

2uasimojo: Done.

Since we're here, I might suggest calling your build target something without `mock` in the name, since `go generate` can be used to generate other things. But I didn't want to change that here in case you have something else relying on the current name.

yeya24: Great suggestion. I think that makes sense. Maybe I can add in a later pr.

PR #42, opened 2020-07-02
Adding AWS cost management program

nickwyl:
There isn't a way to run the cost management program through the CLI yet. I'm storing the code here and will later implement something like `osd-ctl cost <domain> get <OUid>` to run the program.

- Nick Wang (nicwang)


nickwyl:
Hi @clcollins: I made some progress on the cost management program and I've placed it inside osd-utils-cli/cmd/cost instead. I'm going to make a PR for that in a bit and it'd be ready for review! I will remove the current PR. Thank you!

PR #43, opened 2020-07-07
Integrated cost-management program into osd-utils-cli

nickwyl:
- Added command to use cost-management program with `osdctl cost`
- Can manage AWS Cost Explorer with `osdctl cost aws`. This leaves the possibility of adding cost management support for other platforms e.g. `osdctl cost gcp`
- Get cost of OUs with `osdctl cost aws get`
- Create cost category with `osdctl cost aws create`
- Reconcile cost categories with `osdctl cost aws reconcile`

- files are located under `osd-utils-cli/cmd/cost`


.gitignore
```diff
@@ -17,3 +17,6 @@ bin/
 
 # IDE
 .idea
+
+# MacOS
+.DS_Store
```
yeya24, line 22: Could you please add a new line to all the files with the error `No newline at end of file`?

cmd/root.go
```diff
@@ -62,6 +63,9 @@ func NewCmdRoot(streams genericclioptions.IOStreams) *cobra.Command {
 	templates.ActsAsRootCommand(rootCmd, []string{"options"})
 	rootCmd.AddCommand(newCmdOptions(streams))
 
+	//Add cost command to use AWS Cost Manager
+	rootCmd.AddCommand(cost.CostCmd)
```
yeya24, line 67: Could you please make `cost.CostCmd` a function instead of a variable? Just like `cost.NewCmdCost(streams)`.

It would be great if you can add this to L58

cmd/cost/aws.go
```diff
@@ -0,0 +1,51 @@
...
+	},
+}
+
+func init() {
+	awsCmd.AddCommand(getCmd)
+	awsCmd.AddCommand(createCmd)
+	awsCmd.AddCommand(reconcileCmd)
+}
+
+func initAWSClients() (*organizations.Organizations, *costexplorer.CostExplorer) {
```
yeya24, line 35: There is already two functions to initialize AWS Clients https://github.com/openshift/osd-utils-cli/blob/master/pkg/provider/aws/client.go here. You can add costexplorer and organizations clients there and add more interfaces.

cmd/cost/create.go
```diff
@@ -0,0 +1,72 @@
...
+		createCostCategory(&OUid, &OU, org, ce)
+	},
+}
+
+func init() {
+	createCmd.Flags().String("ou", "", "get OU ID")
+	err := createCmd.MarkFlagRequired("ou")
+	if err != nil {
+		log.Fatalln("OU flag:", err)
+	}
```
yeya24, line 45: It would be great if you make createCmd a function and initialize all the flags at that function.

nickwyl:
Superseded by #55

PR #55, opened 2020-07-20
Integrated cost management tool into osdctl

nickwyl:
Added command to use cost-management program with osdctl cost

- Get cost of OUs with `osdctl cost get`

- Create cost category with `osdctl cost create`

- Reconcile cost categories with `osdctl cost reconcile`

- List cost categories with `osdctl cost list`

- files are located under `osd-utils-cli/cmd/cost`

### DIFF
--- a/cmd/cost/cost.go
+++ b/cmd/cost/cost.go
@@ -1,0 +1,94 @@
+package cost
+
+import (
+	"fmt"
+	"github.com/openshift/osd-utils-cli/cmd/common"
+	awsprovider "github.com/openshift/osd-utils-cli/pkg/provider/aws"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+// costCmd represents the cost command
+func NewCmdCost(streams genericclioptions.IOStreams) *cobra.Command {
+	opsCost = newCostOptions(streams)
+	costCmd := &cobra.Command{
+		Use:   "cost",
+		Short: "Cost Management related utilities",
+		Long: `The cost command allows for cost management on the AWS platform (other 
+platforms may be added in the future)`,
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(opsCost.complete(cmd, args))
+		},
+	}
+
+	//Set flags
+	costCmd.PersistentFlags().StringVarP(&opsCost.accessKeyID, "aws-access-key-id", "a", "", "AWS Access Key ID")
+	costCmd.PersistentFlags().StringVarP(&opsCost.secretAccessKey, "aws-secret-access-key", "x", "", "AWS Secret Access Key")
+	costCmd.PersistentFlags().StringVarP(&opsCost.profile, "aws-profile", "p", "", "specify AWS profile")
+	costCmd.PersistentFlags().StringVarP(&opsCost.configFile, "aws-config", "c", "", "specify AWS config file path")
+	costCmd.PersistentFlags().StringVarP(&opsCost.region, "aws-region", "g", common.DefaultRegion, "specify AWS region")
+
+	//Add commands
+	costCmd.AddCommand(newCmdGet(streams))
+	costCmd.AddCommand(newCmdReconcile(streams))
+	costCmd.AddCommand(newCmdCreate(streams))
+	costCmd.AddCommand(newCmdList(streams))
+
+	return costCmd
+}
+
+var opsCost *costOptions
+
+// costOptions defines the struct for running Cost command
+type costOptions struct {
+	// AWS config
+	accessKeyID     string
+	secretAccessKey string
+	configFile      string
+	profile         string
+	region          string
+
+	genericclioptions.IOStreams
+}
+
+func newCostOptions(streams genericclioptions.IOStreams) *costOptions {
+	return &costOptions{
+		IOStreams: streams,
+	}
+}
+
+//Initiate AWS clients for Organizations and Cost Explorer services using, if given, credentials in flags, else, credentials in the environment
+func (opsCost *costOptions) initAWSClients() (awsprovider.Client, error) {
+	//Initialize AWS clients
+	var (
+		awsClient awsprovider.Client
+		err       error
+	)
+	if opsCost.accessKeyID == "" && opsCost.secretAccessKey == "" {
+		awsClient, err = awsprovider.NewAwsClient(opsCost.profile, opsCost.region, opsCost.configFile)
+	} else {
+		awsClient, err = awsprovider.NewAwsClientWithInput(&awsprovider.AwsClientInput{
+			AccessKeyID:     opsCost.accessKeyID,
+			SecretAccessKey: opsCost.secretAccessKey,
+			Region:          opsCost.region,
+		})
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	return awsClient, err
+}
+
+//Initializes costOptions struct
+func (opsCost *costOptions) complete(cmd *cobra.Command, _ []string) error {
+	if opsCost.accessKeyID == "" && opsCost.secretAccessKey == "" {
+		fmt.Fprintln(opsCost.Out, "AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are not provided, reading credentials from config file or env vars.")
+	} else if opsCost.accessKeyID == "" || opsCost.secretAccessKey == "" {
+		return cmdutil.UsageErrorf(cmd, "The flag aws-access-key-id and aws-secret-access-key should be set or not set at the same time")
+	}
+
+	return nil
+}
--- a/cmd/cost/create.go
+++ b/cmd/cost/create.go
@@ -1,0 +1,85 @@
+package cost
+
+import (
+	"fmt"
+	awsprovider "github.com/openshift/osd-utils-cli/pkg/provider/aws"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/costexplorer"
+	"github.com/aws/aws-sdk-go/service/organizations"
+	"log"
+
+	"github.com/spf13/cobra"
+)
+
+// createCmd represents the create command
+func newCmdCreate(streams genericclioptions.IOStreams) *cobra.Command {
+	createCmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create a cost category for the given OU",
+		Run: func(cmd *cobra.Command, args []string) {
+
+			awsClient, err := opsCost.initAWSClients()
+			cmdutil.CheckErr(err)
+
+			//OU Flag
+			OUid, err := cmd.Flags().GetString("ou")
+			if err != nil {
+				log.Fatalln("OU flag:", err)
+			}
+
+			//Get Organizational Unit
+			OU := organizations.OrganizationalUnit{Id: aws.String(OUid)}
+
+			if err := createCostCategory(&OUid, &OU, awsClient); err != nil {
+				log.Fatalf("Error creating cost category for %s: %v", OUid, err)
+			}
+		},
+	}
+	createCmd.Flags().String("ou", "", "get OU ID")
+	if err := createCmd.MarkFlagRequired("ou"); err != nil {
+		log.Fatalln("OU flag:", err)
+	}
+
+	return createCmd
+}
+
+//Store flag options for create command
+type createOptions struct {
+	ou string
+	genericclioptions.IOStreams
+}
+
+//Create Cost Category for OU given as argument for -ou flag
+func createCostCategory(OUid *string, OU *organizations.OrganizationalUnit, awsClient awsprovider.Client) error {
+	//Gets all (not only immediate) accounts under the given OU
+	accounts, err := getAccountsRecursive(OU, awsClient)
+	if err != nil {
+		return err
+	}
+
+	_, err = awsClient.CreateCostCategoryDefinition(&costexplorer.CreateCostCategoryDefinitionInput{
+		Name:        OUid,
+		RuleVersion: aws.String("CostCategoryExpression.v1"),
+		Rules: []*costexplorer.CostCategoryRule{
+			{
+				Rule: &costexplorer.Expression{
+					Dimensions: &costexplorer.DimensionValues{
+						Key:    aws.String("LINKED_ACCOUNT"),
+						Values: accounts,
+					},
+				},
+				Value: OUid,
+			},
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Created Cost Category for", *OUid)
+
+	return nil
+}
--- a/cmd/cost/get.go
+++ b/cmd/cost/get.go
@@ -1,0 +1,270 @@
+package cost
+
+import (
+	"fmt"
+	awsprovider "github.com/openshift/osd-utils-cli/pkg/provider/aws"
+	"github.com/spf13/cobra"
+	"log"
+	"strconv"
+	"time"
+
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/costexplorer"
+	"github.com/aws/aws-sdk-go/service/organizations"
+)
+
+// getCmd represents the get command
+func newCmdGet(streams genericclioptions.IOStreams) *cobra.Command {
+	ops := newGetOptions(streams)
+	getCmd := &cobra.Command{
+		Use:   "get",
+		Short: "Get total cost of a given OU. If no OU given, then gets total cost of v4 OU.",
+		Run: func(cmd *cobra.Command, args []string) {
+
+			awsClient, err := opsCost.initAWSClients()
+			cmdutil.CheckErr(err)
+
+			//Get Organizational Unit
+			OU := organizations.OrganizationalUnit{Id: aws.String(ops.ou)}
+			//Store cost
+			var cost float64
+
+			if ops.recursive { //Get cost of given OU by aggregating costs of all (including immediate) accounts under OU
+				if err := getOUCostRecursive(&cost, &OU, awsClient, &ops.time); err != nil {
+					log.Fatalln("Error getting cost of OU recursively:", err)
+				}
+				fmt.Printf("Cost of %s recursively is: %f\n", ops.ou, cost)
+			} else { //Get cost of given OU by aggregating costs of only immediate accounts under given OU
+				if err := getOUCost(&cost, &OU, awsClient, &ops.time); err != nil {
+					log.Fatalln("Error getting cost of OU:", err)
+				}
+				fmt.Printf("Cost of %s is: %f\n", ops.ou, cost)
+			}
+		},
+	}
+	getCmd.Flags().BoolVarP(&ops.recursive, "recursive", "r", false, "recurse through OUs")
+	getCmd.Flags().StringVarP(&ops.time, "time", "t", "", "set time")
+	getCmd.Flags().StringVar(&ops.ou, "ou", "", "get OU ID")
+
+	if err := getCmd.MarkFlagRequired("ou"); err != nil {
+		log.Fatalln("OU flag:", err)
+	}
+
+	return getCmd
+}
+
+//Store flag options for get command
+type getOptions struct {
+	ou        string
+	recursive bool
+	time      string
+
+	genericclioptions.IOStreams
+}
+
+func newGetOptions(streams genericclioptions.IOStreams) *getOptions {
+	return &getOptions{
+		IOStreams: streams,
+	}
+}
+
+//Get account IDs of immediate accounts under given OU
+func getAccounts(OU *organizations.OrganizationalUnit, awsClient awsprovider.Client) ([]*string, error) {
+	var accountSlice []*string
+	var nextToken *string
+
+	//Populate accountSlice with accounts by looping until accounts.NextToken is null
+	for {
+		accounts, err := awsClient.ListAccountsForParent(&organizations.ListAccountsForParentInput{
+			ParentId:  OU.Id,
+			NextToken: nextToken,
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		for i := 0; i < len(accounts.Accounts); i++ {
+			accountSlice = append(accountSlice, accounts.Accounts[i].Id)
+		}
+
+		if accounts.NextToken == nil {
+			break
+		}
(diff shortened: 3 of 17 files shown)


nickwyl:
Supersedes #43

cmd/cost/get.go
```diff
@@ -0,0 +1,244 @@
...
+		if accounts.NextToken == nil {
+			break
+		}
+
+		//Get accounts
+		accounts, err = org.ListAccountsForParent(&organizations.ListAccountsForParentInput{
+			ParentId:  OU.Id,
+			NextToken: accounts.NextToken,
+		})
+	}
```
sam-nguyen7, line 97: good start here, but I think we can make this logic flow better. currently you have 2 ListAccountsForParents calls. Its possible to only do this once by putting the call in line 74 inside the for loop and set the first token to 0. Here is an example of pagination in another place https://github.com/openshift/aws-account-operator/blob/master/pkg/totalaccountwatcher/totalaccountwatcher.go#L108 PTAL and reach out if you need more help fixing this

cmd/cost/get.go
```diff
@@ -0,0 +1,244 @@
...
+	return append(accountsIDs, getAccounts(OU, org)...)
+}
+
+//Get immediate OUs (child nodes) directly under given OU
+func getOUs(OU *organizations.OrganizationalUnit, org awsprovider.OrganizationsClient) []*organizations.OrganizationalUnit {
+	//OUSlice stores OUs
+	var OUSlice []*organizations.OrganizationalUnit
+
+	//Get child OUs under parent OU
+	OUs, err := org.ListOrganizationalUnitsForParent(&organizations.ListOrganizationalUnitsForParentInput{
```
sam-nguyen7, line 124: same comment as above

cmd/cost/get.go
```diff
@@ -0,0 +1,244 @@
...
+		OUs = append(OUs, getOUsRecursive(currentOU, org)...)
+	}
+
+	return OUs
+}
+
+//Get cost of given account
+func getAccountCost(accountID *string, ce awsprovider.CostExplorerClient, timePtr *string, cost *float64) {
+
+	start := strconv.Itoa(time.Now().Year()-1) + time.Now().Format("-01-") + "01" //Starting from the 1st of the current month last year i.e. if today is 2020-06-29, then start date is 2019-06-01
```
sam-nguyen7, line 171: move comment to the above line

cmd/cost/get.go
```diff
@@ -0,0 +1,244 @@
...
+	end := time.Now().Format("2006-01-02")
+	granularity := "MONTHLY"
+	metrics := []string{
+		"NetUnblendedCost",
+	}
+
+	switch *timePtr {
```
sam-nguyen7, line 178: explain why this is needed

cmd/cost/list.go
```diff
@@ -0,0 +1,66 @@
...
+			listCostsUnderOU(&OU, org, ce, &time)
+		},
+	}
+	listCmd.Flags().String("cc", "ou-0wd6-aff5ji37", "get name of Cost Category (default is name of v4's OU)")
+	listCmd.Flags().StringP("time", "t", "", "set time")
+
+	return listCmd
+}
+
+func listCostsUnderOU(OU *organizations.OrganizationalUnit, org awsprovider.OrganizationsClient, ce awsprovider.CostExplorerClient, timePtr *string) {
```
sam-nguyen7, line 48: add comment for func

cmd/cost/reconcile.go
```diff
@@ -0,0 +1,76 @@
...
+func reconciliateCostCategories(OU *organizations.OrganizationalUnit, org awsprovider.OrganizationsClient, ce awsprovider.CostExplorerClient) {
+	costCategoryCreated := false
+
+	OUs := getOUsRecursive(OU, org)
+	costCategoriesSet := mapset.NewSet()
+
+	existingCostCategories, err := ce.ListCostCategoryDefinitions(&costexplorer.ListCostCategoryDefinitionsInput{})
+
+	//Populate costCategoriesSet with cost categories by looping until existingCostCategories.NextToken is null
+	for {
```
sam-nguyen7, line 47: same comment as other for loops

pkg/provider/aws/client.go
```diff
@@ -54,12 +56,29 @@ type Client interface {
 	AttachRolePolicy(*iam.AttachRolePolicyInput) (*iam.AttachRolePolicyOutput, error)
 	DetachRolePolicy(*iam.DetachRolePolicyInput) (*iam.DetachRolePolicyOutput, error)
 	ListAttachedRolePolicies(*iam.ListAttachedRolePoliciesInput) (*iam.ListAttachedRolePoliciesOutput, error)
+
+	// Organizations and Cost Explorer
+	GetOrg() *organizations.Organizations
+	GetCE() *costexplorer.CostExplorer
```
sam-nguyen7, line 62: lets separate organizations and cost explorer

cmd/cost/cost.go
```diff
@@ -0,0 +1,93 @@
...
+		log.Fatalln("Error getting AWS clients:", err)
+	}
+
+	return awsClient.GetOrg(), awsClient.GetCE(), err
+}
+
+func (opsCost *costOptions) complete(cmd *cobra.Command, _ []string) error {
+	if opsCost.accessKeyID == "" && opsCost.secretAccessKey == "" {
+		fmt.Fprintln(opsCost.Out, "AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are not provided, reading credentials from config file or env vars.")
+	} else if opsCost.accessKeyID == "" || opsCost.secretAccessKey == "" {
```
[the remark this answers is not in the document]
nickwyl: What's the reasoning behind using another if statement?

cmd/cost/get.go
```diff
@@ -0,0 +1,244 @@
...
+		Run: func(cmd *cobra.Command, args []string) {
+
+			cmdutil.CheckErr(opsCost.complete(cmd, args))
+			org, ce, err := opsCost.initAWSClients()
+			cmdutil.CheckErr(err)
+
+			//Get Organizational Unit
+			OU := organizations.OrganizationalUnit{Id: aws.String(ops.ou)}
+			//Store cost
+			var cost float64 = 0
```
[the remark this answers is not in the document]
nickwyl: I wanted to get a precise representation of spendings, so I used float64. If I used an int, since we're getting and returning the cost of each account under a given OU (could be up to thousands of accounts), I thought we might be rounding too much and losing precision.

cmd/cost/get.go
```diff
@@ -0,0 +1,244 @@
...
+
+			if ops.recursive {
+				getOUCostRecursive(&OU, org, ce, &ops.time, &cost)
+				fmt.Printf("Cost of %s recursively is: %f\n", ops.ou, cost)
+			} else {
```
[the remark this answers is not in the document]
nickwyl: Ah, I think I'm unsure about best practices for handling if/else! Would it be better if I just did a return, or?

cmd/cost/get.go
nickwyl: Why "NetUnblendedCost" or why metrics is needed? AWS API seems to require metrics

cmd/cost/cost.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func newCostOptions(streams genericclioptions.IOStreams) *costOptions {
+	return &costOptions{
+		IOStreams: streams,
+	}
+}
+
+//Initiate AWS clients for Organizations and Cost Explorer services using, if given, credentials in flags, else, credentials in the environment
+func (opsCost *costOptions) initAWSClients() (awsprovider.OrganizationsClient, awsprovider.CostExplorerClient, error) {
```
yeya24, line 62: Not necessary to have this because there is already https://github.com/openshift/osd-utils-cli/blob/master/pkg/provider/aws/client.go#L66 and https://github.com/openshift/osd-utils-cli/blob/master/pkg/provider/aws/client.go#L103 to init AWS Client.

If organization APIs and costExplorer APIs are added to the Client interface, then we can just use that interface to do all the API calls, don't need to have a separate orgClient and a ceClient.

cmd/cost/cost.go
```diff
@@ -0,0 +1,95 @@
+package cost
+
+import (
+	"fmt"
+	"github.com/openshift/osd-utils-cli/cmd/common"
+	awsprovider "github.com/openshift/osd-utils-cli/pkg/provider/aws"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"log"
```
yeya24, line 10: It would be good to group and sort all the imports. But this is not urgent, we can definitely do it in follow-up PRs and add linters to ensure all our imports are sorted.

cmd/cost/create.go
```diff
@@ -0,0 +1,80 @@
...
+	createCmd.Flags().String("ou", "", "get OU ID")
+	err := createCmd.MarkFlagRequired("ou")
+	if err != nil {
+		log.Fatalln("OU flag:", err)
+	}
+
+	return createCmd
+}
+
+//Store flag options for get command
```
yeya24, line 49: `Store flag options for create command`?

cmd/cost/cost.go
```diff
@@ -0,0 +1,95 @@
...
+
+// costCmd represents the cost command
+func NewCmdCost(streams genericclioptions.IOStreams) *cobra.Command {
+	opsCost = newCostOptions(streams)
+	costCmd := &cobra.Command{
+		Use:   "cost",
+		Short: "Cost Management related utilities",
+		Long: `The cost command allows for cost management on the AWS platform (other 
+platforms may be added in the future)`,
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
```
yeya24, line 21: You can add `cmdutil.CheckErr(opsCost.complete(cmd, args))` here since opsCost is a global variable.

cmd/cost/get.go
```diff
@@ -0,0 +1,233 @@
...
+			if ops.recursive { //Get cost of given OU by aggregating costs of all (including immediate) accounts under OU
+				getOUCostRecursive(&OU, org, ce, &ops.time, &cost)
+				fmt.Printf("Cost of %s recursively is: %f\n", ops.ou, cost)
+			} else { //Get cost of given OU by aggregating costs of only immediate accounts under given OU
+				getOUCost(&OU, org, ce, &ops.time, &cost)
+				fmt.Printf("Cost of %s is: %f\n", ops.ou, cost)
+			}
+		},
+	}
+	getCmd.Flags().StringVar(&ops.ou, "ou", "ou-0wd6-aff5ji37", "get OU ID (default is v4)") //Default OU is v4
```
yeya24, line 46: Is this default value sensitive `ou-0wd6-aff5ji37`? Is it okay to keep it here?

cmd/cost/get.go
```diff
@@ -0,0 +1,233 @@
...
+
+	//Populate OUs
+	OUs := getOUs(OU, org)
+
+	//Loop through all child OUs, get their costs, and store it to cost of current OU
+	for _, childOU := range OUs {
+		accountsIDs = append(accountsIDs, getAccountsRecursive(childOU, org)...)
+	}
+
+	//*accountsIDs = append(*accountsIDs, getAccounts(OU, org)...)
```
yeya24, line 108: Please remove this line

cmd/cost/cost.go
nickwyl: Thanks for all the suggestions, Ben! I’ve made changes accordingly. Also good eye for noticing the hard-coded OU. I’m gonna ask Aaron to see if it’s okay to have that

cmd/cost/get.go
nickwyl: @aweiteka can this information be made public?

cmd/cost/cost.go
fahlmant: I think `gofmt` should sort these

cmd/cost/cost.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+//Initiate AWS clients for Organizations and Cost Explorer services using, if given, credentials in flags, else, credentials in the environment
+func (opsCost *costOptions) initAWSClients() (awsprovider.Client, error) {
+	//Initialize AWS clients
+	var (
+		awsClient awsprovider.Client
+		err       error
+	)
+	if opsCost.accessKeyID == "" && opsCost.secretAccessKey == "" {
```
fahlmant, line 69: Is  there a way to make if the user provides profile, then its mutually exclusive with accessKeyId/SecretAccessKey? Then this could be simplified to 

if opsCost.profile != "" {

} else {

}

cmd/cost/create.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+//Create Cost Category for OU given as argument for -ou flag
+func createCostCategory(OUid *string, OU *organizations.OrganizationalUnit, awsClient awsprovider.Client) {
+	//Gets all (not only immediate) accounts under the given OU
+	accounts := getAccountsRecursive(OU, awsClient)
+
+	_, err := awsClient.CreateCostCategoryDefinition(&costexplorer.CreateCostCategoryDefinitionInput{
+		Name:        OUid,
+		RuleVersion: aws.String("CostCategoryExpression.v1"),
```
fahlmant, line 61: Just for completion, could you provide the documentation as to why this is being set?

cmd/cost/create.go
```diff
@@ -0,0 +1,79 @@
...
+						Key:    aws.String("LINKED_ACCOUNT"),
+						Values: accounts,
+					},
+				},
+				Value: OUid,
+			},
+		},
+	})
+	if err != nil {
+		log.Fatalf("Error creating cost category for %s: %v", *OUid, err)
```
fahlmant, line 75: Needs a return. Otherwise, the output would look like:
```
Error creating cost category for myou: some error
Created Cost Category for myou
```

cmd/cost/get.go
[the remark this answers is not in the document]
fahlmant: I think you can just use `var cost float64` and it will initialize to 0

cmd/cost/get.go
fahlmant: I would say remove it. In this case, I think there should be no default, but I'll defer to @aweiteka

cmd/cost/get.go
```diff
@@ -0,0 +1,230 @@
...
+	var nextToken *string
+
+	//Populate accountSlice with accounts by looping until accounts.NextToken is null
+	for {
+		accounts, err := awsClient.ListAccountsForParent(&organizations.ListAccountsForParentInput{
+			ParentId:  OU.Id,
+			NextToken: nextToken,
+		})
+		if err != nil {
+			log.Fatalln("Unable to retrieve accounts under OU:", err)
```
fahlmant, line 78: I think this needs a return. Otherwise line 81 may try to iterate over a nil pointer.

cmd/cost/get.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	return OUs
+}
+
+//Get cost of given account
+func getAccountCost(accountID *string, awsClient awsprovider.Client, timePtr *string, cost *float64) {
+	//Starting from the 1st of the current month last year i.e. if today is 2020-06-29, then start date is 2019-06-01
+	start := strconv.Itoa(time.Now().Year()-1) + time.Now().Format("-01-") + "01"
+	end := time.Now().Format("2006-01-02")
```
fahlmant, line 158: Why is this end date in 2006?

cmd/cost/get.go
```diff
@@ -0,0 +1,230 @@
...
+		monthCost, err := strconv.ParseFloat(*costs.ResultsByTime[month].Total["NetUnblendedCost"].Amount, 64)
+		if err != nil {
+			log.Fatalln("Unable to get cost:", err)
+		}
+		*cost += monthCost
+	}
+}
+
+//Get cost of given OU by aggregating costs of only immediate accounts under given OU
+func getOUCost(OU *organizations.OrganizationalUnit, awsClient awsprovider.Client, timePtr *string, cost *float64) {
```
fahlmant, line 208: For all of the `getXCost`, why not return a float64 instead of having to deal with a pointer?

cmd/cost/reconcile.go
```diff
@@ -0,0 +1,76 @@
...
+
+			reconciliateCostCategories(&OU, awsClient)
+		},
+	}
+
+	return reconcileCmd
+}
+
+//Checks if there's a cost category for every OU. If not, creates the missing cost category. This should be ran every 24 hours.
+func reconciliateCostCategories(OU *organizations.OrganizationalUnit, awsClient awsprovider.Client) {
```
fahlmant, line 37: `reconcileCostCategories`

cmd/cost/create.go
nickwyl: Hmm I think the output actually ends at the message in log.Fatalf(). For example, if I try to create an existing cost category, the output is:
```
Error creating cost category for ou-example: ValidationException: Failed to create Cost Category: Cost category name ou-example already exists
        status code: 400, request id: 999example
```

However, I will change createCostCategory() to return an err and let newCmdCreate() handle it. Does that work better? Also, would returning an err make unit testing easier? Are these correct thoughts?

cmd/cost/create.go
nickwyl: Yep, the AWS API makes RuleVersion required:
 `RuleVersion *string `type:"string" required:"true" enum:"CostCategoryRuleVersion"``

cmd/cost/get.go
nickwyl: For Go, the constants "2006" = Current Year, "01" = Current Month, and "02" = Current Day.

cmd/cost/create.go
fahlmant: I'm more curious what the Rule version specified here is:
`CostCategoryExpression.v1`

cmd/cost/create.go
fahlmant: ah duh. Missed the Fatalf.

fahlmant: But I do think these functions should be returning the error upward. The calling code can then decide how to handle the err

cmd/cost/get.go
nickwyl: My initial thought was that it's neater to use pointers, as it saves a line for returning 😅. Also probably overthinking this, but passing and copying a pointer as opposed to a float64 might be more efficient for older architectures.

Do you think I should return instead to make it more intuitive, or?

cmd/cost/get.go
fahlmant: Ah okay, I see its the format. Does AWS require this format?

cmd/cost/get.go
fahlmant: For me, it adds a bit of confusion. If I call `getCost` I expect a return of some number. If I was going to use a pointer (personally) I would have it as the first argument because that's the value I'm really interested in.

cmd/cost/create.go
nickwyl: Seems to be the only valid version!

cmd/cost/get.go
nickwyl: Removed. The --ou flag is now mandatory for the get, list, and reconcile commands.

cmd/cost/cost.go
nickwyl: I haven't found a way to do this yet

cmd/cost/create.go
nickwyl: Done! 

However, I’m doing a lot of `if err != nil { return err }` or `if err := someFunc(); err != nil { return nil }` e.g.
- https://github.com/nickwyl/osd-utils-cli/blob/a5e43f2305e6a16f0775dfc74a8aa85df3a0092b/cmd/cost/get.go#L246
- https://github.com/nickwyl/osd-utils-cli/blob/a5e43f2305e6a16f0775dfc74a8aa85df3a0092b/cmd/cost/list.go#L50.
I was wondering if many of them are redundant?

cmd/cost/get.go
nickwyl: I believe so. For example, the AWS documentation gives the format for dates as “2017-05-01” and “2017-01-01”.

cmd/cost/get.go
nickwyl: Thanks for the advice, I think that would be really helpful for everyone. I've changed the pointer to the first argument.

cmd/cost/get.go
fahlmant: Would you mind finding the documentation for the AWS requirement?

cmd/cost/cost.go
fahlmant: https://github.com/openshift/osd-utils-cli/blob/master/cmd/federatedrole/apply.go#L70

cmd/cost/get.go
nickwyl: Sorry, here's the link: https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_DateInterval.html

cmd/cost/create.go
fahlmant: Welcome to go :)

cmd/cost/get.go
fahlmant: :+1:

cmd/cost/cost.go
nickwyl: Ahh I think I misunderstood you. I'm looking at this further and I think there is a way. However, we cannot do 
```
if opsCost.profile != "" {

} else {

}
```
without refactoring the NewAwsClient and NewAwsClientWithInput functions, AFAIK. Since we need to call NewAwsClient in both cases where `opsCost.profile!=""` and `opsCost.profile == ""` (the latter occurs when all opsCost fields are null (with the exception of opsCost.region, which has a default value), in which we search for env vars), we can use your if/else format by moving implementing `sess := session.Must(session.NewSessionWithOptions(opt))` (https://github.com/openshift/osd-utils-cli/blob/72d26ea1f906232344c8ac6401a14bd904a5f447/pkg/provider/aws/client.go#L83) somewhere in NewAwsClientWithInput (https://github.com/openshift/osd-utils-cli/blob/72d26ea1f906232344c8ac6401a14bd904a5f447/pkg/provider/aws/client.go#L103).

I might need to change some of Ben's code as well if we want to do that. What are your thoughts?

cmd/cost/create.go
nickwyl: hahahaha

cmd/cost/cost.go
fahlmant: I see what your saying. What happens if a user doesn't provide access keys OR a profile?

nickwyl: If that happens, the first if statement would execute to check for env vars. Namely, `NewAwsClient` runs, then `sess := session.Must(session.NewSessionWithOptions(opt))` runs. NewSessionWithOptions(opt) first loads the env vars (could be nil), then--if profile was given--sets the session's profile. NewSessionWithOptions exits and we're back at NewAwsClient. In our scenario, we'll then hit a "NoCredentialProviders" switch statement and return an error message. Program terminates with the message, `Error initializing AWS clients: Could not create AWS session: NoCredentialProviders: no valid providers in chain. Deprecated.`

rogbas:
@rogbas @drpaneas @drewandersonnz please take a look and add feedback

PR #58, opened 2020-08-07
Output OU Name instead of OU ID

nickwyl:
Outputs OU name instead of OU ID. For example, running `osdctl cost list —ou <OUid>` gives:

```
Cost of <OU Name>: <Spending>

Cost of child OUs:
Cost of <Child OU Name>: <Spending>
Cost of <Child OU Name>: <Spending>
Cost of <Child OU Name>: <Spending>
…
```

Whereas previously the command gives:

```
Cost of <OU ID>: <Spending>

Cost of child OUs:
Cost of <Child OU ID>: <Spending>
Cost of <Child OU ID>: <Spending>
Cost of <Child OU ID>: <Spending>
…
```


nickwyl:
Superseded by #59

PR #59, opened 2020-08-07
Output OU Name instead of OU ID

nickwyl:
Outputs OU name instead of OU ID. For example, running `osdctl cost list —ou <OUid>` gives:
```
Cost of <OU Name>: <Spending>

Cost of child OUs:
Cost of <Child OU Name>: <Spending>
Cost of <Child OU Name>: <Spending>
Cost of <Child OU Name>: <Spending>
…
```

Whereas previously the command gives:
```
Cost of <OU ID>: <Spending>

Cost of child OUs:
Cost of <Child OU ID>: <Spending>
Cost of <Child OU ID>: <Spending>
Cost of <Child OU ID>: <Spending>
…
```

### DIFF
--- a/cmd/cost/cost.go
+++ b/cmd/cost/cost.go
@@ -2,11 +2,14 @@ package cost
 
 import (
 	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/organizations"
 	"github.com/openshift/osd-utils-cli/cmd/common"
 	awsprovider "github.com/openshift/osd-utils-cli/pkg/provider/aws"
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"log"
 )
 
 // costCmd represents the cost command
@@ -58,6 +61,16 @@ func newCostOptions(streams genericclioptions.IOStreams) *costOptions {
 	}
 }
 
+func (opsCost *costOptions) complete(cmd *cobra.Command, _ []string) error {
+	if opsCost.accessKeyID == "" && opsCost.secretAccessKey == "" {
+		fmt.Fprintln(opsCost.Out, "AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are not provided, reading credentials from config file or env vars.")
+	} else if opsCost.accessKeyID == "" || opsCost.secretAccessKey == "" {
+		return cmdutil.UsageErrorf(cmd, "The flag aws-access-key-id and aws-secret-access-key should be set or not set at the same time")
+	}
+
+	return nil
+}
+
 //Initiate AWS clients for Organizations and Cost Explorer services using, if given, credentials in flags, else, credentials in the environment
 func (opsCost *costOptions) initAWSClients() (awsprovider.Client, error) {
 	//Initialize AWS clients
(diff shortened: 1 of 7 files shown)


nickwyl:
Supersedes #58

cmd/cost/cost.go
```diff
@@ -82,13 +95,13 @@ func (opsCost *costOptions) initAWSClients() (awsprovider.Client, error) {
...
-//Initializes costOptions struct
-func (opsCost *costOptions) complete(cmd *cobra.Command, _ []string) error {
-	if opsCost.accessKeyID == "" && opsCost.secretAccessKey == "" {
-		fmt.Fprintln(opsCost.Out, "AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are not provided, reading credentials from config file or env vars.")
-	} else if opsCost.accessKeyID == "" || opsCost.secretAccessKey == "" {
-		return cmdutil.UsageErrorf(cmd, "The flag aws-access-key-id and aws-secret-access-key should be set or not set at the same time")
+func getOU(org awsprovider.Client, OUid string) *organizations.OrganizationalUnit {
+	result, err := org.DescribeOrganizationalUnit(&organizations.DescribeOrganizationalUnitInput{
+		OrganizationalUnitId: aws.String(OUid),
+	})
```
sam-nguyen7, line 101: what happens if `result` return empty? eg. given an OU that does not have any child OUs, does this hit an error or will it return an empty list?

nickwyl: The way I understand DescribeOrganizationalUnit is that it returns information regarding an OU. As long as the OUid given for the field corresponds to an existing OU, there won't be an error. If there doesn't exist an OU with the OUid, there will be an "InvalidInputException: You provided a value that does not match the required pattern" error.

If the OU doesn't have any child OUs, the information for the OU would still be returned to result. Later on, the different commands (get, reconcile) will not loop through the child OUs of the chosen OU, as there aren't any to loop through.

ArielLima:
@nickwyl @sam-nguyen7 Why are we outputting OU IDs vs names?

nickwyl:
@ArielLima Previously, it was only outputting OU IDs, which is hard to read. By outputting OU names, it's easier to understand which OU/cluster/legal identity we're dealing with.

cmd/cost/cost.go
fahlmant: Does this function call operate on child OUs? It seems to only return the name of the OU specified.

cmd/cost/create.go
```diff
@@ -30,10 +30,9 @@ func newCmdCreate(streams genericclioptions.IOStreams) *cobra.Command {
 				log.Fatalln("OU flag:", err)
 			}
 
-			//Get Organizational Unit
-			OU := organizations.OrganizationalUnit{Id: aws.String(OUid)}
+			OU := getOU(awsClient, OUid)
```
fahlmant, line 33: Can you add back the comments?

cmd/cost/cost.go
nickwyl: Sorry I wasn't clear--it does not operate on child OUs.

fahlmant: :+1: Yep, looks good to me then. @sam-nguyen7 I don't think we have to worry about child OUs here

cmd/cost/create.go
nickwyl: Done (:

PR #63, opened 2020-08-25
Add higher level secret command and move related commands into it

ArielLima:
This has no card and no priority. 
Since we had more than one command related to secrets I thought it made sense to move them into a secret specific command.


rogbas:
@rogbas check/try locally please

rogbas:
@ArielLima The cli is not returning an error in case of an invalid credential (it's only displaying an error message), and no output message is displayed if the credential is valid.

rogbas:
Closing the PR, since we opted for a cluster-level approach for osdctl

PR #64, opened 2020-09-09
check-secrets -> verify-secrets, add CCS validation

fahlmant:
The PR does the following:
1. Changes `osdctl account check-secrets` to `osdctl account verify-secrets`
2. Changes the default behaviour to expect an Account CR name
3. Adds the `-A` option to validate all Account CR's secrets
4. Adds the BYOC secret to be validated if the Account CR has BYOC as true

From the done criteria:
`output command necessary to reset invalid credentials (action to be added)` 
This is not done as it relies on upcoming changes to generate new secrets, and can be done as part of that PR instead.

### DIFF
--- a/cmd/account/verify-secrets.go
+++ b/cmd/account/verify-secrets.go
@@ -2,6 +2,7 @@ package account
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/service/sts"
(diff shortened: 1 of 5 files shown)


cmd/account/verify-secrets.go
```diff
@@ -19,52 +19,54 @@ import (
...
 			cmdutil.CheckErr(ops.run())
 		},
 	}
 
-	checkSecretsCmd.Flags().StringVar(&ops.accountNamespace, "account-namespace", common.AWSAccountNamespace,
+	verifySecretsCmd.Flags().StringVar(&ops.accountNamespace, "account-namespace", common.AWSAccountNamespace,
 		"The namespace to keep AWS accounts. The default value is aws-account-operator.")
-	checkSecretsCmd.Flags().BoolVarP(&ops.verbose, "verbose", "v", false, "Verbose output")
+	verifySecretsCmd.Flags().BoolVarP(&ops.verbose, "verbose", "v", false, "Verbose output")
+	verifySecretsCmd.Flags().BoolVarP(&ops.all, "all", "a", false, "Verify all Account CRs")
```
iamkirkbater, line 42: For UX, can we make `all` be `-A`?  `-a` is used by other commands to pass the account-name.

cmd/account/verify-secrets.go
```diff
@@ -149,15 +169,22 @@ func (o *checkSecretsOptions) run() error {
...
-			continue
+			if o.all {
+				continue
+			}
+			return err
 		}
 	}
 
+	fmt.Fprintf(o.IOStreams.Out, "Credentials valid for %s\n", o.accountName)
 	return nil
```
iamkirkbater, line 188: If we're checking for "all" should we return an err if _any_ of them errored?

iamkirkbater (commented):
Just a few small things.

cmd/account/verify-secrets.go
clcollins: `-A` matches the `oc` command in v4 as well.

cmd/account/verify-secrets.go
fahlmant: I'm not sure 100% what to return here, but I realize now that line 187 should be conditional on o.all or not

cmd/account/verify-secrets.go
fahlmant: Done

cmd/account/verify-secrets.go
rogbas: In case `o.all == true` and all creds are valid, could we output a message confirming it?

fahlmant: Done

PR #65, opened 2020-09-09
Adds ClusterID parameters to on-demand credential generation

iamkirkbater:
This PR adds a ClusterID parameter to on-demand credential generation for SRE UX so they don't need to know the inner workings of the AWS Account Operator


cmd/account/console.go
```diff
@@ -104,6 +117,16 @@ func (o *consoleOptions) run() error {
...
 	var accountID string
+	if o.clusterID != "" {
+		accountClaim, err := k8s.GetAccountClaimFromClusterID(ctx, o.kubeCli, o.clusterID)
+		if err != nil {
+			return err
+		}
+		if accountClaim == nil {
+			return fmt.Errorf("Could not find any accountClaims for cluster with ID: %s", o.clusterID)
+		}
+		o.accountName = accountClaim.Spec.AccountLink
```
rogbas, line 128: Not a major thing, but if the accounClaim has not been claimed yet, the AccountLink could be empty. Should we return an error here?

If we continue it will trigger a weird error below, because we'll end up getting the AWS Account ID (which is empty as well)

cmd/account/cli.go
```diff
@@ -104,6 +117,19 @@ func (o *cliOptions) run() error {
 
 	ctx := context.TODO()
 	var accountID string
+	if o.clusterID != "" {
+		accountClaim, err := k8s.GetAccountClaimFromClusterID(ctx, o.kubeCli, o.clusterID)
+		if err != nil {
+			return err
+		}
+		if accountClaim == nil {
```
fahlmant, line 125: Wouldn't this be covered by the err? `ErrNotFound` or something similar?

iamkirkbater: In this case, no, because GetAccountClaimFromClusterID has to use the k8s List function, and not Get.  While Get will throw an error of ErrNotFound or whatever it is, List will just return an empty list.

iamkirkbater:
/hold

Need to test with new build because of the go module missing.

PR #66, opened 2020-09-11
Split generate and rotate secrets

fahlmant:
Splits generate and rotate secrets into two commands

1. Rotate secrets will rotate all associated secrets for a given Account CR

2. Generate secrets will generate a single secret and print out the yaml for a given IAM user in a given AWS account.

The idea is rotate will be for a high level, one shot command to fix up an Account, where generate is more detailed and specific for more control.

### DIFF
--- a/cmd/account/generate-secret.go
+++ b/cmd/account/generate-secret.go
@@ -1,0 +1,316 @@
+package account
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/aws/aws-sdk-go/service/sts"
+	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
+	"github.com/spf13/cobra"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift/osd-utils-cli/cmd/common"
+	"github.com/openshift/osd-utils-cli/pkg/k8s"
+	awsprovider "github.com/openshift/osd-utils-cli/pkg/provider/aws"
+)
+
+// newCmdGenerateSecret implements the generate-secret command which generates an new set of IAM User credentials
+func newCmdGenerateSecret(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
+	ops := newGenerateSecretOptions(streams, flags)
+	generateSecretCmd := &cobra.Command{
+		Use:               "generate-secret <IAM User name>",
+		Short:             "Generate IAM credentials secret",
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(ops.complete(cmd, args))
+			cmdutil.CheckErr(ops.run())
+		},
+		Aliases: []string{"generate-secrets"},
+	}
+
+	generateSecretCmd.Flags().StringVarP(&ops.accountName, "account-name", "a", "", "AWS Account CR name")
+	generateSecretCmd.Flags().StringVar(&ops.accountNamespace, "account-namespace", common.AWSAccountNamespace,
+		"The namespace to keep AWS accounts. The default value is aws-account-operator.")
+	generateSecretCmd.Flags().StringVarP(&ops.accountID, "account-id", "i", "", "AWS Account ID")
+	generateSecretCmd.Flags().StringVarP(&ops.profile, "aws-profile", "p", "", "specify AWS profile")
+	generateSecretCmd.Flags().StringVar(&ops.secretName, "secret-name", "", "Specify name of the generated secret")
+	generateSecretCmd.Flags().StringVar(&ops.secretNamespace, "secret-namespace", "aws-account-operator", "Specify namespace of the generated secret")
+	generateSecretCmd.Flags().BoolVar(&ops.quiet, "quiet", false, "Suppress logged output")
+	generateSecretCmd.Flags().BoolVar(&ops.ccs, "ccs", false, "Only generate specific secret for osdCcsAdmin. Requires Account CR name")
+
+	return generateSecretCmd
+}
+
+// generateSecretOptions defines the struct for running generate-secret command
+type generateSecretOptions struct {
+	accountName      string
+	accountID        string
+	accountNamespace string
+	iamUsername      string
+
+	secretName      string
+	secretNamespace string
+	quiet           bool
+	ccs             bool
+	outputPath      string
+
+	// AWS config
+	region  string
+	profile string
+	cfgFile string
+
+	flags *genericclioptions.ConfigFlags
+	genericclioptions.IOStreams
+	kubeCli client.Client
+}
+
+func newGenerateSecretOptions(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *generateSecretOptions {
+	return &generateSecretOptions{
+		flags:     flags,
+		IOStreams: streams,
+	}
+}
+
+func (o *generateSecretOptions) complete(cmd *cobra.Command, args []string) error {
+
+	// Opinionated config for CCS rotation
+	if o.ccs {
+		// If no account CR name is provided via the -a option, try and get it via argument
+		if o.accountName == "" {
+			return cmdutil.UsageErrorf(cmd, "Account CR name argument is required")
+		}
+
+		return nil
+	}
+
+	if len(args) != 1 {
+		return cmdutil.UsageErrorf(cmd, "IAM User name argument is required")
+	}
+	o.iamUsername = args[0]
+
+	// account CR name and account ID cannot be empty at the same time
+	if o.accountName == "" && o.accountID == "" {
+		return cmdutil.UsageErrorf(cmd, "AWS account CR name and AWS account ID cannot be empty at the same time")
+	}
+
+	if o.accountName != "" && o.accountID != "" {
+		return cmdutil.UsageErrorf(cmd, "AWS account CR name and AWS account ID cannot be set at the same time")
+	}
+
+	// only initialize kubernetes client when account name is set
+	if o.accountName != "" {
+		var err error
+		o.kubeCli, err = k8s.NewClient(o.flags)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (o *generateSecretOptions) run() error {
+
+	if o.ccs {
+		err := o.generateCcsSecret()
+
+		return err
+	}
+
+	ctx := context.TODO()
+	var err error
+	awsSetupClient, err := awsprovider.NewAwsClient(o.profile, o.region, o.cfgFile)
+	if err != nil {
+		return err
+	}
+
+	// Get the accountID
+	var accountID string
+	if o.accountName != "" {
+		account, err := k8s.GetAWSAccount(ctx, o.kubeCli, o.accountNamespace, o.accountName)
+		if err != nil {
+			return err
+		}
+		if account.Spec.AwsAccountID != "" {
+			accountID = account.Spec.AwsAccountID
+		} else {
+			return fmt.Errorf("Account CR is missing AWS Account ID")
+		}
+	} else {
+		accountID = o.accountID
+	}
+
+	// Ensure creds are valid
+	callerIdentityOutput, err := awsSetupClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	if err != nil {
+		return err
+	}
+
+	// Assume
+	roleArn := aws.String(fmt.Sprintf("arn:aws:iam::%s:role/%s", accountID, awsv1alpha1.AccountOperatorIAMRole))
+	credentials, err := awsprovider.GetAssumeRoleCredentials(awsSetupClient, aws.Int64(900),
+		callerIdentityOutput.UserId, roleArn)
+	if err != nil {
+		return err
+	}
+
+	awsClient, err := awsprovider.NewAwsClientWithInput(&awsprovider.AwsClientInput{
+		AccessKeyID:     *credentials.AccessKeyId,
+		SecretAccessKey: *credentials.SecretAccessKey,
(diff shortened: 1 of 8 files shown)


cmd/account/rotate-secret.go
```diff
@@ -80,128 +57,171 @@ func newRotateSecretOptions(streams genericclioptions.IOStreams, flags *genericc
...
-	// account CR name and account ID cannot be empty at the same time
-	if o.accountName == "" && o.accountID == "" {
-		return cmdutil.UsageErrorf(cmd, "AWS account CR name and AWS account ID cannot be empty at the same time")
-	}
+	o.accountCRName = args[0]
 
-	if o.accountName != "" && o.accountID != "" {
-		return cmdutil.UsageErrorf(cmd, "AWS account CR name and AWS account ID cannot be set at the same time")
+	if o.profile == "" {
+		o.profile = "default"
```
sam-nguyen7, line 68: what happens if default profile doesn't exist?

cmd/account/rotate-secret.go
```diff
@@ -80,128 +57,171 @@ func newRotateSecretOptions(streams genericclioptions.IOStreams, flags *genericc
...
+			return err
+		}
+	}
+
 	awsClient, err := awsprovider.NewAwsClientWithInput(&awsprovider.AwsClientInput{
 		AccessKeyID:     *credentials.AccessKeyId,
 		SecretAccessKey: *credentials.SecretAccessKey,
 		SessionToken:    *credentials.SessionToken,
-		Region:          o.region,
+		Region:          "us-east-1",
```
sam-nguyen7, line 124: any reason why this is hard coded?

cmd/account/rotate-secret.go
```diff
@@ -80,128 +57,171 @@ func newRotateSecretOptions(streams genericclioptions.IOStreams, flags *genericc
...
+	awsSetupClient, err := awsprovider.NewAwsClient(o.profile, "", "")
 	if err != nil {
 		return err
 	}
 
-	roleArn := aws.String(fmt.Sprintf("arn:aws:iam::%s:role/%s", accountID, awsv1alpha1.AccountOperatorIAMRole))
-	credentials, err := awsprovider.GetAssumeRoleCredentials(awsSetupClient, aws.Int64(900),
-		callerIdentityOutput.UserId, roleArn)
+	// Ensure AWS calls are succesful with client
+	callerIdentityOutput, err := awsSetupClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
```
sam-nguyen7, line 102: guess the best way to check if aws calls are successful is to make an actual aws call :) this works, seems.. hacky though

cmd/account/rotate-secret.go
fahlmant: Then creating a client will throw an error. This is just following how the aws cli works.

cmd/account/rotate-secret.go
fahlmant: Because region doesn't matter for IAM operations. No need to have a user provide one.

cmd/account/rotate-secret.go
fahlmant: This call was already in this command. Why do you think its hacky?

cmd/account/generate-secret.go
```diff
@@ -0,0 +1,204 @@
...
+		return err
+	}
+
+	secret := k8s.NewAWSSecret(
+		o.secretName,
+		o.secretNamespace,
+		*newKey.AccessKey.AccessKeyId,
+		*newKey.AccessKey.SecretAccessKey,
+	)
+	if !o.printSecret {
```
rogbas, line 191: The name of that argument variable conveys the opposite behaviour, can we rename it to `o.quiet` or something equivalent?

cmd/account/generate-secret.go
```diff
@@ -0,0 +1,204 @@
...
+		flags:     flags,
+		IOStreams: streams,
+	}
+}
+
+func (o *generateSecretOptions) complete(cmd *cobra.Command, args []string) error {
+	if len(args) != 1 {
+		return cmdutil.UsageErrorf(cmd, "IAM User name argument is required")
+	}
+	o.iamUsername = args[0]
```
rogbas, line 82: Should we set a default value for the username, instead of getting it via cli parameter?

cmd/account/rotate-secret.go
rogbas: IMO get-caller-identity is a valid way of checking if the credentials are good.

cmd/account/rotate-secret.go
```diff
@@ -80,128 +57,171 @@ func newRotateSecretOptions(streams genericclioptions.IOStreams, flags *genericc
...
-			return fmt.Errorf("account ID %s does not match the input account ID %s",
-				*callerIdentityOutput.Account, o.accountID)
-		}
+	// Ensure that the credentials are for the target account
+	callerIdentityOutput, err = awsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	if err != nil {
+		return err
+	}
+
+	if *callerIdentityOutput.Account != accountID {
```
rogbas, line 136: In this cmd there is no input param for the AWS Account ID, probably we don't need to validate it?

cmd/account/rotate-secret.go
```diff
@@ -80,128 +57,171 @@ func newRotateSecretOptions(streams genericclioptions.IOStreams, flags *genericc
...
-			return err
+	// Update existing osdManagedAdmin secret
+	err = common.UpdateSecret(o.kubeCli, o.accountCRName+"-osdManagedAdminSRE-secret", common.AWSAccountNamespace, newOsdManagedAdminSRESecretData)
+	if err != nil {
+		return err
+	}
+
+	// Only update osdCcsAdmin credential if specified
+	if o.osdCcsAdmin {
+		if account.Spec.BYOC {
```
rogbas, line 203: In case the `--ccs` argument has been passed we should output a confirmation, indicating we've successfully updated the CCS creds

cmd/account/rotate-secret.go
```diff
@@ -34,38 +32,17 @@ func newCmdRotateSecret(streams genericclioptions.IOStreams, flags *genericcliop
...
-	rotateSecretCmd.Flags().StringVarP(&ops.accountName, "account-name", "a", "", "AWS Account CR name")
-	rotateSecretCmd.Flags().StringVarP(&ops.accountID, "account-id", "i", "", "AWS Account ID")
 	rotateSecretCmd.Flags().StringVarP(&ops.profile, "aws-profile", "p", "", "specify AWS profile")
-	rotateSecretCmd.Flags().StringVarP(&ops.cfgFile, "aws-config", "c", "", "specify AWS config file path")
-	rotateSecretCmd.Flags().StringVarP(&ops.region, "aws-region", "r", common.DefaultRegion, "Specify AWS region")
-	rotateSecretCmd.Flags().StringVar(&ops.secretName, "secret-name", "byoc", "Specify name of the generated secret")
-	rotateSecretCmd.Flags().StringVar(&ops.secretNamespace, "secret-namespace", "aws-account-operator", "Specify namespace of the generated secret")
-	rotateSecretCmd.Flags().BoolVar(&ops.printSecret, "print", true, "Print the generated secret")
-	rotateSecretCmd.Flags().StringVarP(&ops.outputPath, "output", "o", "", "Output path for secret yaml file")
+	rotateSecretCmd.Flags().BoolVar(&ops.osdCcsAdmin, "ccs", false, "Also rotates osdCcsAdmin credential. Use caution.")
```
rogbas, line 36: I suggest we rename that variable to `ops.updateCcsCreds` or something that describes the action

rogbas (commented):
@fahlmant  +1 to the overall workflow, left a couple of comments/questions

cmd/account/generate-secret.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/openshift/osd-utils-cli/pkg/k8s"
+	awsprovider "github.com/openshift/osd-utils-cli/pkg/provider/aws"
+)
+
+// newCmdGenerateSecret implements the generate-secret command which generates an new set of IAM User credentials
+func newCmdGenerateSecret(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
+	ops := newGenerateSecretOptions(streams, flags)
+	generateSecretCmd := &cobra.Command{
+		Use:               "generate-secret <IAM User name>",
+		Short:             "Generate IAM credentials secret",
```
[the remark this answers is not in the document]
fahlmant: Yep, still need to tackle this anyway

cmd/account/generate-secret.go
```diff
@@ -0,0 +1,204 @@
...
+		return err
+	}
+
+	var accountID string
+	if o.accountName != "" {
+		account, err := k8s.GetAWSAccount(ctx, o.kubeCli, o.accountNamespace, o.accountName)
+		if err != nil {
+			return err
+		}
+		accountID = account.Spec.AwsAccountID
```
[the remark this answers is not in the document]
fahlmant: Good point

cmd/account/generate-secret.go
```diff
@@ -0,0 +1,204 @@
...
+			cmdutil.CheckErr(ops.complete(cmd, args))
+			cmdutil.CheckErr(ops.run())
+		},
+	}
+
+	generateSecretCmd.Flags().StringVar(&ops.accountNamespace, "account-namespace", common.AWSAccountNamespace,
+		"The namespace to keep AWS accounts. The default value is aws-account-operator.")
+	generateSecretCmd.Flags().StringVarP(&ops.accountID, "account-id", "i", "", "AWS Account ID")
+	generateSecretCmd.Flags().StringVarP(&ops.profile, "aws-profile", "p", "", "specify AWS profile")
+	generateSecretCmd.Flags().StringVarP(&ops.region, "aws-region", "r", common.DefaultRegion, "Specify AWS region")
```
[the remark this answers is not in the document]
fahlmant: Yeah, still need to work on generate

cmd/account/rotate-secret.go
fahlmant: 👍

cmd/account/rotate-secret.go
fahlmant: 👍

cmd/account/rotate-secret.go
fahlmant: 👍

cmd/account/generate-secret.go
fahlmant: Since this is a more generic option, I think we should make it required. Otherwise, someone could accidentally delete and create an AccessKey for the default user without meaning to.

cmd/account/generate-secret.go
```diff
@@ -0,0 +1,204 @@
...
+		SecretAccessKey: *credentials.SecretAccessKey,
+		SessionToken:    *credentials.SessionToken,
+		Region:          o.region,
+	})
+	if err != nil {
+		return err
+	}
+
+	// if the account ID is specified, will
+	if o.accountID != "" {
```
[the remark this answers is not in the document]
fahlmant: I guess. This was already here when I started. I feel its redundant though. I'll remove it

cmd/account/rotate-secret.go
clcollins: I *doubt* this would bite anyone, but silently defaulting to the default account kinda creeps me out.  Should we not require the user to explicitly say which account they want to use?

iamkirkbater: I'm in the camp of silently defaulting to the default account set, just to align our user-experience with that of the aws cli, like Taylor mentioned.

sam-nguyen7 (commented):
looks good to me. Will test this locally when I get a chance

cmd/account/rotate-secret.go
rogbas: +1 to following the same behaviour as aws cli regarding the default profile

cmd/account/generate-secret.go
```diff
@@ -0,0 +1,319 @@
...
+func (o *generateSecretOptions) complete(cmd *cobra.Command, args []string) error {
+
+	// Opinionated config for CCS rotation
+	if o.ccs {
+		// If no account CR name is provided via the -a option, try and get it via argument
+		if o.accountName == "" {
+			if len(args) != 1 {
+				return cmdutil.UsageErrorf(cmd, "Account CR name argument is required")
+			}
+			o.accountName = args[0]
```
rogbas, line 92: I don't get why we populate `o.accountName` from `args[0]` here, and below on L101 we populate `o.iamUsername` from the same arg.

Shouldn't only the IAM Username be passed as a cli arg, while accountName should come via `-a account-name`?

cmd/account/generate-secret.go
```diff
@@ -0,0 +1,319 @@
...
+		Short:             "Generate IAM credentials secret",
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(ops.complete(cmd, args))
+			cmdutil.CheckErr(ops.run())
+		},
+		Aliases: []string{"generate-secrets"},
+	}
+
+	generateSecretCmd.Flags().StringVarP(&ops.accountName, "account-name", "a", "", "AWS Account CR name")
```
rogbas, line 40: I don't see this parameter being used anywhere in this command, am I missing something?

fahlmant: It's used on line 141.

cmd/account/generate-secret.go
```diff
@@ -0,0 +1,316 @@
...
+		return err
+	}
+
+	username := aws.String(o.iamUsername)
+	ok, err := awsprovider.CheckIAMUserExists(awsClient, username)
+	if err != nil {
+		return err
+	}
+
+	// the specified user not exist, create one
```
clcollins, line 182:
```suggestion
	// if the specified user does not exist, create one
```

cmd/common/helpers.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// UpdateSecret updates a specified k8s secret with the provided data
+func UpdateSecret(kubeClient client.Client, secretName string, secretNamespace string, secretBody map[string][]byte) error {
```
clcollins, line 12: For my own better understanding: does this need to be exported? Or is it done for some reason I'm not seeing?  Or does it even matter?

I guess because it's in /common/, even if it's not yet used outside that package?

fahlmant: I does need to be exported. It's being used outside of common: https://github.com/openshift/osd-utils-cli/pull/66/files#diff-2e4c77053adcf10316f9947fd2d309faR197

PR #67, opened 2020-09-14
move AWS code to Factory

drewandersonnz:
Remove duplicate code and move AWS access to a Factory to make accessing the object easier across multiple cobra commands.


cmd/account/servicequotas/cmd.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+// NewCmdServiceQuotas implements commands related to AWS service-quotas
+func NewCmdServiceQuotas(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
+	baseCmd := &cobra.Command{
+		Use:               "servicequotas",
```
rogbas, line 11: Cobra Command supports `Aliases`, consider adding an alias for servicequota (singular)

pkg/k8s/cloudfactory.go
```diff
@@ -0,0 +1,197 @@
...
+)
+
+// CloudFactoryOptions defines the struct for running list account command
+type CloudFactoryOptions struct {
+	AccountName      string
+	AccountID        string
+	AccountNamespace string
+	ClusterID        string
+
+	Awscloudfactory awsprovider.CloudFactoryOptions
```
iamkirkbater, line 31: I'm wondering if it would be less complicated to pull the aws stuff in separately inside each command, rather than do so inside this k8s factory?

Or if we did so here, the cloud factory would take a second argument of which cloud to provide defaults for?

so AttachCobraCliFlags would have `AttachCobraCLIFlags(cmd *cobra.Command, cloud string)` where `cloud` would be validated as one of `AWS`, or `GCP`, and then we populate that cloud providers' specific flags through that.

Otherwise, I think it would be more straightforward to do the following:

```go
k8sFactory.AttachCliOptions(cmd)
awsFactory.AttachCliOptions(cmd)
```

iamkirkbater (commented):
I really like this factory pattern here. Just a few comments.

cmd/account/console.go
```diff
@@ -36,62 +27,41 @@ func newCmdConsole(streams genericclioptions.IOStreams, flags *genericclioptions
...
 
-	consoleCmd.Flags().StringVar(&ops.accountNamespace, "account-namespace", common.AWSAccountNamespace,
-		"The namespace to keep AWS accounts. The default value is aws-account-operator.")
-	consoleCmd.Flags().StringVarP(&ops.accountName, "account-name", "a", "", "The AWS account cr we need to create AWS console URL for")
-	consoleCmd.Flags().StringVarP(&ops.accountID, "account-id", "i", "", "The AWS account ID we need to create AWS console URL for -- This argument will not work for CCS accounts")
-	consoleCmd.Flags().StringVarP(&ops.clusterID, "cluster-id", "C", "", "The Internal Cluster ID from Hive to create AWS console URL for")
-	consoleCmd.Flags().StringVarP(&ops.profile, "aws-profile", "p", "", "specify AWS profile")
-	consoleCmd.Flags().StringVarP(&ops.cfgFile, "aws-config", "c", "", "specify AWS config file path")
-	consoleCmd.Flags().StringVarP(&ops.region, "aws-region", "r", common.DefaultRegion, "specify AWS region")
-	consoleCmd.Flags().Int64VarP(&ops.consoleDuration, "duration", "d", 3600, "The duration of the console session. "+
```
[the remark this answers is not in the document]
fahlmant: :+1:  Not all options are applicable to all commands.

pkg/k8s/cloudfactory.go
fahlmant: We should make the k8s cloudfactory more flexible at some point. GCP account interaction will likely land here. We could go with this and then refactor later I suppose. At least something to think about

cmd/account/console.go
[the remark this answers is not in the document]
drewandersonnz: yes, the duration is required for STS tokens :-( I thought the same, but its needed everywhere

pkg/k8s/cloudfactory.go
drewandersonnz: The intention is that we shouldn't care which cloud we are using, we are asking for a cluster in most cases, which is getting its information from the k8s objects.

Ultimately I think we should add a layer above this which is `ocm`, so we ask `ocm` for cluster information and the tool works its way through the correct shards until it has what it needs. This may require some more careful design though.

cmd/account/console.go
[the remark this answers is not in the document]
drewandersonnz: I mean, everywhere where AWS STS is used. Which includes all STS tokens, and all console URLs. 

It would not be required for GCP however.

pkg/k8s/cloudfactory.go
fahlmant: Could we rename it `CluserCloudFactory`?

rogbas:
@drewandersonnz the PR check seems to be failing, could you please check?

drewandersonnz:
I have removed the Service Quotas section of this PR to focus on the Factory itself.
Intention is to get the factory merged then move on with the Service Quotas code.
I want to keep PRs as small and focused as possible.

pkg/k8s/cloudfactory.go
rogbas: IMO the CloudFactory name doesn't make much sense here, bottom line what we want to manage here are accounts, shouldn't we call it `AccountFactory`? That way it could handle GCP accounts in the future too.

drewandersonnz: After discussion with @rogbas we considered the name `ClusterResourceFactory`

cmd/account/cli.go
```diff
@@ -36,62 +27,41 @@ func newCmdCli(streams genericclioptions.IOStreams, flags *genericclioptions.Con
...
 
-	flags *genericclioptions.ConfigFlags
 	genericclioptions.IOStreams
-	kubeCli client.Client
 }
 
 func newCliOptions(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cliOptions {
 	return &cliOptions{
-		flags:     flags,
+		k8scloudfactory: k8spkg.ClusterResourceFactoryOptions{
```
rogbas, line 48: This parameter is still named `cloudfactory`, shouldn't it be also renamed to `clusterresourcefactory`?

cmd/account/cli.go
```diff
@@ -36,62 +27,41 @@ func newCmdCli(streams genericclioptions.IOStreams, flags *genericclioptions.Con
...
-	accountID        string
-	accountNamespace string
-	clusterID        string
-	cliDuration      int64
-
-	// AWS config
-	region  string
-	profile string
-	cfgFile string
+	k8scloudfactory k8spkg.ClusterResourceFactoryOptions
```
rogbas, line 39: This parameter is still named cloudfactory, shouldn't it be also renamed to clusterresourcefactory?

cmd/account/cli.go
```diff
@@ -36,62 +27,41 @@ func newCmdCli(streams genericclioptions.IOStreams, flags *genericclioptions.Con
...
 
 func (o *cliOptions) complete(cmd *cobra.Command) error {
-	// account CR name and account ID cannot be empty at the same time
-	if o.accountName == "" && o.accountID == "" && o.clusterID == "" {
-		return cmdutil.UsageErrorf(cmd, "AWS account CR name, AWS account ID and Cluster ID cannot be empty at the same time")
-	}
-
-	if !o.hasOnlyOneTarget() {
-		return cmdutil.UsageErrorf(cmd, "AWS account CR name, AWS account ID, or Cluster ID cannot be set at the same time")
+	k8svalid, err := o.k8scloudfactory.ValidateIdentifiers()
```
rogbas, line 56: same as above

cmd/account/cli.go
```diff
@@ -36,62 +27,41 @@ func newCmdCli(streams genericclioptions.IOStreams, flags *genericclioptions.Con
...
+		if err != nil {
+			return err
+		}
 	}
 
-	// only initialize kubernetes client when account name or cluster ID is set
-	if o.accountName != "" || o.clusterID != "" {
-		var err error
-		o.kubeCli, err = k8s.NewClient(o.flags)
+	awsvalid, err := o.k8scloudfactory.Awscloudfactory.ValidateIdentifiers()
```
rogbas, line 63: same

cmd/account/cli.go
```diff
@@ -100,118 +70,17 @@ func (o *cliOptions) complete(cmd *cobra.Command) error {
...
-		accountID = account.Spec.AwsAccountID
-		isBYOC = account.Spec.BYOC
-		acctSuffix = account.Labels["iamUserId"]
-	} else {
-		accountID = o.accountID
-		isBYOC = false
-	}
-
-	callerIdentityOutput, err := awsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	awsClient, err := o.k8scloudfactory.GetCloudProvider(o.verbose)
```
rogbas, line 74: :arrow_up:

cmd/account/cli.go
```diff
@@ -100,118 +70,17 @@ func (o *cliOptions) complete(cmd *cobra.Command) error {
...
-		awsClient, err = awsprovider.NewAwsClientWithInput(awsClientInput)
-		if err != nil {
-			klog.Error("Failed to assume jump-role for CCS")
-			return err
-		}
-	}
-	credentials, err := awsprovider.GetAssumeRoleCredentials(awsClient, &o.cliDuration,
-		callerIdentityOutput.UserId, aws.String(fmt.Sprintf("arn:aws:iam::%s:role/%s", accountID, roleName)))
+	credentials, err := awsprovider.GetAssumeRoleCredentials(awsClient, &o.k8scloudfactory.Awscloudfactory.ConsoleDuration,
+		o.k8scloudfactory.Awscloudfactory.CallerIdentity.UserId,
```
rogbas, line 80: same as above, missing renaming the property name

rogbas (commented):
I gave a final review and added comments, basically the only missing thing is renaming the k8scloudfactory  property of cliOptions struct to match the new name.

Other than that it LGTM, I've built the cli from the PR and confirmed existing functionality is working fine.

PR #68, opened 2020-09-14
Adds ability to generate console URL for CCS acct

iamkirkbater:
Fulfills OSD-4814

To Test:

* Pull and start https://github.com/openshift/aws-account-operator/pull/469
* `make create-ccs`
* `oc get account -n aws-account-operator` and then note the account CR name that you just created.
* Build this code
* `./bin/osdctl account console -n [account-cr-name]`
* Click the URL Given
* Verify the aws account ID is correct and that you're logged in with the BYOCAdminAccess-[account-cr-suffix] in the top right corner of the aws console

### DIFF
--- a/cmd/account/console.go
+++ b/cmd/account/console.go
@@ -3,11 +3,14 @@ package account
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/sts"
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/klog"
@@ -36,7 +39,7 @@ func newCmdConsole(streams genericclioptions.IOStreams, flags *genericclioptions
 	consoleCmd.Flags().StringVar(&ops.accountNamespace, "account-namespace", common.AWSAccountNamespace,
 		"The namespace to keep AWS accounts. The default value is aws-account-operator.")
 	consoleCmd.Flags().StringVarP(&ops.accountName, "account-name", "a", "", "The AWS account cr we need to create AWS console URL for")
-	consoleCmd.Flags().StringVarP(&ops.accountID, "account-id", "i", "", "The AWS account ID we need to create AWS console URL for")
+	consoleCmd.Flags().StringVarP(&ops.accountID, "account-id", "i", "", "The AWS account ID we need to create AWS console URL for -- This argument will not work for CCS accounts")
 	consoleCmd.Flags().StringVarP(&ops.clusterID, "cluster-id", "C", "", "The Internal Cluster ID from Hive to create AWS console URL for")
 	consoleCmd.Flags().StringVarP(&ops.profile, "aws-profile", "p", "", "specify AWS profile")
 	consoleCmd.Flags().StringVarP(&ops.cfgFile, "aws-config", "c", "", "specify AWS config file path")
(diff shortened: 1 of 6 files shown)


cmd/account/console.go
```diff
@@ -104,14 +105,22 @@ func (o *consoleOptions) run() error {
...
+	var acctSuffix string
 	if o.accountName != "" {
 		account, err := k8s.GetAWSAccount(ctx, o.kubeCli, o.accountNamespace, o.accountName)
 		if err != nil {
 			return err
 		}
 		accountID = account.Spec.AwsAccountID
+		isBYOC = account.Spec.BYOC
+		if isBYOC {
+			acctSuffixSlice := strings.Split(o.accountName, "osd-creds-mgmt-")
```
fahlmant, line 118: This should be as a label on the Account CR, and IMO it should be considered the source of truth

cmd/account/console.go
```diff
@@ -123,7 +132,42 @@ func (o *consoleOptions) run() error {
...
 	}
 
+	// If BYOC we need to role-chain to use the right creds.
+	// Use the OrgAccess Role by default, override if BYOC
 	roleName := awsv1alpha1.AccountOperatorIAMRole
+
+	// TODO: Come back to this and do a lookup for the account CR if the account ID is the only one set so we can do this too.
+	if isBYOC {
+		// TODO: Change this second param to 'awsv1alpha1.SREAccessRoleName' once that gets merged
+		roleArn := fmt.Sprintf("arn:aws:iam::%s:role/%s", *callerIdentityOutput.Account, "RH-SRE-CCS-Access")
```
fahlmant, line 142: For #66 I got this from the ConfigMap (since that is the source of truth from the a-a-o point of view)

cmd/account/console.go
iamkirkbater: Oooh, that's so much easier!  Thanks!

Something something monday.

cmd/account/console.go
iamkirkbater: Hmm, originally I didn't get it from the configmap because I was thinking that this might be run independently of any of the hives, but the more that I think about it the less I think that will ever be the case, considering we need to query the hive to get the account CR, right?  That makes it a bit easier.

cmd/account/console.go
```diff
@@ -123,7 +132,42 @@ func (o *consoleOptions) run() error {
...
+
+		// Build the role-name for Access:
+		if acctSuffix == "" {
+			klog.Error("Unexpected error parsing the account CR suffix")
+			return fmt.Errorf("Unexpected error parsing the account CR suffix.")
+		}
+		roleName = fmt.Sprintf("BYOCAdminAccess-%s", acctSuffix)
+		// Get STS Credentials
+		creds, err := awsprovider.GetAssumeRoleCredentials(awsClient, &o.consoleDuration, aws.String(fmt.Sprintf("SRE-Console-Access-%s", *callerIdentityOutput.UserId)), aws.String(roleArn))
```
fahlmant, line 151: The roleSessionName should include the IAM user name of the SREP user

cmd/account/console.go
fahlmant: Yeah, I think it should always query the a-a-o on hive since it needs the CR anyway. If we ever change the name of the role too, then we only need to change it in a-a-o

cmd/account/console.go
fahlmant: You can look at #66 for this as well :)

cmd/account/console.go
iamkirkbater: Fixed.

cmd/account/console.go
```diff
@@ -104,14 +107,19 @@ func (o *consoleOptions) run() error {
...
+	var isBYOC bool
+	var acctSuffix string
 	if o.accountName != "" {
 		account, err := k8s.GetAWSAccount(ctx, o.kubeCli, o.accountNamespace, o.accountName)
 		if err != nil {
 			return err
 		}
 		accountID = account.Spec.AwsAccountID
+		isBYOC = account.Spec.BYOC
+		acctSuffix = account.Labels["iamUserId"]
```
fahlmant, line 119: It would be good to make sure this isn't blank, but its a nice to have, not required.

iamkirkbater: Because of the scope of the `account` var here I am just setting the variable here and I'm checking later: https://github.com/openshift/osd-utils-cli/pull/68/files/e9bb1f87dfdd93402bad49af075503753dd604b8#diff-0749f070a0fba513d2ec280163a6951aR157-R161

rogbas:
@rogbas review and test in stage and comment. In stage look for CCS clusters created after Sept 16

rogbas:
PR tested/validated in CCS stage clusters.

/lgtm

PR #71, opened 2020-10-10
Fix Makefile

cblecker:
Fixing a bunch of the boilerplate Makefile stuff to be less complex. This also fixes compilation on darwin, because go is smart enough to always use the current GOOS/GOARCH, so we don't need to hard code this to linux.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,22 +1,9 @@
-GOOS := $(if $(GOOS),$(GOOS),linux)
-GOARCH := $(if $(GOARCH),$(GOARCH),amd64)
-GO=CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) GO111MODULE=on go
-GOVERSION = $(shell $(GO) version | cut -c 14- | cut -d' ' -f1)
-
-REPOSITORY = github.com/openshift/osd-utils-cli
-
+REPOSITORY = $(shell go list -m)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -28,26 +16,25 @@ fmt:
...
+	go generate ${BUILDFLAGS} ./...
 	@git diff --exit-code -- ./pkg/provider/aws/mock
 
 ensure-mockgen:
-	GO111MODULE=off go get -u github.com/golang/mock/mockgen
+	go get github.com/golang/mock/mockgen@v1.4.4
 
 test:
-	$(GO) test ${BUILDFLAGS} ./... -cover $(PACKAGES)
+	$(GO) test ${BUILDFLAGS} ./... -covermode=atomic -coverpkg=./...
```
iamkirkbater, line 40: I'm getting the following error with this:

```> make test
test  ./... -covermode=atomic -coverpkg=./...
/bin/sh: line 0: test: -covermode=atomic: binary operator expected
make: *** [test] Error 2
```

iamkirkbater: Ahh, 

```suggestion
	go test ${BUILDFLAGS} ./... -covermode=atomic -coverpkg=./...
```

iamkirkbater (changes requested):
One small change, other than that it lgtm.

iamkirkbater (approved):

clcollins (commented):
/lgtm

PR #72, opened 2020-10-14
OSD4804: Add ServiceQuotas command routing with aliases

drewandersonnz:
ServiceQuotas: Add ServiceQuotas command routing with aliases
ServiceQuotas: Query for specific QuotaCode

Account CLI: Have different output formats including default, json, env

Factory: use target credentials from assumed role

### DIFF
--- a/cmd/account/cli.go
+++ b/cmd/account/cli.go
@@ -3,14 +3,12 @@ package account
 import (
 	"fmt"
 
-	"github.com/aws/aws-sdk-go/aws"
 	"github.com/spf13/cobra"
 
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 
 	k8spkg "github.com/openshift/osd-utils-cli/pkg/k8s"
-	awsprovider "github.com/openshift/osd-utils-cli/pkg/provider/aws"
 )
 
 // newCmdCli implements the Cli command which generates temporary STS cli credentials for the specified account cr
@@ -29,6 +27,7 @@ func newCmdCli(streams genericclioptions.IOStreams, flags *genericclioptions.Con
 
 	ops.k8sclusterresourcefactory.AttachCobraCliFlags(cliCmd)
 
+	cliCmd.Flags().StringVarP(&ops.output, "out", "o", "default", "Output format [default | json | env]")
 	cliCmd.Flags().BoolVarP(&ops.verbose, "verbose", "v", false, "Verbose output")
 
 	return cliCmd
@@ -38,6 +37,7 @@ func newCmdCli(streams genericclioptions.IOStreams, flags *genericclioptions.Con
 type cliOptions struct {
 	k8sclusterresourcefactory k8spkg.ClusterResourceFactoryOptions
 
+	output  string
 	verbose bool
 
 	genericclioptions.IOStreams
@@ -71,20 +71,29 @@ func (o *cliOptions) complete(cmd *cobra.Command) error {
 }
 
 func (o *cliOptions) run() error {
-	awsClient, err := o.k8sclusterresourcefactory.GetCloudProvider(o.verbose)
+	_, err := o.k8sclusterresourcefactory.GetCloudProvider(o.verbose)
 	if err != nil {
 		return err
 	}
 
-	credentials, err := awsprovider.GetAssumeRoleCredentials(awsClient, &o.k8sclusterresourcefactory.Awscloudfactory.ConsoleDuration,
-		o.k8sclusterresourcefactory.Awscloudfactory.CallerIdentity.UserId,
-		aws.String(fmt.Sprintf("arn:aws:iam::%s:role/%s",
-			o.k8sclusterresourcefactory.AccountID,
-			o.k8sclusterresourcefactory.Awscloudfactory.RoleName)))
-	if err != nil {
-		return err
+	creds := o.k8sclusterresourcefactory.Awscloudfactory.Credentials
+
+	if o.output == "default" {
+		fmt.Fprintf(o.IOStreams.Out, "Temporary AWS Credentials:\n%s\n", creds)
 	}
-	fmt.Fprintf(o.IOStreams.Out, "Temporary AWS Credentials:\n%s\n", credentials)
+
+	if o.output == "json" {
+		fmt.Fprintf(o.IOStreams.Out, "%s\n", creds)
+	}
+
+	if o.output == "env" {
+		fmt.Fprintf(o.IOStreams.Out, "AWS_ACCESS_KEY_ID=%s AWS_SECRET_ACCESS_KEY=%s AWS_SESSION_TOKEN=%s AWS_DEFAULT_REGION=%s",
+			*creds.AccessKeyId,
+			*creds.SecretAccessKey,
+			*creds.SessionToken,
+			o.k8sclusterresourcefactory.Awscloudfactory.Region,
+		)
+	}
 
 	return nil
 }
--- a/cmd/account/servicequotas/cmd.go
+++ b/cmd/account/servicequotas/cmd.go
@@ -1,0 +1,26 @@
+package servicequotas
+
+import (
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+// NewCmdServiceQuotas implements commands related to AWS service-quotas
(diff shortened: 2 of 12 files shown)


cmd/account/cli.go
```diff
@@ -71,20 +71,29 @@ func (o *cliOptions) complete(cmd *cobra.Command) error {
 }
 
 func (o *cliOptions) run() error {
-	awsClient, err := o.k8sclusterresourcefactory.GetCloudProvider(o.verbose)
+	_, err := o.k8sclusterresourcefactory.GetCloudProvider(o.verbose)
```
fahlmant, line 74: Is this call still needed if we aren't using its output?

cmd/account/cli.go
```diff
@@ -71,20 +71,29 @@ func (o *cliOptions) complete(cmd *cobra.Command) error {
...
 	}
 
-	credentials, err := awsprovider.GetAssumeRoleCredentials(awsClient, &o.k8sclusterresourcefactory.Awscloudfactory.ConsoleDuration,
-		o.k8sclusterresourcefactory.Awscloudfactory.CallerIdentity.UserId,
-		aws.String(fmt.Sprintf("arn:aws:iam::%s:role/%s",
-			o.k8sclusterresourcefactory.AccountID,
-			o.k8sclusterresourcefactory.Awscloudfactory.RoleName)))
-	if err != nil {
-		return err
+	creds := o.k8sclusterresourcefactory.Awscloudfactory.Credentials
```
fahlmant, line 79: ah okay I see now, the GetCloudProvider assumes the right role and places the credentials here. I think that GetCloudProvider has too much logic in it then if there are cases where we don't use the awsClient it provides. There can also be cases where we want the root level creds and not assume a role.

cmd/account/cli.go
```diff
@@ -71,20 +71,29 @@ func (o *cliOptions) complete(cmd *cobra.Command) error {
...
+
+	if o.output == "default" {
+		fmt.Fprintf(o.IOStreams.Out, "Temporary AWS Credentials:\n%s\n", creds)
+	}
+
+	if o.output == "json" {
```
fahlmant, line 85: json and default have the same output minus a few words of preamble. Can we consolidate these?

cmd/account/servicequotas/cmd.go
```diff
@@ -0,0 +1,26 @@
...
+
+// NewCmdServiceQuotas implements commands related to AWS service-quotas
+func NewCmdServiceQuotas(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
+	baseCmd := &cobra.Command{
+		Use:               "servicequotas",
+		Short:             "Interact with AWS service-quotas",
+		Args:              cobra.NoArgs,
+		DisableAutoGenTag: true,
+		Run:               help,
+		Aliases:           []string{"service-quotas", "service-quota"},
```
fahlmant, line 16: `servicequotas` might be good here too.

cmd/account/servicequotas/describe.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (o *describeOptions) run() error {
+	awsClient, err := o.k8sclusterresourcefactory.GetCloudProvider(o.verbose)
+	if err != nil {
+		return err
+	}
+
+	var serviceCode string = "ec2"
+	var searchMaxResults int64 = 100
```
fahlmant, line 85: Why have a specific max? Shouldn't this be paginated instead?

fahlmant (commented):
There's a lot of changes to the cloud factory and to the cli command that aren't related to the changes described in the title of the PR, nor the commit message. At a minimum I think the changes should be broken up by commit, but ideally in a separate PR.

cmd/account/cli.go
drewandersonnz: json == default without the pre-text. The pre-text was there before I started. Keeping that as `default` to maintain compatibility.

Personally, I wanted the `env` output for testing, the `json` is just an extra for almost zero cost.

I vote to leave it as is.

cmd/account/servicequotas/cmd.go
drewandersonnz: `servicequotas` is the root command name. Should the alias also have the original name in the list?

cmd/account/servicequotas/describe.go
drewandersonnz: this is related to the underlying AWS calls, I would need to understand how the pagination works to continue. At the moment it seems to default to 20 results per page with 64 total results, this is just being lazy.

cmd/account/cli.go
drewandersonnz: Yes. To do the k8s calls to get the target account information and return the creds required.

cmd/account/cli.go
drewandersonnz: A fair point, is that necessary for this commit/change? Could that be a future enhancement?

drewandersonnz:
@fahlmant the changes to the factory are to support the ServiceQuotas and improve the factory. I had a misunderstanding of how the codebase worked when branching out to create the factory and it didn't behave as expected. This solves that, while supporting the ServiceQuotas call, and backporting to `account cli` to allow for faster manual testing.

fahlmant:
Okay. Please separate the cloudfactory changes into a separate commit at the least. CLI changes should be in their own commit as well

cmd/account/servicequotas/cmd.go
fahlmant: :facepalm: nah ignore me.

cmd/account/servicequotas/describe.go
fahlmant: The return from ListServiceQuotas give a NextToken
https://docs.aws.amazon.com/sdk-for-go/api/service/servicequotas/#ListServiceQuotasOutput

which can be used in the ListServieQuotasInput
https://docs.aws.amazon.com/sdk-for-go/api/service/servicequotas/#ListAWSDefaultServiceQuotasInput

This patter in used in the reuse code of the accountclaim controller in the aws-account-operator, and all over the place in the aws-account-shredder

cmd/account/cli.go
fahlmant: Keeping default to maintain compatibility with what?

cmd/account/servicequotas/describe.go
```diff
@@ -0,0 +1,106 @@
...
+	return nil
+}
+
+func (o *describeOptions) run() error {
+	awsClient, err := o.k8sclusterresourcefactory.GetCloudProvider(o.verbose)
+	if err != nil {
+		return err
+	}
+
+	var serviceCode string = "ec2"
```
rogbas, line 84: As a good practice I would define the serviceCode as a global variable in this file.

cmd/account/servicequotas/describe.go
rogbas: @drewandersonnz I would be ok with addressing the pagination on the 2nd PR of that card. For this initial PR this max results covers all the available ec2 quotas. 

I'm happy to have a quick sync to go over it this week.

cmd/account/servicequotas/describe.go
```diff
@@ -0,0 +1,106 @@
...
+	})
+	if err != nil {
+		return err
+	}
+
+	if o.queryQuotaCode == "" {
+		fmt.Println(servicequotas.Quotas)
+	} else {
+		for _, quota := range servicequotas.Quotas {
+			if *quota.QuotaCode == o.queryQuotaCode {
```
rogbas, line 99: We should be more verbose here when the specified quota is not found, outputting a msg for example.

cmd/account/servicequotas/describe.go
drewandersonnz: pagination done :-) thanks @fahlmant for the pointer to examples

cmd/account/servicequotas/describe.go
drewandersonnz: moved to CLI param instead, with default = ec2

cmd/account/servicequotas/describe.go
drewandersonnz: added error return when not found

iamkirkbater (commented):
LGTM so far :)

cmd/account/cli.go
iamkirkbater: OSD-5676 created to address the global output and cleaning it up across the entire application.

PR #74, opened 2020-12-01
OSD-4804 service-quota --all-regions

drewandersonnz:
Allow use of --all-regions to loop through supported regions and apply servicequota commands in sequence

### DIFF
--- a/cmd/account/servicequotas/common.go
+++ b/cmd/account/servicequotas/common.go
@@ -6,11 +6,18 @@ import (
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 )
 
-// GetSupportedRegions returns a []string of regions supported
+/*GetSupportedRegions returns a []string of regions supported
+return []string of any regions found
+	if allRegions: return all found regions, ignore filter
+	if !allRegions:
+		filter set: return 1 element if region matches
+		filter == "": return all found regions
+return error if no regions are returned
+*/
 func GetSupportedRegions(filter string, allRegions bool) ([]string, error) {
 	var results []string
 	for i := range awsv1alpha1.CoveredRegions {
-		if (filter != "") && (allRegions == false) {
+		if (filter != "") && !allRegions {
 			if filter == i {
 				results = append(results, i)
 				return results, nil
--- a/cmd/account/servicequotas/common.go
+++ b/cmd/account/servicequotas/common.go
@@ -1,0 +1,29 @@
+package servicequotas
+
+import (
+	"errors"
+
+	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
+)
+
+// GetSupportedRegions returns a []string of regions supported
+func GetSupportedRegions(filter string, allRegions bool) ([]string, error) {
+	var results []string
+	for i := range awsv1alpha1.CoveredRegions {
+		if (filter != "") && (allRegions == false) {
+			if (filter == i) {
+				results = append(results, i)
+				return results, nil
+			}
+			continue
+		}
+
+		results = append(results, i)
+	}
+
+	if (len(results) == 0) {
+		return results, errors.New("Cannot find Supported Region")
+	}
+
+	return results, nil
+}
--- a/cmd/account/servicequotas/common.go
+++ b/cmd/account/servicequotas/common.go
@@ -11,7 +11,7 @@ func GetSupportedRegions(filter string, allRegions bool) ([]string, error) {
 	var results []string
 	for i := range awsv1alpha1.CoveredRegions {
 		if (filter != "") && (allRegions == false) {
-			if (filter == i) {
+			if filter == i {
 				results = append(results, i)
 				return results, nil
 			}
@@ -21,7 +21,7 @@ func GetSupportedRegions(filter string, allRegions bool) ([]string, error) {
 		results = append(results, i)
 	}
 
-	if (len(results) == 0) {
+	if len(results) == 0 {
 		return results, errors.New("Cannot find Supported Region")
 	}
 
--- a/cmd/account/servicequotas/common.go
+++ b/cmd/account/servicequotas/common.go
@@ -6,14 +6,7 @@ import (
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 )
 
-/*GetSupportedRegions returns a []string of regions supported
-return []string of any regions found
-	if allRegions: return all found regions, ignore filter
-	if !allRegions:
-		filter set: return 1 element if region matches
-		filter == "": return all found regions
-return error if no regions are returned
-*/
+// GetSupportedRegions returns a []string of all supported regions if found and an error. Filtering for a single region is also supported.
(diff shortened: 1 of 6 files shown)


cmd/account/servicequotas/describe.go
```diff
@@ -63,24 +66,45 @@ func newDescribeOptions(streams genericclioptions.IOStreams, flags *genericcliop
 }
 
 func (o *describeOptions) complete(cmd *cobra.Command) error {
-	k8svalid, err := o.k8sclusterresourcefactory.ValidateIdentifiers()
+	k8svalid, err1 := o.k8sclusterresourcefactory.ValidateIdentifiers()
```
drpaneas, line 69: Optionally, since you are using `err1` only within this scope of `if` statement, I would suggest to keep it limited in that, instead of making it available to the whole function:

```go
if k8svalid, err := o.k8sclusterresourcefactory.ValidateIdentifiers(); !k8svalid && err != nil {
		return err
	}
```

cmd/account/servicequotas/describe.go
```diff
@@ -63,24 +66,45 @@ func newDescribeOptions(streams genericclioptions.IOStreams, flags *genericcliop
...
 		}
 	}
 
-	awsvalid, err := o.k8sclusterresourcefactory.Awscloudfactory.ValidateIdentifiers()
+	awsvalid, err2 := o.k8sclusterresourcefactory.Awscloudfactory.ValidateIdentifiers()
 	if !awsvalid {
-		if err != nil {
-			return err
+		if err2 != nil {
+			return err2
```
drpaneas, line 79: Same here:

```go
if awsvalid, err := o.k8sclusterresourcefactory.Awscloudfactory.ValidateIdentifiers(); !awsvalid && err != nil{
			return err
	}
```

cmd/account/servicequotas/describe.go
```diff
@@ -63,24 +66,45 @@ func newDescribeOptions(streams genericclioptions.IOStreams, flags *genericcliop
...
 		}
 	}
 
+	_, err3 := GetSupportedRegions(o.k8sclusterresourcefactory.Awscloudfactory.Region, o.allRegions)
+	if err3 != nil {
+		return err3
```
drpaneas, line 85: Same here:

```go
if _, err := GetSupportedRegions(o.k8sclusterresourcefactory.Awscloudfactory.Region, o.allRegions); err !=nil {
		return err
	}
```

cmd/account/servicequotas/update.go
```diff
@@ -64,24 +67,45 @@ func newUpdateOptions(streams genericclioptions.IOStreams, flags *genericcliopti
 }
 
 func (o *updateOptions) complete(cmd *cobra.Command) error {
-	k8svalid, err := o.k8sclusterresourcefactory.ValidateIdentifiers()
+	k8svalid, err1 := o.k8sclusterresourcefactory.ValidateIdentifiers()
```
drpaneas, line 70: Same pattern applies here with `err{1,2,3}`.

cmd/account/servicequotas/update.go
```diff
@@ -64,24 +67,45 @@ func newUpdateOptions(streams genericclioptions.IOStreams, flags *genericcliopti
...
 	if !k8svalid {
-		if err != nil {
-			return err
+		if err1 != nil {
+			return err1
 		}
 	}
 
-	awsvalid, err := o.k8sclusterresourcefactory.Awscloudfactory.ValidateIdentifiers()
+	awsvalid, err2 := o.k8sclusterresourcefactory.Awscloudfactory.ValidateIdentifiers()
```
drpaneas, line 77: Same pattern applies here with `err{1,2,3}`.

cmd/account/servicequotas/update.go
```diff
@@ -64,24 +67,45 @@ func newUpdateOptions(streams genericclioptions.IOStreams, flags *genericcliopti
...
 	if !awsvalid {
-		if err != nil {
-			return err
+		if err2 != nil {
+			return err2
 		}
 	}
 
+	_, err3 := GetSupportedRegions(o.k8sclusterresourcefactory.Awscloudfactory.Region, o.allRegions)
```
drpaneas, line 84: Same pattern applies here with `err{1,2,3}`.

drpaneas (changes requested):
Great work @drewandersonnz :D Thank you very much for your time working on this and excellent work on writing tests for for `common.go`. Please see my comments on the review, here's a summary:

- I think we need to provide a better comment for the `GetSupportRegions` as it's not clear to me how it behaves unless I read your tests.
- Some minor optional syntactical glitches
- Some function calls return values you don't check.
- There is no testing for `update.go` and `describe.go`. We can consider putting that in another PR, so we don't block this one, since there was no testing before either for those files. I really have a bad feeling, because I think I saw functions no used and some dead code. Oh well ...

I'm open to discussion for any of the points :)

cmd/account/servicequotas/common.go
```diff
@@ -0,0 +1,29 @@
+package servicequotas
+
+import (
+	"errors"
+
+	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
+)
+
+// GetSupportedRegions returns a []string of regions supported
```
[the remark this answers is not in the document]
drewandersonnz: I've updated the comment to add clarity :-) thanks!

drewandersonnz:
@drpaneas @rogbas this should be in a state for review and merge now, I have tidied this from @drpaneas 's previous comments.

drpaneas:
Thanks Drew I will check it soon

cmd/account/servicequotas/common.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+/*GetSupportedRegions returns a []string of regions supported
+return []string of any regions found
+	if allRegions: return all found regions, ignore filter
+	if !allRegions:
+		filter set: return 1 element if region matches
+		filter == "": return all found regions
+return error if no regions are returned
+*/
```
drpaneas, line 16:
```suggestion
```

drpaneas (commented):
Is there any particular reason we need the commented code?

cmd/account/servicequotas/common.go
drewandersonnz: Its a pseudo-code comment describing how the function behaves.
You did ask to understand what the function did.

drewandersonnz: Comment changed as requested.

drpaneas (approved):

cmd/account/servicequotas/common.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"errors"
+
+	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
+)
+
+// GetSupportedRegions returns a []string of all supported regions if found and an error. Filtering for a single region is also supported.
+func GetSupportedRegions(filter string, allRegions bool) ([]string, error) {
+	var results []string
+	for i := range awsv1alpha1.CoveredRegions {
```
rogbas, line 12: Nice, I like how it's consuming from the main AWS account operator code.

rogbas (approved):
/lgtm
/approve

PR #79, opened 2021-02-11
OSD-6004 - Allow generic var replacement for any usecase

drpaneas:
Right now we support only two parameters `osdctl servicelog post -t $template -c $CLUSTER -i $CASE_ID `. This PR is removing this logic of specific flags (e.g. one flag for the cluster, one flag for the caseid, etc) and it replaces it with any custom flag introduced by the user, via the `-p`.

e.g. 
```bash
servicelog post -t /path/to/template -p FOO=asdf -p BAR=qwert
# you need to have ${FOO} and ${BAR} placeholders in the template, otherwise it fails
```

### DIFF
--- a/cmd/post.go
+++ b/cmd/post.go
@@ -19,13 +19,13 @@ import (
 )
 
 var (
-	template                                  string
-	isURL                                     bool
-	HTMLBody                                  []byte
-	Message                                   servicelog.Message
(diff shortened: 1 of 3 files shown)


mmazur:
tested, works, lgtm, plz mrg, kthxbye

PR #80, opened 2021-03-12
Clean up cost output

aweiteka:
- remove noisy AWS credential reminders
- add help text for `-t` params
- clean up list and csv output to align with usage


iamkirkbater:
Looks like it's getting a diff on the docs output.

I think you need to run `make docs` and then commit the diff.

PR #81, opened 2021-04-22
georgettica/add list servicelogs [OSD-6723]

georgettica:


georgettica:
currently has an issue that it prints out the ClusterList and not the ServiceLog details. I am not sure why it printed one and not the other but if someone who is more farmilliar can help i'll be glad

jaybeeunix:
> currently has an issue that it prints out the ClusterList and not the ServiceLog details. I am not sure why it printed one and not the other but if someone who is more farmilliar can help i'll be glad

i found that when i used the [search function of the api](https://github.com/openshift/ocm-container/commit/f1f94ae26a94cc76fdaf13d8268ff572dabe6436#diff-16f159b41a02e544f7a542414178988191ae2f25df6e17a285e4828a08896a0e), it gave me a clusterlist instead of the cluster directly, even when there was only one cluster returned. maybe something similar is at play here?

georgettica:
@jaybeeunix maybe, I basically want to suppress the first api query and expose the second one.

the first prints as if it has been passed via `jq` and the second is discarded.

cmd/servicelog/common.go
```diff
@@ -0,0 +1,55 @@
...
+	return response
+}
+
+func check(response *sdk.Response, dir string) {
+	status := response.Status()
+
+	body := response.Bytes()
+
+	if status < 400 {
+		validateGoodResponse(body)
```
mrbarge, line 46: @georgettica - I had a look at the 'why is the cluster list being printed out' problem - it's happening as a result of this call. This function always performs a dump of the response body contents to `stdout` ([link to the code](https://github.com/openshift/osdctl/blob/4e873844ea7b6bda85b15f76d16b79e56a37ca39/cmd/post.go#L308)).

The `validateGoodResponse()` function could maybe be refactored to instead pass `nil` to the `dump.Pretty()` function if no output is desired.. maybe via a `debug` control parameter or something.

georgettica: Thank you!
I will update it accordingly :))

georgettica:
@rogbas as discussed, can you TAL on this?

cmd/servicelog/cmd.go
```diff
@@ -0,0 +1,23 @@
+package servicelog
+
+import (
+	"github.com/spf13/cobra"
+)
+
+//func NewCmdServiceLog(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
```
dastergon, line 7: is this meant for removal?

cmd/servicelog/cmd.go
```diff
@@ -0,0 +1,23 @@
...
+func NewCmdServiceLog() *cobra.Command {
+	// servicelogCmd represents the servicelog command
```
dastergon, line 9: this is straightforward in the code, so I don't think it's needed

cmd/servicelog/list.go
```diff
@@ -0,0 +1,125 @@
...
+		err = dump.Pretty(os.Stdout, response.Bytes())
+		if err != nil {
+			return err
+		}
+
+		return nil
+	},
+}
+
+var serviceLogListFlags = struct {
```
dastergon, line 56: since you have only one flag you could just have 
```
var allMessages bool
```
Once we get to support multiple flags, we can create a new struct

cmd/servicelog/common.go
dastergon: Was this updated? If so, make sure to resolve the comment

dastergon (changes requested):
Looks good, sent a few suggestions. I'd love to see some tests, but since this is a kind of urgent feature, we can work on them later.

cmd/servicelog/common.go
georgettica: I didn't do it and cleaned the use of `dump.Pretty()` so resolving and maybe in the future we can add more verbosity information

cmd/servicelog/list.go
georgettica: fixed

cmd/servicelog/cmd.go
georgettica: solved

cmd/servicelog/list.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		log.Fatalf("Can't parse API path '%s': %v\n", targetAPIPath, err)
+	}
+	var empty []string
+	arguments.ApplyParameterFlag(request, []string{fmt.Sprintf(`search=cluster_uuid = '%s'`, clusterId)})
+	arguments.ApplyHeaderFlag(request, empty)
+	return request
+}
+
+func Bcheck(response *sdk.Response, dir string) {
```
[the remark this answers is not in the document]
georgettica: fixed

dastergon (approved):
/lgtm

georgettica:
@dastergon who should I ping in the team with lgtm perms?
@clcollins ?

fahlmant (approved):
/lgtm

PR #82, opened 2021-05-03
[OSD-7154] add support for user impersonation

dastergon:
As we rolled out backplane, some actions require access escalation by impersonating `backplane-cluster-admin`, and `osdctl` needs to be updated to follow the same behavior when needed.

This PR adds support for the `--as` flag used in Kubernetes for [user impersonation](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation).

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -2,11 +2,12 @@ package cmd
 
 import (
 	"flag"
+	"os"
+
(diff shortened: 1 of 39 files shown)


rogbas:
Feature validated in stage.

/lgtm

PR #83, opened 2021-05-06
adjust all references to osd-utils-cli into osdctl

RiRa12621:


georgettica:
/approve

seems this will make reading the code a little easier

PR #84, opened 2021-05-07
[OSD-7214] Added account recycle functionality to osdctl command

dkeohane:
[OSD-7214](https://issues.redhat.com/browse/OSD-7214)
- Added new flag to account reset to reset Legal Entity ID as well (allows for account reuse for other legal entities)
- [Updated SOP documentation](https://github.com/openshift/ops-sop/pull/1320)

### DIFF
--- a/cmd/account/reset.go
+++ b/cmd/account/reset.go
@@ -39,8 +39,7 @@ func newCmdReset(streams genericclioptions.IOStreams, flags *genericclioptions.C
 	resetCmd.Flags().BoolVarP(&ops.skipCheck, "skip-check", "y", false,
 		"Skip the prompt check")
 	resetCmd.Flags().BoolVar(&ops.resetLegalEntity, "reset-legalentity", false,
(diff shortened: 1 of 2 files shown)


cmd/account/reset.go
```diff
@@ -120,19 +124,30 @@ func (o *resetOptions) run() error {
...
+	}
+
 	if err := o.kubeCli.Update(ctx, account, &client.UpdateOptions{}); err != nil {
 		return err
 	}
 
 	// reset fields in status
 	var mergePatch []byte
+
+	status := map[string]interface{}{
```
dastergon, line 138: I believe it would make the code cleaner if you could convert the map to an anonymous struct.

dkeohane: I'll leave it the way it currently is - thanks

dastergon (approved):
/lgtm

rogbas:
/lgtm
cc @lisa @iamkirkbater

PR #86, opened 2021-05-12
Adds goReleaser for release creation and binary builds

clcollins:
Adds [goReleaser](https://goreleaser.com/) support for creating binary builds and releases in a standardized way.

Maintainers need to install the goReleaser binary, and then generate a Github  token with the `repo` scope and save it in `~/.config/goreleaser/token`.  

Then builds can be created by running `make build`, and a release can be created from the most recent tag by running `make release`.

Binaries will be built, a release created in Github, and the binaries added to a *.tar.gz file with any README.md and LICENSE files.

An sha256sum.txt file is generated for each of the artifacts, and added to the release.

Signed-off-by: Christopher Collins <collins.christopher@gmail.com>

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,48 @@
+# This is an example .goreleaser.yml file with some sane defaults.
+# Make sure to check the documentation at http://goreleaser.com
+env_files:
+  github_token: ~/.config/goreleaser/token
+
+before:
+  hooks:
+    # You may remove this if you don't use go modules.
+    - go mod tidy
+    # you may remove this if you don't need go generate
+    # - go generate ./...
+
+builds:
+  - env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - darwin
+    goarch:
+      - amd64
+    ldflags:
+      - -s -w -X 'github.com/openshift/osdctl/cmd.GitCommit={{.ShortCommit}}'
+
+archives:
+  - replacements:
+      darwin: Darwin
+      linux: Linux
+      amd64: x86_64
+
+checksum:
+  name_template: 'sha256sum.txt'
+  algorithm: sha256
+
+snapshot:
+  name_template: "{{ .Tag }}-next"
+
+changelog:
+  sort: asc
+  filters:
+    exclude:
+      - '^docs:'
+      - '^test:'
+
+release:
+  github:
+    owner: "openshift"
+    name: "osdctl"
+  prerelease: auto
--- a/Makefile
+++ b/Makefile
@@ -1,21 +1,33 @@
-REPOSITORY = $(shell go list -m)
-GIT_COMMIT = $(shell git rev-parse --short HEAD)
-
 BUILDFLAGS ?=
-LDFLAGS = -ldflags="-X '${REPOSITORY}/cmd.GitCommit=${GIT_COMMIT}'"
+GORELEASER_BUILD_ARGS = "--rm-dist"
 unexport GOFLAGS
 
 all: format mod build test
 
-format: vet fmt mockgen docs
+format: vet fmt mockgen ci-build docs
 
 fmt:
 	@echo "gofmt"
 	@gofmt -w -s .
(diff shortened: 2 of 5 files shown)


.goreleaser.yml
```diff
@@ -0,0 +1,51 @@
...
+      - amd64
+    ldflags:
+      - -s -w -X 'github.com/openshift/osdctl/cmd.GitCommit={{.ShortCommit}}'
+
+archives:
+  - replacements:
+      darwin: Darwin
+      linux: Linux
+      windows: Windows
+      386: i386
```
drpaneas, line 30: since your `goarch` is `amd64` only then I guess you can drop `386: i386`.
also since `goos` is not using Windows, then you can also drop the `windows: Windows`.

Makefile
```diff
@@ -15,7 +11,10 @@ fmt:
 	@git diff --exit-code .
 
 build:
-	go build ${BUILDFLAGS} ${LDFLAGS} -o ./bin/osdctl main.go
+	goreleaser --snapshot --skip-publish --rm-dist
```
drpaneas, line 14: The `goreleaser --snapshot --skip-publish --rm-dist` is used for a dry-run of the release locally. Is this what you want to do or just build? 

Otherwise, if you want to just build, you can do `goreleaser build --rm-dist`

drpaneas (commented):
Since you are creating a `.goreleaser.yml` we need to make sure this is not going to break in the future as well. You can add a github workflow to validate its syntax for the future, like this:

```bash
touch  .github/workflows/validate_goreleaser.yml
```

And put this content:

```yaml
name: validate_goreleaser

on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - master

jobs:
  validate_goreleaser:
    runs-on: ubuntu-latest

    steps:
      - name: Checkout
        uses: actions/checkout@v2.3.3
        with:
          fetch-depth: 0

      - name: Set up Go
        uses: actions/setup-go@v2.1.3
        with:
          go-version: 1.15.2

      - name: Download GoReleaser
        run: |
          curl -sfL https://install.goreleaser.com/github.com/goreleaser/goreleaser.sh | sh
      - name: Check goreleaser validity
        run: |
          ./bin/goreleaser check
```


Also in case you would like to automatically run GoReleaser every time you push a new `tag`, you can do something like this:

```bash
touch .github/workflows/gorelease.yml
```

And put this content:

```yaml
name: goreleaser
on:
  push:
    tags:
      - "*"
jobs:
  goreleaser:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v2.3.3
        with:
          fetch-depth: 0

      - name: Set up Go
        uses: actions/setup-go@v2.1.3

      - name: Cache Go modules
        uses: actions/cache@v2.1.2
        with:
          path: ~/go/pkg/mod
          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
          restore-keys: |
            ${{ runner.os }}-go-

      - name: Run GoReleaser
        uses: goreleaser/goreleaser-action@v2.2.1
        with:
          version: latest
          args: release --rm-dist
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

Anyways, these ^^^ are just some optional suggestions from my side. As for the the PR itself I left some in-line comments.

Makefile
clcollins: Thanks, I somehow missed that!

.goreleaser.yml
clcollins: Done!

clcollins:
@drpaneas I added a modified version of the goreleaser check workflow you suggested (just changed the piped-to-sh bit to something more specific).

I think the automatic tag is an end-goal, but I want to make sure this is doing what we want before we pull out the breaks.  I think that's the way to go, though.

fahlmant (approved):
/lgtm

fahlmant (approved):
/lgtm

PR #90, opened 2021-05-13
Removes Github workflow for goReleaser check in favor of Prow

clcollins:
Removes the Github Actions workflow for validating goReleaser config
and moves it into the Prow build workflow instead.

Signed-off-by: Christopher Collins <collins.christopher@gmail.com>


clcollins:
This won't work this way

PR #92, opened 2021-05-17
Fix README for build destination

2uasimojo:
`make build` now seems to put binaries under `./dist/{arch-specific-subdir}/`.


fahlmant (approved):
/lgtm

PR #93, opened 2021-05-18
[OSD-5505] Add flag to open AWS console url upon generation

dkeohane:
[OSD-5505](https://issues.redhat.com/browse/OSD-5505)
- Added a flag to the `osdctl account console` command (`--launch`) which launches a browser window to the console url at the end of the command
- removed `check-secrets` documentation as it was changed to `verify-secrets` [here](https://github.com/openshift/osdctl/commit/62926a0b6657c763344bc1d13a8d91915af911c6)


dastergon:
@dkeohane can you rebase and force-push again with only your commit? It look like the branch is derived from an older version

README.md
```diff
@@ -129,14 +129,14 @@ osdctl account clean-velero-snapshots -p <profile name> -c <config file path>
 
 ### AWS Account IAM User Credentials validation
 
-`check-secrets` command checks the IAM User Secret associated with Account Accout CR.
+`verify-secrets` command verifies the IAM User Secret associated with Account Accout CR.
```
dastergon, line 132: This change makes sense, but it might break SOP and people might be used to the older version. Also, we will need a major version bump to `osdctl` after the rename. 

Unless it's not a hard requirement from a ticket, I would recommend leaving it as is and only keep the flag feature. 
Feel free to open a separate PR and discuss it with other folks too so we can move forward with the rename.

dastergon (changes requested):

dastergon (approved):
/lgtm

fahlmant:
Can you add to the pr description that check-secrets was removed as well?

dkeohane:
> Can you add to the pr description that check-secrets was removed as well?
Added! Sorry about that, had commented in my commit message, forgot to add it to the PR :)

fahlmant (approved):
/lgtm

PR #96, opened 2021-06-07
First draft of account-assign cmd

katherinelc321:
Added the new cmd and added some apis to the aws client
@iamkirkbater

### DIFF
--- a/cmd/account/account-assign.go
+++ b/cmd/account/account-assign.go
@@ -14,7 +14,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// assignCmd represents the assign command
+// Global variables
+var (
+	defaultBYOCEnv    = "ou-rs3h-i0v69q47"
+	defaultNonBYOCEnv = "ou-0wd6-z6tzkjek"
+	defaultRootId     = "r-0wd6"
+)
+
+// assignCmd assigns an aws account to user under osd-staging-2 by default unless osd-staging-1 is specified
 func newCmdAccountAssign(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
 	ops := newAccountAssignOptions(streams, flags)
 	accountAssignCmd := &cobra.Command{
@@ -70,21 +77,24 @@ func (o *accountAssignOptions) run() error {
 
 	var (
 		accountAssignID string
+		destinationOu   string
+		defaultPayer    = "osd-staging-2"
 	)
 
 	if o.username != "" && o.payerAccount == "" {
-		o.payerAccount = "osd-staging-2"
+		o.payerAccount = defaultPayer
 	}
 	//Instantiating aws client
-	AwsClient, err := awsprovider.NewAwsClient(o.payerAccount, "us-east-1", "")
+	awsClient, err := awsprovider.NewAwsClient(o.payerAccount, "us-east-1", "")
 	if err != nil {
 		return err
 	}
 	//Listing accounts
 	input := &organizations.ListAccountsInput{}
-	accounts, err := AwsClient.ListAccounts(input)
+	accounts, err := awsClient.ListAccounts(input)
 	if err != nil {
 		fmt.Println(err.Error())
+		return err
 	}
 
 	//Lopping through the list to find an unassigned account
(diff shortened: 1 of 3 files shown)


cmd/account/account-assign.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/openshift/osdctl/pkg/printer"
+	awsprovider "github.com/openshift/osdctl/pkg/provider/aws"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// assignCmd represents the assign command
+func newCmdAccountAssign(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
```
dastergon, line 18: Make sure to add `accountCmd.AddCommand(newCmdAccountAssign(streams, flags))` in account/cmd.go  to assign it in the `account` sub-command.

cmd/account/account-assign.go
```diff
@@ -0,0 +1,147 @@
...
+	return &accountAssignOptions{
+		flags:      flags,
+		printFlags: printer.NewPrintFlags(),
+		IOStreams:  streams,
+	}
+}
+
+func (o *accountAssignOptions) complete(cmd *cobra.Command, _ []string) error {
+	if o.username == "" {
+		return cmdutil.UsageErrorf(cmd, "RedHat username was not provided")
```
dastergon, line 57:
```suggestion
		return cmdutil.UsageErrorf(cmd, "Red Hat username was not provided")
```

cmd/account/account-assign.go
```diff
@@ -0,0 +1,147 @@
...
+	return nil
+}
+
+func (o *accountAssignOptions) run() error {
+
+	var (
+		accountAssignID string
+	)
+
+	if o.username != "" && o.payerAccount == "" {
```
dastergon, line 75: You could create a default Payer Account variable and if it's empty just add the user-declared.

cmd/account/account-assign.go
```diff
@@ -0,0 +1,147 @@
...
+		o.payerAccount = "osd-staging-2"
+	}
+	//Instantiating aws client
+	AwsClient, err := awsprovider.NewAwsClient(o.payerAccount, "us-east-1", "")
```
dastergon, line 79: since `AwsClient` is used only within this context, you can rename it to `awsClient`

cmd/account/account-assign.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return err
+	}
+	//Listing accounts
+	input := &organizations.ListAccountsInput{}
+	accounts, err := AwsClient.ListAccounts(input)
+	if err != nil {
+		fmt.Println(err.Error())
```
dastergon, line 87: should we continue if there is an error? if not, then we should return the err here.

cmd/account/account-assign.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	_, err = AwsClient.TagResource(inputTag)
+	if err != nil {
+		return err
+	}
+
+	//Moving to developers OU
+	inputMove1 := &organizations.MoveAccountInput{
+
```
dastergon, line 119:
```suggestion
```

cmd/account/account-assign.go
```diff
@@ -0,0 +1,147 @@
...
+		DestinationParentId: aws.String("ou-0wd6-z6tzkjek"),
+		SourceParentId:      aws.String("r-0wd6"),
+	}
+	inputMove2 := &organizations.MoveAccountInput{
+		AccountId:           aws.String(accountAssignID),
+		DestinationParentId: aws.String("ou-rs3h-i0v69q47"),
+		SourceParentId:      aws.String("r-0wd6"),
+	}
+
+	if o.payerAccount == "osd-staging-2" {
```
dastergon, line 130: using global vars to map our environments will help with compatibility in the future in case they change. 

You could name them:
```
defaultNonBYOCEnv for "osd-staging-1"
defaultBYOCEnv for  "osd-staging-2"
```
cc: @iamkirkbater

cmd/account/account-assign.go
```diff
@@ -0,0 +1,147 @@
...
+		_, err = AwsClient.MoveAccount(inputMove2)
+		if err != nil {
+			return err
+		}
+	} else if o.payerAccount == "osd-staging-1" {
+		_, err = AwsClient.MoveAccount(inputMove1)
+		if err != nil {
+			fmt.Println(err.Error())
```
dastergon, line 138: should we continue if there is an error? if not, then we should return the err here.

dastergon (changes requested):

PR #98, opened 2021-06-09
AssignCmd ready to review

katherinelc321:
The Cmd lists all the accounts in root and that are not in any OU. It tags one with the username and moves it into the developers OU.


cmd/account/list/account.go
```diff
@@ -111,11 +114,18 @@ func (o *listAccountOptions) run() error {
...
 		matched         bool
 		reused          bool
 		claimed         bool
 		err             error
 	)
+
+	if o.accountID != "" {
+		inputListTag := &organizations.ListTagsForResourceInput{
+			ResourceId: aws.String(o.accountID)}
+		_, err = aws
```
dastergon, line 127: Could you elaborate on this line? as far as I can see it's not used anywhere and the error is not checked.

dastergon (changes requested):

cmd/account/list/account.go
katherinelc321: Oh i shouldn't have committed that. It's from another ticket that i started working on

cmd/account/account-assign.go
```diff
@@ -0,0 +1,163 @@
...
+func newAccountAssignOptions(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *accountAssignOptions {
+	return &accountAssignOptions{
+		flags:      flags,
+		printFlags: printer.NewPrintFlags(),
+		IOStreams:  streams,
+	}
+}
+
+func (o *accountAssignOptions) complete(cmd *cobra.Command, _ []string) error {
+	if o.username == "" {
```
dastergon, line 64: I see that the username is a requirement, but there's no flag to capture that from the CLI.

dastergon (changes requested):
One more note, I see that in some code paths you print the err.Error() before returning and in some, you skip them. I would recommend either print them in all returns to keep consistency.

cmd/account/account-assign.go
```diff
@@ -141,13 +143,15 @@ func (o *accountAssignOptions) run() error {
 
 	if o.payerAccount == defaultPayer {
 		destinationOu = defaultBYOCEnv
```
dastergon, line 145: I see that you first create`inputMove` and set its fields`DestinationParentId` to `aws.String(destinationOu)`, and `SourceParentId` to  `aws.String(rootId)`. Then, you check for the different conditions, and you update `destinationOu` but you don't update `inputMove`, so when you pass `inputMove` again to `MoveAccount` it will see the outdated values that you have set initially.

cmd/account/account-assign.go
```diff
@@ -141,13 +143,15 @@ func (o *accountAssignOptions) run() error {
...
+		rootId = profileRootId2
 		_, err = awsClient.MoveAccount(inputMove)
 		if err != nil {
 			fmt.Println(err.Error())
 			return err
 		}
 	} else if o.payerAccount == nonDefaultPayer {
 		destinationOu = defaultNonBYOCEnv
+		rootId = profileRootId1
 		_, err = awsClient.MoveAccount(inputMove)
```
dastergon, line 155: No need to execute `MoveAccount()` twice. You can update `inputMove` and just run `MoveAccount()` outside of the if statement.

dastergon (changes requested):

cmd/account/account-assign.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+	} else if o.payerAccount == nonDefaultPayer {
+		destinationOu = defaultNonBYOCEnv
+		rootId = profileRootId1
+		_, err = awsClient.MoveAccount(inputMove)
+		if err != nil {
+			fmt.Println(err.Error())
+			return err
+		}
+	}
```
iamkirkbater, line 160: I'm not sure if this is going to work correctly, although it might because of the pointers, but it's not very clear.  It might be better to do something like:

```
inputMove := organizations.MoveAccountInput

if default payer account {
  fill in inputMove with defaultPayeraccount stuff
} else {
  fill in inputMove with non-default stuff
}

_, err = awsClient.MoveAccount(inputMove)
if err != nil {
  fmt.Println(err.Error())
  return err
}
return nil

**Review by iamkirkbater — CHANGES_REQUESTED** [review]
Overall this looks really great!  A few minor comments.  Let's schedule some time next week to go over some unit tests and we can add tests to this.

File: cmd/account/account-assign.go [hunk 6]
```diff
@@ -0,0 +1,150 @@
...
+	ops.printFlags.AddFlags(accountAssignCmd)
+	accountAssignCmd.Flags().StringVarP(&ops.output, "output", "o", "", "Output format. One of: json|yaml|jsonpath=...|jsonpath-file=... see jsonpath template [http://kubernetes.io/docs/user-guide/jsonpath].")
+	accountAssignCmd.Flags().StringVarP(&ops.payerAccount, "payer-account", "p", "", "Payer account type")
+	accountAssignCmd.Flags().StringVarP(&ops.username, "username", "u", "", "LDAP username")
+
+	return accountAssignCmd
+}
+
+type accountAssignOptions struct {
```
dastergon, line 45: you can move `newAccountAssignOptions` and `accountAssignOptions`before `newCmdAccountAssign` to match the flow of declaration

cmd/account/account-assign.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func (o *accountAssignOptions) complete(cmd *cobra.Command, _ []string) error {
+	if o.username == "" {
+		return cmdutil.UsageErrorf(cmd, "LDAP username was not provided")
+	}
+	if o.payerAccount == "" {
+		return cmdutil.UsageErrorf(cmd, "Payer account was not provided")
+	}
+	var err error
```
dastergon, line 71:
```suggestion

	var err error
```

cmd/account/account-assign.go
```diff
@@ -0,0 +1,150 @@
...
+	o.kubeCli, err = k8s.NewClient(o.flags)
+	if err != nil {
+		fmt.Println(err.Error())
+		return err
+	}
+	return nil
```
dastergon, line 77:
```suggestion

	return nil
```

cmd/account/account-assign.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	//List accounts that are not in any OU
+	input := &organizations.ListAccountsInput{}
+	accounts, err := awsClient.ListAccounts(input)
+	if err != nil {
+		fmt.Println(err.Error())
+		return err
+	}
+	if len(accounts.Accounts) == 0 {
+		return fmt.Errorf("No accounts available to assign\n")
```
dastergon, line 109:
```suggestion
		return fmt.Errorf("No accounts available to assign")
```

cmd/account/account-assign.go
```diff
@@ -0,0 +1,150 @@
...
+	inputMove := &organizations.MoveAccountInput{
+		AccountId:           aws.String(accountAssignID),
+		DestinationParentId: aws.String(destinationOU),
+		SourceParentId:      aws.String(rootID),
+	}
+	_, err = awsClient.MoveAccount(inputMove)
+	if err != nil {
+		fmt.Println(err.Error())
+		return err
+	}
```
dastergon, line 144:
```suggestion
	}
	
```

cmd/account/account-assign.go
```diff
@@ -0,0 +1,150 @@
...
+	//Move account to developers OU
+	inputMove := &organizations.MoveAccountInput{
+		AccountId:           aws.String(accountAssignID),
+		DestinationParentId: aws.String(destinationOU),
+		SourceParentId:      aws.String(rootID),
+	}
```
dastergon, line 139:
```suggestion
	}
	
```

cmd/account/account-assign.go
```diff
@@ -0,0 +1,150 @@
...
+			{
+				Key:   aws.String("owner"),
+				Value: aws.String(o.username),
+			},
+			{
+				Key:   aws.String("claimed"),
+				Value: aws.String("true"),
+			},
+		},
+	}
```
dastergon, line 128:
```suggestion
	}
	
```

cmd/account/account-assign.go
```diff
@@ -0,0 +1,150 @@
...
+	_, err = awsClient.TagResource(inputTag)
+	if err != nil {
+		fmt.Println(err.Error())
+		return err
+	}
+	//Move account to developers OU
```
dastergon, line 134:
```suggestion
	// Move account to developers OU
```

cmd/account/account-assign.go
```diff
@@ -0,0 +1,150 @@
...
+	//Get one account and tag it
+	a := accounts.Accounts[0]
+	accountAssignID = *a.Id
+	//Create input for tagging
+	inputTag := &organizations.TagResourceInput{
+		ResourceId: aws.String(accountAssignID),
+		Tags: []*organizations.Tag{
+			{
+				Key:   aws.String("owner"),
+				Value: aws.String(o.username),
```
dastergon, line 121: what happens when username is empty, do we want to have the "owner" tag empty?

cmd/account/account-assign.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (o *accountAssignOptions) run() error {
+
+	var (
+		accountAssignID string
+		destinationOU   string
+		rootID          string
+	)
+
+	if o.username != "" && o.payerAccount == "osd-staging-2" {
```
dastergon, line 88: You could declutter a bit the if condition, by initializing some values initially to rootID / destinationOU (lets say for osd-staging-1) and then if you notice that the payer account is osd-staging-2 then update the values of the vars

dastergon (changes requested):

cmd/account/account-assign.go
katherinelc321: The username can't be empty. If it's not provided with the flag, the cmd will throw an error. It's handled by the func in line 64

PR #102, opened 2021-06-15
List account by user cmd

katherinelc321:
This cmd does the following:

    - If the username is provided, it lists all the accounts assigned to the user. Otherwise it outputs all the accounts in that profile with the respective owner
    - if an account ID is provided, it outputs its username


katherinelc321:
This cmd does the following:
- If the username is provided, it lists all the accounts assigned to the user. Otherwise it outputs all the accounts in that profile with the respective owner
- if an account ID is provided, it outputs its username

dastergon (changes requested):
Nice job @katherinelc321

Some comments:

I see that you don't separate the code with newlines. Can you separate the different blocks of code with some new lines? It helps with reading the code and separates the boundaries for each code path. 

Can you also add a description to your PR?

Finally, can you separate your code from PR #98?

PR #103, opened 2021-06-17
List dependent resources of a clusterdeployment

NautiluX:
* Add command `osdctl clusterdeployment listresources`
* This will be necessary to identify resources of a given cluster for
  relocating them from one hive to another

jira: [OSD-7169](https://issues.redhat.com/browse/OSD-7169)


NautiluX:
This work is not complete yet, however, if we agree this is useful and should go into osdctl you can merge it and we add resources as we discover they need to be added.

cmd/clusterdeployment/listresources.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func runListResources(cmd *cobra.Command, o *listResourcesOptions) error {
+	if o.clusterId == "" {
+		return cmdutil.UsageErrorf(cmd, "No cluster ID specified, use -C to set one")
```
georgettica, line 66: I added support to use clusterid, external and cluster name.. maybe to phase 2 add support to any value?

georgettica (approved):
The code looks cohesive and usefull overall, and I don't see a problem with it in the codebase as of now

I did add a comment on phase 2 (maybe repo wide) we might want to take

cmd/clusterdeployment/listresources.go
NautiluX: is there a utility function I can use to extract the cluster ID from those identifiers?

fahlmant (approved):
/lgtm

PR #105, opened 2021-06-18
Add assign cmd

katherinelc321:
Assign Cmd takes a LDAP username and payer account type, finds an unassigned account in root, adds the username and claimed as tag, and moves the acct into the developers OU


dastergon (approved):
/lgtm but I would wait for an LGTM from @iamkirkbater as well.

PR #106, opened 2021-06-18
Add list accounts cmd

katherinelc321:
This cmd takes an LDAP username and payer profile and lists the accounts assigned to that user. If no username is provided, it lists all the account from the developers OU


dastergon (approved):
Only some small comments

/lgtm 

but I would also wait for an LGTM from @iamkirkbater as well.

PR #107, opened 2021-06-22
fixed the region problem for : osdctl account cli -i  -o env

lnguyen1401:
Changes made:
1. in osdctl/pkg/k8s/clusterresourcefactory.go , line 230 changed `Region:          factory.Awscloudfactory.Region,` to ` Region:          accountClaim.Spec.Aws.Regions[0].Name,` because Region is already available in the accountClaim 
2. changed the order so that it would go in and try to get account claim from cluster id first(if there's any) and then try to create a new AWS client profile if there's not accountClaim.

### DIFF
--- a/pkg/k8s/clusterresourcefactory.go
+++ b/pkg/k8s/clusterresourcefactory.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+
 	"strings"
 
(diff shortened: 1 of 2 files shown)


lnguyen1401:
assign @fahlmant

fahlmant:
@lnguyen1401 Can you update the description to explain what you're changing and why?

fahlmant (approved):
/lgtm

PR #109, opened 2021-06-24
Add assign account cmd

katherinelc321:
The assign cmd takes an LDAP username, finds an available account in the root, tags it as claimed with the username and moves it to the developers OU. [OSD-7350](https://issues.redhat.com/browse/OSD-7350)

### DIFF
--- a/cmd/account/mgmt/account-assign.go
+++ b/cmd/account/mgmt/account-assign.go
@@ -1,0 +1,199 @@
+package mgmt
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/organizations"
+	"github.com/openshift/osdctl/pkg/printer"
+	awsprovider "github.com/openshift/osdctl/pkg/provider/aws"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+// Global variables
+var (
+	OSDStaging2RootID = "r-rs3h"
+	OSDStaging2OuID   = "ou-rs3h-ry0hn2l9"
+	OSDStaging1RootID = "r-0wd6"
+	OSDStaging1OuID   = "ou-0wd6-z6tzkjek"
+)
+
+type accountAssignOptions struct {
+	awsClient    awsprovider.Client
+	username     string
+	payerAccount string
+
+	flags      *genericclioptions.ConfigFlags
+	printFlags *printer.PrintFlags
+	genericclioptions.IOStreams
+}
+
+func newAccountAssignOptions(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *accountAssignOptions {
+	return &accountAssignOptions{
+		flags:      flags,
+		printFlags: printer.NewPrintFlags(),
+		IOStreams:  streams,
+	}
+}
+
+// assignCmd assigns an aws account to user under osd-staging-2 by default unless osd-staging-1 is specified
+func newCmdAccountAssign(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
+	ops := newAccountAssignOptions(streams, flags)
+	accountAssignCmd := &cobra.Command{
+		Use:               "assign",
+		Short:             "Assign account to user",
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(ops.complete(cmd, args))
+			cmdutil.CheckErr(ops.run())
+		},
+	}
+	ops.printFlags.AddFlags(accountAssignCmd)
+	accountAssignCmd.Flags().StringVarP(&ops.payerAccount, "payer-account", "p", "", "Payer account type")
+	accountAssignCmd.Flags().StringVarP(&ops.username, "username", "u", "", "LDAP username")
+
+	return accountAssignCmd
+}
+
+func (o *accountAssignOptions) complete(cmd *cobra.Command, _ []string) error {
+	if o.username == "" {
+		return cmdutil.UsageErrorf(cmd, "LDAP username was not provided")
+	}
+	if o.payerAccount == "" {
+		return cmdutil.UsageErrorf(cmd, "Payer account was not provided")
+	}
+
+	return nil
+}
+
+func (o *accountAssignOptions) run() error {
+
+	var (
+		accountAssignID string
+		destinationOU   string
+		rootID          string
+	)
+
+	if o.payerAccount == "osd-staging-1" {
+		rootID = OSDStaging1RootID
+		destinationOU = OSDStaging1OuID
+	} else if o.payerAccount == "osd-staging-2" {
+		rootID = OSDStaging2RootID
+		destinationOU = OSDStaging2OuID
+	} else {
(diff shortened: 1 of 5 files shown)


cmd/account/mgmt/account-assign.go
```diff
@@ -0,0 +1,185 @@
...
+	OSDStaging2OuID   = "ou-rs3h-ry0hn2l9"
+	OSDStaging1RootID = "r-0wd6"
+	OSDStaging1OuID   = "ou-0wd6-z6tzkjek"
+)
+
+type accountAssignOptions struct {
+	awsClient    awsprovider.Client
+	username     string
+	payerAccount string
+	output       string
```
iamkirkbater, line 27: Similar to the listAccounts command, can we remove the `output` flag here since we're not using it.

cmd/account/mgmt/account-assign.go
```diff
@@ -0,0 +1,185 @@
...
+		destinationOU   string
+		rootID          string
+	)
+
+	rootID = OSDStaging2RootID
+	destinationOU = OSDStaging2OuID
+	if o.username != "" && o.payerAccount == "osd-staging-1" {
+		rootID = OSDStaging1RootID
+		destinationOU = OSDStaging1OuID
+	}
```
iamkirkbater, line 85: We should explicitly check for the payer account here, otherwise we default any string entered to osd-staging-2 which could cause confusion if someone was trying to run this on rhcontrol.  The ideal behavior would be that we explicitly fail.  Maybe we should add that check in the `complete` function above, ensure that it's either `osd-staging-1` or `osd-staging-2`.  If we don't add the check there, we should do something like this:

```suggestion
if o.payerAccount == "osd-staging-1" {
    rootID = OSDStaging1RootID
    destinationOU = OSDStaging1OuID
} else if o.payerAccount == "osd-staging-2" {
 	rootID = OSDStaging2RootID
	destinationOU = OSDStaging2OuID
} else {
    return fmt.Errorf("Invalid payer account provided")
}
```

cmd/account/mgmt/account-assign.go
```diff
@@ -0,0 +1,185 @@
...
+
+	var (
+		accountAssignID string
+		destinationOU   string
+		rootID          string
+	)
+
+	rootID = OSDStaging2RootID
+	destinationOU = OSDStaging2OuID
+	if o.username != "" && o.payerAccount == "osd-staging-1" {
```
iamkirkbater, line 82: What is the purpose of checking for the username to not be an empty string here?  Do we need that check?  We shouldn't need that as the check is happening in the `complete` function above.

cmd/account/mgmt/account-assign.go
```diff
@@ -0,0 +1,185 @@
...
+		return err
+	}
+
+	fmt.Fprintln(o.IOStreams.Out, accountAssignID)
+
+	return nil
+}
+
+var ErrNoUntaggedAccounts = fmt.Errorf("no untagged accounts available")
+var ErrNoAccountsInRoot = fmt.Errorf("no accounts available in root")
```
iamkirkbater, line 114: This message might be better to change to something like `no available accounts`.  The "in root" part might just add unnecessary confusion to users running this who may or may not understand AWS OUs.

To clarify: the error name is fine and descriptive for us, but let's change the wording of the error that gets output to the end-user.

cmd/account/mgmt/account-assign.go
```diff
@@ -0,0 +1,185 @@
...
+	for _, a := range accounts.Accounts {
+
+		inputListTags := &organizations.ListTagsForResourceInput{
+			ResourceId: a.Id,
+		}
+		tags, err := o.awsClient.ListTagsForResource(inputListTags)
+		if err != nil {
+			return "", err
+		}
+		if len(tags.Tags) == 0 {
```
iamkirkbater, line 138: I'm not sure if we want to specifically look at the length of the tags slice, this might put us in a situation where we tag accounts for other reasons and this starts failing, even if those accounts are eligible to be claimed.

We should at least check for the `claimed: true` tag.

pkg/provider/aws/client.go
```diff
@@ -66,6 +66,16 @@ type Client interface {
...
+	DeleteLoginProfile(*iam.DeleteLoginProfileInput) (*iam.DeleteLoginProfileOutput, error)
+	ListSigningCertificates(*iam.ListSigningCertificatesInput) (*iam.ListSigningCertificatesOutput, error)
+	DeleteSigningCertificate(*iam.DeleteSigningCertificateInput) (*iam.DeleteSigningCertificateOutput, error)
+	ListUserPolicies(*iam.ListUserPoliciesInput) (*iam.ListUserPoliciesOutput, error)
+	DeleteUserPolicy(*iam.DeleteUserPolicyInput) (*iam.DeleteUserPolicyOutput, error)
+	ListAttachedUserPolicies(*iam.ListAttachedUserPoliciesInput) (*iam.ListAttachedUserPoliciesOutput, error)
+	DetachUserPolicy(*iam.DetachUserPolicyInput) (*iam.DetachUserPolicyOutput, error)
+	ListGroupsForUser(*iam.ListGroupsForUserInput) (*iam.ListGroupsForUserOutput, error)
+	RemoveUserFromGroup(*iam.RemoveUserFromGroupInput) (*iam.RemoveUserFromGroupOutput, error)
+	DeleteUser(*iam.DeleteUserInput) (*iam.DeleteUserOutput, error)
```
iamkirkbater, line 78: I think these functions can be moved over to the other PR for unassign account.

pkg/provider/aws/client.go
```diff
@@ -75,12 +85,14 @@ type Client interface {
 	RequestServiceQuotaIncrease(*servicequotas.RequestServiceQuotaIncreaseInput) (*servicequotas.RequestServiceQuotaIncreaseOutput, error)
 
 	// Organizations
-	ListAccountsForParentPages(input *organizations.ListAccountsForParentInput, fn func(*organizations.ListAccountsForParentOutput, bool) bool) error
```
iamkirkbater, line 78: I think we need to make sure that this stays, otherwise we break the `account mgmt list` functionality.

iamkirkbater (changes requested):
Overall this is a great start. All the changes are relatively minor. I'll build this locally and test it out and leave any additional feedback in another review if it's needed.

cmd/account/mgmt/account-assign.go
```diff
@@ -0,0 +1,188 @@
...
+	return nil
+}
+
+var ErrNoUntaggedAccounts = fmt.Errorf("no untagged accounts available")
+var ErrNoAccountsInRoot = fmt.Errorf("no accounts available in root")
+
+func (o *accountAssignOptions) findUntaggedAccount() (string, error) {
+	//List accounts that are not in any OU
+	input := &organizations.ListAccountsInput{}
+	accounts, err := o.awsClient.ListAccounts(input)
```
iamkirkbater, line 122: I think we need to use the `ListAccountsForParent` function, passing in the root ou here.  Otherwise we get _all_ accounts, whether they're in a different OU or not, which will throw an error when attempting to "Move Account" later because AWS will say the account "doesn't exist" but it doesn't exist in the root OU, it's in another OU.

iamkirkbater (changes requested):

pkg/provider/aws/client.go
katherinelc321: that function was taken out and refactored to make it easier for testing

iamkirkbater (approved):

iamkirkbater:
/lgtm

Great work.

PR #111, opened 2021-06-24
Add list account cmd

katherinelc321:
This cmd takes an username and lists all the accounts assigned to it, or takes an account ID and prints the owner of it. Otherwise it prints all the accounts IDs grouped by owners

### DIFF
--- a/cmd/account/mgmt/account-list.go
+++ b/cmd/account/mgmt/account-list.go
@@ -1,0 +1,233 @@
+package mgmt
+
+import (
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/organizations"
+	"github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi"
+	"github.com/openshift/osdctl/pkg/printer"
+	awsprovider "github.com/openshift/osdctl/pkg/provider/aws"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+type accountListOptions struct {
+	awsClient    awsprovider.Client
+	m            map[string][]string
+	username     string
+	payerAccount string
+	accountID    string
+
+	flags      *genericclioptions.ConfigFlags
+	printFlags *printer.PrintFlags
+	genericclioptions.IOStreams
+}
+
+func newAccountListOptions(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *accountListOptions {
+	return &accountListOptions{
+		flags:      flags,
+		printFlags: printer.NewPrintFlags(),
+		IOStreams:  streams,
+	}
+}
+
+func newCmdAccountList(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
+	ops := newAccountListOptions(streams, flags)
+	accountListCmd := &cobra.Command{
+		Use:               "list",
+		Short:             "List out accounts for username",
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(ops.complete(cmd, args))
+			cmdutil.CheckErr(ops.run())
+		},
+	}
+	ops.printFlags.AddFlags(accountListCmd)
+	accountListCmd.Flags().StringVarP(&ops.username, "user", "u", "", "LDAP username")
+	accountListCmd.Flags().StringVarP(&ops.payerAccount, "payer-account", "p", "", "Payer account type")
+	accountListCmd.Flags().StringVarP(&ops.accountID, "account-id", "i", "", "Account ID")
+
+	return accountListCmd
+}
+
+func (o *accountListOptions) complete(cmd *cobra.Command, _ []string) error {
+	if o.payerAccount == "" {
+		return cmdutil.UsageErrorf(cmd, "Payer account was not provided")
+	}
+	if o.username != "" && o.accountID != "" {
+		return cmdutil.UsageErrorf(cmd, "Cannot provide both username and account ID")
+	}
+	return nil
+}
+
+func (o *accountListOptions) run() error {
+
+	var (
+		OuID string
+	)
+	// Instantiate Aws client
+	awsClient, err := awsprovider.NewAwsClient(o.payerAccount, "us-east-1", "")
+	if err != nil {
+		return err
+	}
+
+	if o.payerAccount == "osd-staging-2" {
+		OuID = "ou-rs3h-ry0hn2l9"
+	} else if o.payerAccount == "osd-staging-1" {
+		OuID = "ou-0wd6-z6tzkjek"
+	}
+
+	o.awsClient = awsClient
+	if o.accountID != "" {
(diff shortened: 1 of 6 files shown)


cmd/account/account-list.go
```diff
@@ -0,0 +1,179 @@
...
+		user           string
+		OuID           string
+	)
+	// Instantiate Aws client
+	awsClient, err := awsprovider.NewAwsClient(o.payerAccount, "us-east-1", "")
+	if err != nil {
+		return err
+	}
+
+	if o.payerAccount == "osd-staging-2" {
```
georgettica, line 86: Does that mean that any other user cannot set the ouid?

Also this means our personal IDs are defined in the source code.. is there a better way to extract this knowledge?

cmd/account/account-list.go
```diff
@@ -0,0 +1,179 @@
...
+				if t.Key == aws.String("owner") {
+					user = *t.Value
+				}
+			}
+			// Check if user is already in the hashmap
+			// If yes append to list of accounts
+			// If no create list of accounts and add it as value to key
+			_, ok := m[user]
+			if ok {
+				_ = append(m[user], *a.Id)
```
georgettica, line 122: Why use two practices for array expansion in the same fnction?


Which is the one you would like to follow repo wide and choose it

georgettica (commented):
Looks good overall, tiny tiny nits then I will lgtm

Unless there is a logic flaw (didn't test it locally)

cmd/account/account-list.go
```diff
@@ -0,0 +1,179 @@
...
+type accountListOptions struct {
+	username     string
+	payerAccount string
+	accountID    string
+	output       string
+
+	flags      *genericclioptions.ConfigFlags
+	printFlags *printer.PrintFlags
+	genericclioptions.IOStreams
+	kubeCli client.Client
```
iamkirkbater, line 27: We aren't using the kubernetes CLI, so let's get rid of this dependency.

cmd/account/account-list.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (o *accountListOptions) complete(cmd *cobra.Command, _ []string) error {
+	if o.payerAccount == "" {
+		return cmdutil.UsageErrorf(cmd, "Payer account was not provided")
+	}
+	if o.username != "" && o.accountID != "" {
+		return cmdutil.UsageErrorf(cmd, "Cannot provide both username and account ID")
+	}
+	var err error
+	o.kubeCli, err = k8s.NewClient(o.flags)
```
iamkirkbater, line 66: We don't have a dependency on kubernetes for this command, so let's remove this so that we don't need to run this on-cluster.

cmd/account/account-list.go
iamkirkbater: I wonder if we could do something like do a lookup of the OU based on name.  OU lookups though are rate-limited to only a SINGLE lookup at a time across the entire account, so if both user A and user B try to lookup the command at a single time then it will fail for whoever isn't currently doing the lookup.  And AWS Account Operator is almost _constantly_ looking this information up, making this fail _a lot_, especially now that there are multiple shards of that, so it would make this process basically awful.

We can try it, though, because maybe I'm confusing the OU lookup vs Account Lookup APIs and it might not be the _same_ problem.

cmd/account/account-list.go
iamkirkbater: I'm not sure I follow what you're asking in this comment.  Specifically what's happening here is that we're checking the key exists in the map before attempting to append the ID to it. Otherwise we'll get panics.  If the key doesn't exist, we initialize a new list as the value for that key.

cmd/account/account-list.go
```diff
@@ -0,0 +1,179 @@
...
+			},
+		}
+		accounts, err := awsClient.GetResources(inputFilterTag)
+		if err != nil {
+			return err
+		}
+
+		// Get last 9 digits of ResourceARN and append it to account list
+		for _, a := range accounts.ResourceTagMappingList {
+			tempAccountIDs = append(tempAccountIDs, (*a.ResourceARN)[len(*a.ResourceARN)-9:])
```
iamkirkbater, line 148: AWS accounts are 12 characters long, this needs to be changed.

We could probably split on `/` and get the last value for some "future-proofing" in case AWS changes that in the future, but I'll leave that decision up to you.

cmd/account/account-list.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		for _, t := range val.Tags {
+			if t.Key == aws.String("owner") {
+				fmt.Fprintln(o.IOStreams.Out, *t.Value)
+			}
+		}
+	}
+
+	if o.output == "" {
```
iamkirkbater, line 172: Let's remove the `o.output` flag at this time, and then we can iterate back on this to add other output options in a different PR.

iamkirkbater (changes requested):
There are a few changes that need to be made.

Specifically, we _need_ to change the string length for the account ID to 12 from 9, and we really should remove the kubernetes dependency.

Let's also remove the output formatting flag for this PR, and once this lands we can circle back and iterate that back in.

Let's also work together to get this into a `mgmt` subcommand, so that this doesn't collide with the already existing `account list` command, so it would be `account mgmt list`

cmd/account/mgmt/account-list.go
```diff
@@ -0,0 +1,167 @@
...
+		if err != nil {
+			return err
+		}
+
+		for _, t := range val.Tags {
+			if t.Key == aws.String("owner") {
+				fmt.Fprintln(o.IOStreams.Out, *t.Value)
+			}
+		}
+	}
```
iamkirkbater, line 160: Let's move this block above the `if o.username == ""` block, so move this up to line 91.  Let's also add a `return nil` after the for loop within this so we don't worry about building the hashmap when we already have the tag for this one account.

iamkirkbater (changes requested):

iamkirkbater (approved):

iamkirkbater:
@katherinelc321 and I paired this afternoon to make the required changes to get this done.

Should be all set to go.  We will be pairing up some time next week in order to refactor this a bit more and work on adding some unit tests.

cmd/account/mgmt/account-list.go
```diff
@@ -0,0 +1,235 @@
...
+	o.m = make(map[string][]string)
+
+	if o.username != "" {
+
+		user := o.username
+		o.m[user], err = o.listAccountsByUser(user)
+		if err != nil {
+			return err
+		}
+		return nil
```
iamkirkbater, line 103: We need to remove this line, or add the `for key, value...` loop that's on line 115 here.

cmd/account/mgmt/account-list.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	if o.username == "" && o.accountID == "" {
+
+		o.m, err = o.listAllAccounts(OuID)
+		if err != nil {
+			return err
+		}
+		return nil
```
iamkirkbater, line 112: We need to remove this line, or add the `for key, value...` loop that's on line 115 here.

iamkirkbater (changes requested):
One small change, other than that it looks great and I'll merge once that's done.

While you're either removing the line or duplicating the for loop (I'll leave the decision on how to do that to you) can you squash this down to a single commit before you push up that change?

PR #112, opened 2021-06-24
fixed piping through jq functionality

lnguyen1401:
Reconfigure the output of osdctl account cli -C 1jv3v7h0entnarjo6j06ceahlukr0eqj -p rhcontrol -o json because previously it didn't print out json format, therefore can't be piped through jq.


fahlmant (approved):
/lgtm

PR #113, opened 2021-06-25
fixed nil pointer dereference error when no accountclaim is available

lnguyen1401:
Command `./osdctl account cli -i 335462696174 -p osd-staging-2 ' returns 'panic: runtime error: invalid memory address or nil pointer dereference' because it tried to read an accountclaim when there doesn't exist any.


fahlmant (changes requested):
Please remove delete.json and the osdctl binary

fahlmant (approved):
/lgtm

PR #116, opened 2021-07-02
Add IAM Secret do dependent resources list

NautiluX:
* Add option to list external resources only (don't include cluster namespace)
* Add tests
* Add Secret of account IAM user to list
* Fix Dockerfile


fahlmant (approved):
/lgtm

PR #117, opened 2021-07-02
Adds initial `aao pool` command structure

iamkirkbater:
Adds the aao subcommand and `aao pool` commands to osdctl to get the account pool size.

### DIFF
--- a/cmd/aao/pool.go
+++ b/cmd/aao/pool.go
@@ -2,6 +2,7 @@ package aao
 
 import (
 	"fmt"
+	"os/exec"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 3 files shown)


cmd/aao/pool.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+func (o *poolOptions) complete(cmd *cobra.Command) error {
+	return nil
+}
+
+func (o *poolOptions) run() error {
+	cmd := "oc get accounts -o json -n aws-account-operator | jq '.items | map(select(.status.claimed==\"false\" or .status.claimed==null)) | map(select(.status.state!=\"Failed\")) | map(select(.spec.legalEntity==null or .spec.legalEntity.id==\"\" or .spec.legalEntity.id==null)) | length'"
+	output, err := exec.Command("bash", "-c", cmd).Output()
```
dastergon, line 48: Since we are already use the kube client code in osdctl, we can use it to query  accounts from the `aws-account-operator` namespace, and also process the JSON with Go. That will make the code logic a bit more clear and we won't have binaries as dependencies. That sounds like a good task for @katherinelc321

dastergon (changes requested):

cmd/aao/pool.go
iamkirkbater: Yeah, there's definitely room for a refactor here, but I wanted to get a tool out the door as quickly as possible to help alleviate some of the issues being seen by other teams.

dastergon (approved):
/lgtm 

@iamkirkbater can you create a card in our backlog, so that Katherine can rework this, once she finds the time?

PR #118, opened 2021-07-02
Add command osdctl cluster health [OSD-7422]

aliceh:
In this commit I added a command that lists the expected number of nodes for a cluster and the actual number of running instances in AWS. The command outputs an error message if the cloud provider is not AWS.
Usage ' osdctl  cluster health -p [account profile] -C [internal cluster ID] '. This command requires the ocm refresh token set as an enviromental variable OCM_TOKEN.


clcollins:
Your PR includes a vendor dir, but osdctl is using Go modules. That needs to be addressed.

The error, I think, is just `go fmt` in the lint finding files that need to be formatted and committed.  If you address the vendor dir, I think it's just one line in status.go:

```go
-       for idx, _ := range instances.Reservations {
+       for idx := range instances.Reservations {
```

cmd/cluster/status.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (o *statusOptions) run() error {
+
+	// This call gets the availability zone of the cluster, as well as the numbers of compute, infra, and master nodes.
+	az, compute, infra, master, err := ocmDescribe(o.k8sclusterresourcefactory.ClusterID)
```
iamkirkbater, line 84: We should add an explicit check to see what the cloud provider of an cluster is, in order to explicitly fail for non-AWS clusters so SREs running this on, say, a GCP cluster won't spend time debugging why their AWS access isn't working.

cmd/cluster/status.go
```diff
@@ -0,0 +1,199 @@
...
+	log.Println("Getting instances info:")
+
+	instances, err := awsJumpClient.DescribeInstances(&ec2.DescribeInstancesInput{})
+
+	totalRunning := 0
+
+	//Here we count the number of customer's running instances in the region
+	for idx := range instances.Reservations {
+		for _, inst := range instances.Reservations[idx].Instances {
+			if *inst.State.Name == "running" {
```
iamkirkbater, line 144: While we should also check for instances running, we should check the tags on a given instance to make sure that it's part of the cluster that we're looking at.  

With CCS we added the ability to add multiple clusters to the same account, and with BYOVPC customers could be running their own workloads on other instances within their cluster.

iamkirkbater (changes requested):
While the two changes are pretty important functional and usability changes, this is a great first pass at this functionality, and I acknowledge the amount of work that was put into this and it's very appreciated.

cmd/cluster/status.go
aliceh: The command will throw an "Error: cluster provider is not AWS", the check is on L194. I agree that the error message could be better, for example, "This command is supported only for AWS clusters. The command is not supported for <cluster provider extracted by the command> clusters."

cmd/cluster/status.go
aliceh: Good point and good suggestion. I agree.

cmd/cluster/status.go
aliceh: I added the change.

cmd/cluster/status.go
aliceh: I added the change.

iamkirkbater:
Awesome!  This looks great!

/lgtm

iamkirkbater (approved):

PR #119, opened 2021-07-07
Listresources: Print quotes on empty api version

NautiluX:
* allowes for easier combination with column based tools like `awk`


fahlmant (approved):
/lgtm

PR #123, opened 2021-07-13
Add unassign account

katherinelc321:
This cmd takes either an username or account ID and removes its IAM roles, policies, groups, untags it and moves it back to the root Ou. [OSD-7361](https://issues.redhat.com/browse/OSD-7361)

### DIFF
--- a/cmd/account/mgmt/account-unassign.go
+++ b/cmd/account/mgmt/account-unassign.go
@@ -405,6 +405,6 @@ func (o *accountUnassignOptions) deleteUser(user string) error {
 		return err
 	}
 
-	fmt.Printf("user %s successfully deleted", user)
+	fmt.Printf("user %s successfully deleted\n", user)
 	return nil
 }
--- a/cmd/account/mgmt/account-unassign.go
+++ b/cmd/account/mgmt/account-unassign.go
@@ -1,0 +1,410 @@
+package mgmt
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/iam"
+	"github.com/aws/aws-sdk-go/service/organizations"
+	"github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi"
+	"github.com/openshift/osdctl/pkg/printer"
+	awsprovider "github.com/openshift/osdctl/pkg/provider/aws"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+func newCmdAccountUnassign(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
+	ops := newAccountUnassignOptions(streams, flags)
+	accountUnassignCmd := &cobra.Command{
+		Use:               "unassign",
+		Short:             "Unassign account to user",
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(ops.complete(cmd, args))
+			cmdutil.CheckErr(ops.run())
+		},
+	}
+	ops.printFlags.AddFlags(accountUnassignCmd)
+	accountUnassignCmd.Flags().StringVarP(&ops.payerAccount, "payer-account", "p", "", "Payer account type")
+	accountUnassignCmd.Flags().StringVarP(&ops.username, "username", "u", "", "LDAP username")
+	accountUnassignCmd.Flags().StringVarP(&ops.accountID, "account-id", "i", "", "Account ID")
+
+	return accountUnassignCmd
+}
+
+type accountUnassignOptions struct {
+	awsClient    awsprovider.Client
+	username     string
+	payerAccount string
+	accountID    string
+
+	flags      *genericclioptions.ConfigFlags
+	printFlags *printer.PrintFlags
+	genericclioptions.IOStreams
+}
+
+func newAccountUnassignOptions(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *accountUnassignOptions {
+	return &accountUnassignOptions{
+		flags:      flags,
+		printFlags: printer.NewPrintFlags(),
+		IOStreams:  streams,
+	}
+}
+
+func (o *accountUnassignOptions) complete(cmd *cobra.Command, _ []string) error {
+	if o.payerAccount == "" {
+		return cmdutil.UsageErrorf(cmd, "Payer account was not provided")
+	}
+	if o.username == "" && o.accountID == "" {
+		return cmdutil.UsageErrorf(cmd, "Please provide either an username or account ID")
+	}
+	return nil
+}
+
+func (o *accountUnassignOptions) run() error {
+
+	var (
+		accountUsername string
+		accountIdList   []string
+		destinationOU   string
+		rootID          string
+	)
+
+	// Instantiate Aws client
+	awsClient, err := awsprovider.NewAwsClient(o.payerAccount, "us-east-1", "")
+	if err != nil {
+		return err
+	}
+
+	if o.payerAccount == "osd-staging-1" {
+		rootID = OSDStaging1RootID
+		destinationOU = OSDStaging1OuID
+	} else if o.payerAccount == "osd-staging-2" {
+		rootID = OSDStaging2RootID
+		destinationOU = OSDStaging2OuID
+	} else {
+		return fmt.Errorf("invalid payer account provided")
+	}
+
+	o.awsClient = awsClient
+
+	if o.accountID != "" {
+
+		accountUsername, err = o.checkForHiveNameTag(o.accountID)
+		if err != nil {
+			return err
+		}
+		accountIdList = append(accountIdList, o.accountID)
+
+	}
+
+	if o.username != "" {
+		// Check that username is not a hive
+		if strings.HasPrefix(o.username, "hive") {
+			return ErrHiveNameProvided
+		}
+
+		accountUsername = o.username
+		accountIdList, err = o.listAccountsFromUser(accountUsername)
+		if err != nil {
+			return err
+		}
+	}
+
+	fmt.Printf("Are you sure you want to unassign account(s) [%v] from %s? [y/n] ", accountIdList, accountUsername)
+	reader := bufio.NewReader(os.Stdin)
+
+	response, err := reader.ReadString('\n')
+	if err != nil {
+		return err
+	}
+
+	response = strings.ToLower(response[0:1])
+	if response != "y" {
+		os.Exit(0)
+	}
+
+	// loop through accounts list and untag and move them back into root OU
+	for _, id := range accountIdList {
+
+		err = o.untagAccount(id)
+		if err != nil {
+			return err
+		}
+
+		err = o.moveAccount(id, rootID, destinationOU)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Delete login profile
+	err = o.deleteLoginProfile(accountUsername)
+	if err != nil {
+		return err
+	}
+	// Delete access keys
+	err = o.deleteAccessKeys(accountUsername)
+	if err != nil {
+		return err
+	}
+	// Delete signing certificates
+	err = o.deleteSigningCert(accountUsername)
+	if err != nil {
+		return err
+	}
+	// Delete policies
+	err = o.deletePolicies(accountUsername)
+	if err != nil {
+		return err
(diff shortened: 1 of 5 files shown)


cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+	if len(accounts.ResourceTagMappingList) == 0 {
+		return []string{}, ErrNoAccountsForUser
+	}
+
+	var accountIdList []string
+	// Get last 12 digits of ResourceARN and append it to account list
+	for _, a := range accounts.ResourceTagMappingList {
+		accountIdList = append(accountIdList, (*a.ResourceARN)[len(*a.ResourceARN)-12:])
+	}
+	if err != nil {
```
drpaneas, line 266: You forgot to test this situation. Please add this as part of your unit-tests ;)
Not to mention this is condition is always 'false' because 'err' is always 'nil' in this part of the code! Ouch.

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+}
+
+func (o *accountUnassignOptions) deleteLoginProfile(user string) error {
+
+	inputDeleteLogin := &iam.DeleteLoginProfileInput{
+		UserName: &user,
+	}
+	_, err := o.awsClient.DeleteLoginProfile(inputDeleteLogin)
+	if err != nil {
+		return err
```
drpaneas, line 279: You forgot to test this situation. Please add this as part of your unit-tests ;)

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+}
+
+func (o *accountUnassignOptions) deleteAccessKeys(user string) error {
+
+	inputListAccessKeys := &iam.ListAccessKeysInput{
+		UserName: &user,
+	}
+	accessKeys, err := o.awsClient.ListAccessKeys(inputListAccessKeys)
+	if err != nil {
+		return err
```
drpaneas, line 291: You forgot to test this situation. Please add this as part of your unit-tests ;)

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+
+	for _, m := range accessKeys.AccessKeyMetadata {
+
+		inputDelKey := &iam.DeleteAccessKeyInput{
+			AccessKeyId: m.AccessKeyId,
+			UserName:    &user,
+		}
+		_, err = o.awsClient.DeleteAccessKey(inputDelKey)
+		if err != nil {
```
drpaneas, line 301: You forgot to test this situation. Please add this as part of your unit-tests ;)

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+	return nil
+}
+
+func (o *accountUnassignOptions) deleteSigningCert(user string) error {
+
+	inputListCert := &iam.ListSigningCertificatesInput{
+		UserName: &user,
+	}
+	cert, err := o.awsClient.ListSigningCertificates(inputListCert)
+	if err != nil {
```
drpaneas, line 314: You forgot to test this situation. Please add this as part of your unit-tests ;)

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+		return err
+	}
+
+	for _, c := range cert.Certificates {
+		inputDelCert := &iam.DeleteSigningCertificateInput{
+			CertificateId: c.CertificateId,
+			UserName:      &user,
+		}
+		_, err = o.awsClient.DeleteSigningCertificate(inputDelCert)
+		if err != nil {
```
drpaneas, line 324: You forgot to test this situation. Please add this as part of your unit-tests ;)

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+	return nil
+}
+
+func (o *accountUnassignOptions) deletePolicies(user string) error {
+	inputListPolicies := &iam.ListUserPoliciesInput{
+		UserName: &user,
+	}
+	policies, err := o.awsClient.ListUserPolicies(inputListPolicies)
+	if err != nil {
```
drpaneas, line 336: You forgot to test this situation. Please add this as part of your unit-tests ;)

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+		return err
+	}
+
+	for _, p := range policies.PolicyNames {
+		inputDelPolicies := &iam.DeleteUserPolicyInput{
+			PolicyName: p,
+			UserName:   &user,
+		}
+		_, err = o.awsClient.DeleteUserPolicy(inputDelPolicies)
+		if err != nil {
```
drpaneas, line 346: You forgot to test this situation. Please add this as part of your unit-tests ;)

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+	return nil
+}
+
+func (o *accountUnassignOptions) deleteAttachedPolicies(user string) error {
+	inputListAttachedPol := &iam.ListAttachedUserPoliciesInput{
+		UserName: &user,
+	}
+	attachedPol, err := o.awsClient.ListAttachedUserPolicies(inputListAttachedPol)
+	if err != nil {
```
drpaneas, line 358: You forgot to test this situation. Please add this as part of your unit-tests ;)

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+		return err
+	}
+
+	for _, ap := range attachedPol.AttachedPolicies {
+		inputDetachPol := &iam.DetachUserPolicyInput{
+			PolicyArn: ap.PolicyArn,
+			UserName:  &user,
+		}
+		_, err = o.awsClient.DetachUserPolicy(inputDetachPol)
+		if err != nil {
```
drpaneas, line 368: You forgot to test this situation. Please add this as part of your unit-tests ;)

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+	return nil
+}
+
+func (o *accountUnassignOptions) deleteGroups(user string) error {
+
+	inputListGroups := &iam.ListGroupsForUserInput{
+		UserName: &user,
+	}
+	groups, err := o.awsClient.ListGroupsForUser(inputListGroups)
+	if err != nil {
```
drpaneas, line 381: You forgot to test this situation. Please add this as part of your unit-tests ;)

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+		return err
+	}
+
+	for _, g := range groups.Groups {
+		inputRemoveFromGroup := &iam.RemoveUserFromGroupInput{
+			GroupName: g.GroupName,
+			UserName:  &user,
+		}
+		_, err = o.awsClient.RemoveUserFromGroup(inputRemoveFromGroup)
+		if err != nil {
```
drpaneas, line 391: You forgot to test this situation. Please add this as part of your unit-tests ;)

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+	return nil
+}
+
+func (o *accountUnassignOptions) deleteUser(user string) error {
+	inputDelUser := &iam.DeleteUserInput{
+		UserName: &user,
+	}
+	_, err := o.awsClient.DeleteUser(inputDelUser)
+	if err != nil {
```
drpaneas, line 403: You forgot to test this situation. Please add this as part of your unit-tests ;)

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,370 @@
...
+func (o *accountUnassignOptions) untagAccount(id string) error {
+	inputUntag := &organizations.UntagResourceInput{
+		ResourceId: &id,
+		TagKeys: []*string{
+			aws.String("owner"),
+			aws.String("claimed"),
+		},
+	}
+	_, err := o.awsClient.UntagResource(inputUntag)
+	if err != nil {
```
drpaneas: You forgot to test this situation. Please add this as part of your unit-tests ;)

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,407 @@
...
+		return err
+	}
+	return nil
+}
+
+func (o *accountUnassignOptions) moveAccount(id string, rootID string, destinationOU string) error {
```
drpaneas, line 224: This is not tested.

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,407 @@
...
+
+func newAccountUnassignOptions(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *accountUnassignOptions {
+	return &accountUnassignOptions{
+		flags:      flags,
+		printFlags: printer.NewPrintFlags(),
+		IOStreams:  streams,
+	}
+}
+
+func (o *accountUnassignOptions) complete(cmd *cobra.Command, _ []string) error {
```
drpaneas, line 58: This is not tested

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,407 @@
...
+	if o.payerAccount == "" {
+		return cmdutil.UsageErrorf(cmd, "Payer account was not provided")
+	}
+	if o.username == "" && o.accountID == "" {
+		return cmdutil.UsageErrorf(cmd, "Please provide either an username or account ID")
+	}
+	return nil
+}
+
+func (o *accountUnassignOptions) run() error {
```
drpaneas, line 68: This is not tested. For example what happens when you try to unassign the same account twice? e.g. the first time let's say `deleteSigningCert` threw an error. What would happen the next time? Would that throw error (because it's the SignCert has already been deleted?

drpaneas (changes requested):
Great work Katherine! I have dropped you some comments related to testing, feel free to ping me in Slack if you have any questions about them.

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,405 @@
...
+			// Delete groups
+			err = o.deleteGroups(accountUsername)
+			if err != nil {
+				return err
+			}
+			// Delete user
+			err = o.deleteUser(accountUsername)
+			if err != nil {
+				return err
+			}
```
iamkirkbater, line 177: I see that we're not deleting roles at all here.  We'll want to also list and delete any roles.  The only role we won't want to delete is the role named: `OrganizationAccountAccessRole`, otherwise we lose access to that account.

I'll leave the decision to add this into v1 of this function to you, otherwise we can card the deletion of the roles and circle back to this at some other point.

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,405 @@
...
+
+	fmt.Printf("Are you sure you want to unassign the accounts? [y/n] ")
+	reader := bufio.NewReader(os.Stdin)
+
+	response, err := reader.ReadString('\n')
+	if err != nil {
+		return err
+	}
+
+	if response == "y" || response == "Y" {
```
iamkirkbater, line 126: This isn't working for me during testing, because the var `response` includes the delimiter set on ReadString, so in actuality the `response` is `y\n` or `Y\n`.

Also, since this is an easy exit condition, we should be able to negate this check and exit early here, rather than put all of the logic inside a large `if` block, like this:

```go
response = strings.ToLower(response[0:1]) #only get the first letter of the string, and always lowercase it
if response != "y" {
    os.Exit(0)
}

for _, id := range accountIdList {
...
```

iamkirkbater: By only grabbing the first character and lowercasing it, we allow users to be able to do things like `yes | osdctl account mgmt unassign` for future scriptablity without having to do any changes on our end, and then we don't have to check _every_ capitalization permutation of Yes:

```
yes
yeS
yEs
yES
Yes
YeS
YEs
YES
```

iamkirkbater (changes requested):
You're very much on the right track here. A few small usability changes, as well as the parsing of the response that could use some work.

Can you also explain to those who might come after me to review this in the next two weeks how they would go about testing this using a real account or something?

Keep up the great work!

iamkirkbater:
To those coming after me to review, if the changes that I've asked for get merged, then this should be all set to go, provided it can be tested.

Also worthy to note, not having this merged should not effect the ability to demo this. You should still be able to demo off of this branch.

iamkirkbater (approved):
/lgtm

PR #127, opened 2021-07-20
Added improvements to account for autoscaled compute, and filter out non-master/infra/worker nodes.

aliceh:
Added code to account for autoscaled compute instances and filter for master/infra/worker in the instance name.

### DIFF
--- a/cmd/cluster/health.go
+++ b/cmd/cluster/health.go
@@ -15,8 +15,10 @@ import (
 	awsprovider "github.com/openshift/osdctl/pkg/provider/aws"
 	"github.com/openshift/osdctl/pkg/utils"
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/klog"
+
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/cluster/health.go
```diff
@@ -164,7 +176,7 @@ func (o *healthOptions) run() error {
 //This command implements the ocm describe clsuter call via osm-sdk.
 //This call requires the ocm API Token https://cloud.redhat.com/openshift/token be available in the OCM_TOKEN env variable.
 //Example: export OCM_TOKEN=$(jq -r .refresh_token ~/.ocm.json)
-func ocmDescribe(clusterID string) ([]string, string, int, error) {
+func ocmDescribe(clusterID string) ([]string, string, int, int, int, int, int, error) {
```
iamkirkbater, line 179: I think with this many return parameters we should refactor this to be a specific struct, in order to make this more readable.

Maybe something like:

```
struct clusterDescribeOutput {
  az []string
  clusterName string
  compute int
  infra int
  master int
  autoscaleMin int
  autoscaleMax int
}
```
etc.

iamkirkbater (changes requested):
I like the new feature, but it would be nice if it would be refactored for readability while the context is still fresh, please.

cmd/cluster/health.go
georgettica: If you want some inspiration, see https://github.com/openshift/osdctl/pull/129/commits/4333c8aa3d2ab63634512fe93808dd7e530c4390

aliceh: Thanks!!

aliceh:
I am so sorry, I somehow merged this from CLI by accident.

aliceh:
https://github.com/openshift/osdctl/pull/137 this is the PR I wanted to submit. Sorry for accidentally merging something else into master.  @mmazur, @iamkirkbater could you please revert this?

PR #129, opened 2021-07-21
[cluster health] Condense cmd output for clarity

mmazur:
Output after this patch:

```
ID:         1m2q4ujh9s1tmmtmi5u641rcbbqrdskp
Name:       osde2e-229p0
Provider:   AWS
AZs:        [us-east-1a]
Nodes
  Expected: 9
  Running:  10
```

### DIFF
--- a/cmd/cluster/health.go
+++ b/cmd/cluster/health.go
@@ -84,7 +84,9 @@ func (o *healthOptions) run() error {
 	// This call gets the availability zone of the cluster, as well as the expected number of nodes.
 	az, clusterName, nodesExpected, err := ocmDescribe(o.k8sclusterresourcefactory.ClusterID)
 	if az != nil {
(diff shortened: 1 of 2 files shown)


georgettica:
you know, you could do that in a yaml object so it's parsable

mmazur:
> you know, you could do that in a yaml object so it's parsable

Yeah and I'm sure this will be done properly at some point once the list of checks grows. For now I wanted it fast, so the thing is releasable and we can switch to using it asap.

RiRa12621:
/hold
just because we need it quickly doesn't mean that it cannot be done properly.
@georgettica can you work with @mmazur to get  a proper implementation of this?

mmazur:
It can be done properly if somebody has time to do it before end of week. Otherwise it's more important to do it quickly.

fahlmant (approved):
/lgtm

PR #133, opened 2021-07-26
Add create account

katherinelc321:
This creates a new account if there is no accounts available in the OU

### DIFF
--- a/cmd/account/mgmt/account-assign.go
+++ b/cmd/account/mgmt/account-assign.go
@@ -201,38 +201,18 @@ func (o *accountAssignOptions) buildAccount(seedVal int64) (string, error) {
 
 	orgOutput, orgErr := o.createAccount(seedVal)
 	if orgErr != nil {
-		// If email already exists retry up to 100 times
-		if orgErr == ErrEmailAlreadyExist {
-			err := retry(100, 2*time.Second, func() (err error) {
-				orgOutput, err = o.createAccount(seedVal)
-				return
-			})
-			if err != nil {
-				return "", err
+
+		// If email already exists retry until a new email is generated
+		for orgErr == ErrEmailAlreadyExist {
+			orgOutput, orgErr = o.createAccount(seedVal)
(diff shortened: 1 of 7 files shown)


cmd/account/mgmt/account-assign.go
```diff
@@ -183,6 +195,117 @@ func (o *accountAssignOptions) tagAccount(accountIdInput string) error {
...
+		Email:       aws.String(email),
+	}
+
+	createOutput, err := o.awsClient.CreateAccount(createInput)
+	if err != nil {
+		return &organizations.DescribeCreateAccountStatusOutput{}, err
+	}
+
+	describeStatusInput := &organizations.DescribeCreateAccountStatusInput{
+		CreateAccountRequestId: createOutput.CreateAccountStatus.Id,
```
drpaneas, line 262: This seems to be the problem: there is no value for `CreateAccountRequestId`

unreadable: could not read string len protocol error E08 during memory read for packet $mc4834870246c8b48,1...

cmd/account/mgmt/account-assign_test.go
```diff
@@ -130,6 +147,43 @@ func TestFindUntaggedAccount(t *testing.T) {
...
+	createId := "car-random1234"
+
+	mockAWSClient.EXPECT().CreateAccount(&organizations.CreateAccountInput{
+		AccountName: &accountName,
+		Email:       &email,
+	}).Return(&organizations.CreateAccountOutput{
+		CreateAccountStatus: &organizations.CreateAccountStatus{Id: &createId},
+	}, nil)
+
+	awsDescribeOutput := &organizations.DescribeCreateAccountStatusOutput{}
```
dkeohane, line 170: Problem is here - this is returned from the create call, then we try access it here
https://github.com/openshift/osdctl/pull/133/files#diff-1b2f7e3efa0d33561aa4d30779d7dea340714149fd3233ca89527609670aac5dR273

PR #135, opened 2021-07-28
Add create account

katherinelc321:
Cmd assign dev accounts will create a new aws account if there are no accounts available in the root OU

### DIFF
--- a/cmd/account/mgmt/account-assign.go
+++ b/cmd/account/mgmt/account-assign.go
@@ -2,9 +2,12 @@ package mgmt
 
 import (
 	"fmt"
+	"math/rand"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/organizations"
+
 	"github.com/openshift/osdctl/pkg/printer"
 	awsprovider "github.com/openshift/osdctl/pkg/provider/aws"
 	"github.com/spf13/cobra"
@@ -83,7 +86,7 @@ func (o *accountAssignOptions) run() error {
 		rootID = OSDStaging2RootID
 		destinationOU = OSDStaging2OuID
 	} else {
-		return fmt.Errorf("Invalid payer account provided")
+		return fmt.Errorf("invalid payer account provided")
 	}
 	//Instantiate aws client
 	awsClient, err := awsprovider.NewAwsClient(o.payerAccount, "us-east-1", "")
@@ -93,8 +96,18 @@ func (o *accountAssignOptions) run() error {
 
 	o.awsClient = awsClient
 	accountAssignID, err = o.findUntaggedAccount(rootID)
+
 	if err != nil {
-		return err
+		// If the error returned is not because of a lack of accounts, return the error
+		if err != ErrNoUntaggedAccounts {
+			return err
+		}
+		// otherwise, create a new account
+		seed := time.Now().UnixNano()
+		accountAssignID, err = o.buildAccount(seed)
+		if err != nil {
+			return err
+		}
 	}
 
 	err = o.tagAccount(accountAssignID)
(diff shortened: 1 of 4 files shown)


drpaneas (approved):
/lgtm
Thanks for adding unit-tests :D

cmd/account/mgmt/account-assign.go
```diff
@@ -124,12 +130,18 @@ func (o *accountAssignOptions) findUntaggedAccount(rootOu string) (string, error
...
+
 	if len(accounts.Accounts) == 0 {
-		return "", ErrNoAccountsInRoot
+		fmt.Println("No accounts available to assign")
+
+		accountAssignID, err := o.buildAccount(seedVal)
+		if err != nil {
+			return "", err
+		}
+		return accountAssignID, nil
```
iamkirkbater, line 141: This in general is a code smell, but I'm not going to ask for this to be tackled here, this is a good candidate for a refactor.

The function that this body of work is included in is called `findUntaggedAccount`.  Creating a new account if there is no account available is a side-effect, and _can_ get messy when attempting to debug.

I think a better pattern would be to do something on line 98/99 like this:

```go
accountAssignID, err := o.findUntaggedAccount(rootID)
if err != nil {
  # If the error returned is not because of a lack of accounts, return the error
  if err != ErrNoUntaggedAccounts {
    return "", err
  }
  # otherwise, create a new account
  seed := time.Now().UnixNano()
  accountAssignID, err = o.buildAccount(seed)
  if err != nil {
    return "", err
  }
}
```

Again, I'm not going to ask for this change now, and looking at the code I just wrote I'm not sure which is easier to understand at first, but I think the main difference is that we're not relying on a side-effect functionality as part of our core logic.

cmd/account/mgmt/account-assign.go
```diff
@@ -183,6 +195,100 @@ func (o *accountAssignOptions) tagAccount(accountIdInput string) error {
...
+
+	fmt.Println("Creating account")
+	var newAccountId string
+
+	orgOutput, orgErr := o.createAccount(seedVal)
+	if orgErr != nil {
+
+		// If email already exists retry until a new email is generated
+		for orgErr == ErrEmailAlreadyExist {
+			orgOutput, orgErr = o.createAccount(seedVal)
```
iamkirkbater, line 208: If we provide the same seedVal here, shouldn't it provide us with the same randomstring?

cmd/account/mgmt/account-assign.go
fahlmant: +100

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,405 @@
...
+		rootID          string
+	)
+
+	// Instantiate Aws client
+	awsClient, err := awsprovider.NewAwsClient(o.payerAccount, "us-east-1", "")
+	if err != nil {
+		return err
+	}
+
+	if o.payerAccount == "osd-staging-1" {
```
fahlmant, line 83: Instead of baking these in, could these be spun out into some sort of config file? Or maybe require the caller to provide the OU IDs as well? If we're going to leave these in, can these all be made constants that can be referred to via a library? Not going to block on this, but I'd like to avoid baking in SREP specific strings as much as possible, given this is a public repo.

cmd/account/mgmt/account-unassign.go
```diff
@@ -0,0 +1,405 @@
...
+	_, err := o.awsClient.MoveAccount(inputMove)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+var ErrNoAccountsForUser error = fmt.Errorf("user has no aws accounts")
+
+func (o *accountUnassignOptions) listAccountsFromUser(user string) ([]string, error) {
```
fahlmant, line 240: Is this something that could be exported and used in the assign accounts command?

PR #136, opened 2021-07-28
Add cost range

katherinelc321:
Adds the option to get cost from a date range.[ OSD-7822](https://issues.redhat.com/browse/OSD-7822)

### DIFF
--- a/cmd/cost/get.go
+++ b/cmd/cost/get.go
@@ -2,11 +2,12 @@ package cost
 
 import (
 	"fmt"
-	awsprovider "github.com/openshift/osdctl/pkg/provider/aws"
-	"github.com/spf13/cobra"
 	"log"
 	"strconv"
 	"time"
+
+	awsprovider "github.com/openshift/osdctl/pkg/provider/aws"
+	"github.com/spf13/cobra"
 
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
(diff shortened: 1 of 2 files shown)


drpaneas (approved):
/lgtm

cmd/cost/get.go
```diff
@@ -23,48 +24,46 @@ func newCmdGet(streams genericclioptions.IOStreams) *cobra.Command {
...
+	getCmd.Flags().StringVar(&ops.start, "start", "", "set start date range")
+	getCmd.Flags().StringVar(&ops.end, "end", "", "set end date range")
 	getCmd.Flags().BoolVar(&ops.csv, "csv", false, "output result as csv")
 
-	if err := getCmd.MarkFlagRequired("ou"); err != nil {
-		log.Fatalln("OU flag:", err)
+	return getCmd
+}
+
+func (o *getOptions) complete(cmd *cobra.Command, _ []string) error {
```
fahlmant, line 41: This function name is not very clear to me. Is there a reason its named complete and not something like "checkArgs"? Could you add some comments to this as well to explain what this function does and failure case?

cmd/cost/get.go
```diff
@@ -179,10 +205,20 @@ func getOUsRecursive(OU *organizations.OrganizationalUnit, awsClient awsprovider
 }
 
 //Get cost of given account
-func getAccountCost(accountID *string, unit *string, awsClient awsprovider.Client, timePtr *string, cost *float64) error {
+func (o *getOptions) getAccountCost(accountID *string, unit *string, awsClient awsprovider.Client, cost *float64) error {
+
+	var start, end, granularity string
+	if o.time != "" {
```
fahlmant, line 211: Is time mutually exclusive with start and end? If all three are set,  as far as I can read this, the start, end and granularity set here (lines 212-213) will be overwritten (lines 2127-219).

cmd/cost/list.go
```diff
@@ -45,11 +48,33 @@ func newCmdList(streams genericclioptions.IOStreams) *cobra.Command {
 	return listCmd
 }
 
+func (o *listOptions) complete(cmd *cobra.Command, _ []string) error {
```
fahlmant, line 51: Same questions here as the other `complete`

cmd/cost/list.go
```diff
@@ -61,7 +86,7 @@ func newListOptions(streams genericclioptions.IOStreams) *listOptions {
 }
 
 //List the cost of each OU under given OU
-func listCostsUnderOU(OU *organizations.OrganizationalUnit, awsClient awsprovider.Client, ops *listOptions) error {
+func (o *getOptions) listCostsUnderOU(OU *organizations.OrganizationalUnit, awsClient awsprovider.Client, ops *listOptions) error {
```
fahlmant, line 89: Did you mean to use ListOptions here?

cmd/cost/get.go
katherinelc321: Refactoring this to make it more clear

cmd/cost/get.go
katherinelc321: The function from before is a check that errors out if the 3 flags are provided. So yes time and start/end are mutually exclusive

cmd/cost/list.go
katherinelc321: the func listCostsUnderOU calls a func from get.go. I was wanting it to inherit the struct from getOptions. Refactoring this to make it more readable

dkeohane (approved):
Changes look good - do you need to add documentation to the README?

PR #137, opened 2021-07-31
Outputting info on stopped and running instances, updated cluster status to be a struct in the code.

aliceh:
This is the PR I wanted to submit.

### DIFF
--- a/cmd/cluster/health.go
+++ b/cmd/cluster/health.go
@@ -82,9 +82,19 @@ func (o *healthOptions) complete(cmd *cobra.Command, _ []string) error {
 func (o *healthOptions) run() error {
 
 	// This call gets the availability zone of the cluster, as well as the expected number of nodes.
-	az, clusterName, nodesExpected, err := ocmDescribe(o.k8sclusterresourcefactory.ClusterID)
+	az, clusterName, compute, infra, master, ascMin, ascMax, err := ocmDescribe(o.k8sclusterresourcefactory.ClusterID)
 	if az != nil {
-		fmt.Fprintf(o.IOStreams.Out, "\nThe expected number of nodes in availability zone(s) %s is : %v \n", az, nodesExpected)
+		fmt.Fprintf(o.IOStreams.Out, "\nThe expected number of nodes in availability zone(s) %s: ", az)
+		fmt.Fprintf(o.IOStreams.Out, "\nMaster: %v ", master)
+		fmt.Fprintf(o.IOStreams.Out, "\nInfra: %v ", infra)
+		if ascMin != 0 {
+			fmt.Fprintf(o.IOStreams.Out, "\nAutoscaled Compute: %v - %v ", ascMin, ascMax)
+		}
+		if compute != 0 {
+			fmt.Fprintf(o.IOStreams.Out, "\nCompute: %v ", compute)
+		}
+		fmt.Fprintf(o.IOStreams.Out, "\n \n")
+
 	}
 	if err != nil {
 		return err
@@ -137,14 +147,16 @@ func (o *healthOptions) run() error {
 
 	totalRunning := 0
 
-	//Here we count the number of customer's running instances in the cluster in the given region. To decide if the instance belongs to the cluster we are checking the Name Tag on the instance.
+	//Here we count the number of customer's running worker, infra and master instances in the cluster in the given region. To decide if the instance belongs to the cluster we are checking the Name Tag on the instance.
 	for idx := range instances.Reservations {
 		for _, inst := range instances.Reservations[idx].Instances {
 			tags := GetTags(inst)
 			for _, t := range tags {
 				if *t.Key == "Name" {
 					if strings.HasPrefix(*t.Value, clusterName) && *inst.State.Name == "running" {
-						totalRunning += 1
+						if strings.Contains(*t.Value, "worker") || strings.Contains(*t.Value, "infra") || strings.Contains(*t.Value, "master") {
+							totalRunning += 1
+						}
 					}
 				}
 			}
@@ -152,7 +164,7 @@ func (o *healthOptions) run() error {
 		}
 
 	}
-	fmt.Fprintf(o.IOStreams.Out, "\nThe number of running instances that belong to this cluster in region %s is : %v \n", reg, totalRunning)
+	fmt.Fprintf(o.IOStreams.Out, "\nThe number of running worker, infra and master instances that belong to this cluster in region %s is : %v \n", reg, totalRunning)
 
 	if err != nil {
 		log.Fatalf("Error getting instances %v", err)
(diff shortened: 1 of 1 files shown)


cmd/cluster/health.go
```diff
@@ -192,17 +228,31 @@ func ocmDescribe(clusterID string) ([]string, string, int, error) {
...
+	clusterInfo.Name = cluster.Name()
 
 	fmt.Printf("\nCluster %s - %s\n", cluster.Name(), cluster.ID())
 	fmt.Printf("\nCloud provider: %s\n", cloudProviderMessage)
 
 	if cloudProvider != "aws" {
-		return nil, "", 0, fmt.Errorf("This command is only supported for AWS clusters. The command is not supported for %s clusters.", cloudProviderMessage)
-
+		clusterInfo.AZs = nil
+		clusterInfo.Name = ""
```
fahlmant, line 238: You don't need to set all these (238-243). Go will fill them in with the respective type's default value.

mmazur:
This PR needs a rebase because there were changes to the code in the mean time, which added a more structured output that's more concise and easier to read. All the printfs will need to be replaced to match the new code.

mmazur:
This is the output I'd suggest for splitting master/infra/worker nodes):
```
ID: 1m2qrldphin9plr8c759r22j5toas6ng
Name: mmazur1
Provider: AWS
AZs:
- eu-central-1a
Nodes (master/infra/worker):
  Expected: 9 (3/2/4)
  Running: 9 (3/2/4)
```

cmd/cluster/health.go
aliceh: Thanks for the tip, @fahlmant!

aliceh:
@mmazur  I included in the output  the number of stopped nodes because the SL is different for when the nodes are stopped or the nodes are deleted. In addition, some clusters have autoscaled workers, so the expected number of nodes is a range. I made the output to look a lot like what you are suggesting.

cmd/cluster/health.go
```diff
@@ -221,18 +261,24 @@ func ocmDescribe(clusterID string) ([]string, string, int, error) {
...
 	cloudProvider := cluster.CloudProvider().ID()
 	cloudProviderMessage := strings.ToUpper(cloudProvider)
+	clusterInfo.Name = cluster.Name()
 
 	healthObject.ID = cluster.ID()
 	healthObject.Name = cluster.Name()
 	healthObject.Provider = cloudProviderMessage
+	healthObject.AZs = cluster.Nodes().AvailabilityZones()
+	healthObject.Expected.Infra = cluster.Nodes().Infra()
+	healthObject.Expected.Master = cluster.Nodes().Master()
```
iamkirkbater, line 272: Maybe I'm missing this but it's not clear to me where this is being used within this function.  Is this some kind of globally scoped variable `healthObject`?

cmd/cluster/health.go
```diff
@@ -221,18 +261,24 @@ func ocmDescribe(clusterID string) ([]string, string, int, error) {
...
 	if cloudProvider != "aws" {
-		return nil, "", 0, fmt.Errorf("This command is only supported for AWS clusters. The command is not supported for %s clusters.", cloudProviderMessage)
-
+		return clusterInfo, fmt.Errorf("This command is only supported for AWS clusters. The command is not supported for %s clusters.", cloudProviderMessage)
 	}
-	totalNodes := cluster.Nodes().Compute() + cluster.Nodes().Infra() + cluster.Nodes().Master()
+	clusterInfo.AZs = cluster.Nodes().AvailabilityZones()
+	clusterInfo.Nodes.Compute = cluster.Nodes().Compute()
+	clusterInfo.Nodes.AutoscaleMin = cluster.Nodes().AutoscaleCompute().MinReplicas()
+	clusterInfo.Nodes.AutoscaleMax = cluster.Nodes().AutoscaleCompute().MaxReplicas()
```
iamkirkbater, line 279: There seems to be some duplication between the `healthObject` and `clusterInfo` objects.  Is there a reason to have both separated and not just merge them into one?

cmd/cluster/health.go
```diff
@@ -177,15 +205,27 @@ func (o *healthOptions) run() error {
 	if err != nil {
 		log.Fatalf("error: %v", err)
 	}
+	fmt.Fprintf(o.IOStreams.Out, "\n \n")
```
iamkirkbater, line 208: I don't think we need this.

cmd/cluster/health.go
```diff
@@ -177,15 +205,27 @@ func (o *healthOptions) run() error {
...
 	fmt.Printf(string(d))
```
iamkirkbater, line 209: I know this wasn't changed here, but can we use a more descriptive name for the health output string than `d`?  Maybe something like `healthOutput`?

cmd/cluster/health.go
iamkirkbater: Ahh, I see where this is coming from now.

I think we should take _out_ the struct fields from the `healthOptions` struct, and load all of them into the `clusterInfo` struct.  The `healthOptions` struct is more for options being passed into the command from the end-user, and I don't think overloading that struct and storing all of the various data about a cluster in there is the best choice when we can just push all of that data into the `clusterInfo` struct instead.

iamkirkbater (changes requested):
A few relatively minor comments.  Great work so far! I'm excited to see this get done!

cmd/cluster/health.go
aliceh: Done.

cmd/cluster/health.go
aliceh: This line adds an empty line before writing the concise output, it makes the output look better IMO.  I get these Throttling request messages before the output comes, and without an empty line to separate the two, it wasn't easy to see where the command output starts.

cmd/cluster/health.go
```diff
@@ -22,6 +23,7 @@ import (
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 )
 
+//This command requires the ocm API Token https://cloud.redhat.com/openshift/token be available in the OCM_TOKEN env variable.
 // healthOptions defines the struct for running health command
```
georgettica, line 27: this sort-of breaks the godoc cormatting
https://go.dev/blog/godoc
if this can be formatted a bit differently that would be great
```suggestion
// healthOptions defines the struct for running health command
// This command requires the ocm API Token https://cloud.redhat.com/openshift/token be available in the OCM_TOKEN env variable.

```

georgettica (commented):
it seems some of @iamkirkbater'c comments have not been addressed.

just did a grooming session of the PR's and found this was not addressed for a while

let me know if I can further

cmd/cluster/health.go
georgettica: the newline is my bad..
```
go doc -all cmd/cluster
```
won't display the right info with the newline there

you can test this on you machine or I can show you a quick `main.go` of how `go doc` behaves

cmd/cluster/health.go
aliceh: clusterInfo is just a variable holding the output of ocmDescribe(o.k8sclusterresourcefactory.ClusterID) command and it needs to be separate from healthObject which is what we output because the raw input needs a bunch of preprocessing, the clsuetrStats object is the one that can be merged with ClusterHealthCondensedObject. I cleaned up the code.

aliceh: I removed the clusterInfo variable since it wasn't really necessary to introduce it.

PR #139, opened 2021-08-18
Adding reseting to OU functionality

lnguyen1401:


cmd/account/reset.go
```diff
@@ -16,9 +16,17 @@ import (
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/organizations"
 	"github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	"github.com/openshift/osdctl/cmd/common"
 	"github.com/openshift/osdctl/pkg/k8s"
+	awsprovider "github.com/openshift/osdctl/pkg/provider/aws"
+)
+
+var (
+	OSDStaging2RootID = "r-rs3h"
```
fahlmant, line 28: Is there a way to get these from the AWS API instead?

cmd/account/reset.go
```diff
@@ -127,7 +139,24 @@ func (o *resetOptions) run() error {
...
-
+	var rootId string
+	if o.payerAccount == "osd-staging-1" {
+		rootId = OSDStaging1RootID
+	} else if o.payerAccount == "osd-staging-2" {
+		rootId = OSDStaging2RootID
+	} else {
+		return fmt.Errorf("invalid payer account provided")
+	}
+	if o.resetLegalEntity {
```
fahlmant, line 150: This logic block is already present at line 139. Please combine them

cmd/account/reset.go
```diff
@@ -127,7 +139,24 @@ func (o *resetOptions) run() error {
...
 		account.Spec.LegalEntity = v1alpha1.LegalEntity{}
 	}
-
+	var rootId string
+	if o.payerAccount == "osd-staging-1" {
+		rootId = OSDStaging1RootID
+	} else if o.payerAccount == "osd-staging-2" {
+		rootId = OSDStaging2RootID
+	} else {
+		return fmt.Errorf("invalid payer account provided")
```
dkeohane, line 148: Given that we explicitly only support 2 payer account values, it would be more user friendly to include that in the error message. I might suggest something along the lines of:
`invalid payer account provided. Currently supported: osd-staging-1, osd-staging-2`

lnguyen1401: yeah that's a great way to do it!

fahlmant: The payer account should be known from the hive, so I actually don't think we need this check

cmd/account/reset.go
```diff
@@ -51,7 +55,7 @@ func newCmdReset(streams genericclioptions.IOStreams, flags *genericclioptions.C
 type resetOptions struct {
 	accountName      string
 	accountNamespace string
-
+	awsClient        awsprovider.Client
```
fahlmant, line 58: Let's remove this from the Options struct since its not set by a cli option

cmd/account/reset.go
```diff
@@ -114,25 +121,46 @@ func (o *resetOptions) run() error {
...
+	//	payerAccount = account.Spec.IAMUserSecret
 	// reset fields in spec
-	account.Spec.ClaimLink = ""
-	account.Spec.ClaimLinkNamespace = ""
-	account.Spec.IAMUserSecret = ""
+	//	account.Spec.ClaimLink = ""
+	//	account.Spec.ClaimLinkNamespace = ""
+	//	account.Spec.IAMUserSecret = ""
+	//awsClient, err := awsprovider.NewAwsClient("osd-staging-2", "us-east-1", "")
+	awsClient, err := o.getAwsClient("aws-account-operator-credentials", "aws-account-operator")
```
fahlmant, line 137: Can you either move the login into this function, or rename the function to match what's happening (getAwsAccountOperatorSecret/Client). I also think using an &Client would be cleaner.

fahlmant: Here are my thoughts.
1. Get the Secret from the namespace
2. get the Access key and Secret access key from the secret data
3. Call awsprovider.NewAwsClientWithInput similar to https://github.com/openshift/osdctl/blob/e5b4ae5c11a245ebed83ee152bfc35ea9b0fcc29/cmd/account/clean-velero-snapshots.go#L78

cmd/account/reset.go
```diff
@@ -114,25 +121,46 @@ func (o *resetOptions) run() error {
...
+	o.awsClient = awsClient
+	if err != nil {
+		return err
+	}
+	var rootId string
+	rootId, err = o.getRoot(o.awsClient)
```
fahlmant, line 143: This function is both called as a method of resetOptions, and the member var awsClient is being passed. Given the request above, I think this can be reimplemented to be
`rootId, err :=  getRoot(&awsClient)`

cmd/account/reset.go
```diff
@@ -153,3 +181,81 @@ func (o *resetOptions) run() error {
...
+	secretToken := &corev1.Secret{}
+	err1 := o.kubeCli.Get(context.TODO(),
+		types.NamespacedName{
+			Name:      "aws-account-operator-client-token-t4nts",
+			Namespace: nameSpace,
+		},
+		secretToken)
+	if err1 != nil {
+		return nil, err1
+	}
```
fahlmant, line 240: No need to get the kube token

cmd/account/reset.go
```diff
@@ -153,3 +181,81 @@ func (o *resetOptions) run() error {
...
+	accessKeyID := base64.StdEncoding.EncodeToString([]byte(secretAws.Data["aws_access_key_id"]))
+	fmt.Printf("encoded secret is %+v\n", accessKeyID)
+
+	secretkeyID := base64.StdEncoding.EncodeToString([]byte(secretAws.Data["aws_secret_access_key"]))
+	fmt.Printf("encoded secret access key ID is %+v\n", secretkeyID)
+	//token := base64.StdEncoding.EncodeToString([]byte(secretToken.Data["token"]))
+	fmt.Printf("token is  %+v\n", token)
+	var awsInput awsprovider.AwsClientInput
+	awsInput.AccessKeyID = string(accessKeyID)
```
fahlmant, line 249: See above comment about making this a bit more succinct

cmd/account/reset.go
```diff
@@ -142,14 +173,38 @@ func (o *resetOptions) run() error {
...
+	})
 	if o.resetLegalEntity {
 		status["reused"] = false
 	}
+	return o.kubeCli.Status().Patch(ctx, account, client.RawPatch(types.MergePatchType, mergePatch))
+}
 
-	mergePatch, _ = json.Marshal(map[string]interface{}{
-		"status": status,
+func (o *resetOptions) getAwsClient(secretName string, nameSpace string) (awsprovider.Client, error) {
```
fahlmant, line 185:
```suggestion
func (o *resetOptions) getAwsClientFromSecret(secretName string, nameSpace string) (awsprovider.Client, error) {
```

cmd/account/reset.go
```diff
@@ -142,14 +173,38 @@ func (o *resetOptions) run() error {
...
+		types.NamespacedName{
+			Name:      secretName,
+			Namespace: nameSpace,
+		},
+		secretAws)
+	if err != nil {
+		return nil, err
+	}
+	accessId := base64.StdEncoding.EncodeToString([]byte(secretAws.Data["aws_access_key_id"]))
+	accessKeyID, _ := base64.StdEncoding.DecodeString(accessId)
```
fahlmant, line 197: Please handle the error

cmd/account/reset.go
```diff
@@ -142,14 +173,38 @@ func (o *resetOptions) run() error {
...
+	secretId := base64.StdEncoding.EncodeToString([]byte(secretAws.Data["aws_secret_access_key"]))
+	secretkeyID, _ := base64.StdEncoding.DecodeString(secretId)
```
fahlmant, line 199: Please handle the error

cmd/account/reset.go
```diff
@@ -142,14 +176,46 @@ func (o *resetOptions) run() error {
...
 	}
-
 	mergePatch, _ = json.Marshal(map[string]interface{}{
 		"status": status,
 	})
-
 	return o.kubeCli.Status().Patch(ctx, account, client.RawPatch(types.MergePatchType, mergePatch))
 }
+
+func (o *resetOptions) getAwsClientFromSecret(secretName string, nameSpace string) (awsprovider.Client, error) {
```
NautiluX, line 188: `namespace`

cmd/account/reset.go
```diff
@@ -107,32 +110,63 @@ func (o *resetOptions) run() error {
...
+			return err
+		}
+
 		account.Spec.LegalEntity = v1alpha1.LegalEntity{}
+		//get the id of the current OU that the account is living in
+		parent, err := awsClient.ListParents(&organizations.ListParentsInput{
+			ChildId: aws.String(accountId),
+		})
+		parentId := *parent.Parents[0].Id
+		if err != nil {
```
NautiluX, line 151: Shouldn't we first check for errors since otherwise this call may segfault if `parent.Parents` is empty?

cmd/account/reset.go
```diff
@@ -107,32 +110,63 @@ func (o *resetOptions) run() error {
...
 	if o.resetLegalEntity {
+		awsClient, err := o.getAwsClientFromSecret("aws-account-operator-credentials", "aws-account-operator")
+		if err != nil {
+			return err
+		}
+		//awsClient := *AwsClient
+		//get the rootOU from the awsClient above
+		root, err := awsClient.ListRoots(&organizations.ListRootsInput{})
+		rootId := *root.Roots[0].Id
+		if err != nil {
```
NautiluX, line 141: Shouldn't we first check for errors since otherwise this call may segfault if `root.Roots` is empty?

cmd/account/reset.go
```diff
@@ -107,32 +110,63 @@ func (o *resetOptions) run() error {
...
+			return err
+		}
+
 		account.Spec.LegalEntity = v1alpha1.LegalEntity{}
+		//get the id of the current OU that the account is living in
```
NautiluX, line 146: Is there a reason for this line
```
account.Spec.LegalEntity = v1alpha1.LegalEntity{}
```
to be at exactly this position? It seems to be a bit hidden between two blocks of talking to AWS, and the update call only happens after the second block.

cmd/account/reset.go
```diff
@@ -107,32 +110,63 @@ func (o *resetOptions) run() error {
...
 	account.Spec.ClaimLink = ""
 	account.Spec.ClaimLinkNamespace = ""
 	account.Spec.IAMUserSecret = ""
+	//create an awsClient from the credentials in aws-account-operator-credentials
 	if o.resetLegalEntity {
+		awsClient, err := o.getAwsClientFromSecret("aws-account-operator-credentials", "aws-account-operator")
+		if err != nil {
+			return err
+		}
+		//awsClient := *AwsClient
```
NautiluX, line 137: Remove if not needed?

NautiluX (changes requested):

cmd/account/reset.go
```diff
@@ -114,25 +119,44 @@ func (o *resetOptions) run() error {
...
+	}
+	//get the rootOU from the awsClient above
+	rootId, err := o.getRoot(&awsClient)
+	if err != nil {
+		return err
+	}
 	if o.resetLegalEntity {
 		account.Spec.LegalEntity = v1alpha1.LegalEntity{}
+		//get the id of the current OU that the account is living in
+		parentId, err := o.getParent(&awsClient, accountId)
```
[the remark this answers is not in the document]
NautiluX: I was really close to suggesting to split ListParents out into its own function :D
Curious why you think it's better to call the API directly in here?

cmd/account/reset.go
fahlmant: +1, this should be moved down to the Update of the CR.

fahlmant (approved):
/lgtm

PR #140, opened 2021-08-25
feat(build): make goreleaser work on macos aswell

georgettica:


clcollins:
@georgettica This looks great to me, but I'm not sure why the go.mod and go.sum were updated for a Makefile change.  Can you drop those changes, and we can merge this?

georgettica:
I am not here really, but I will do that next week

PR #143, opened 2021-08-27
Remove all account access in unassign

katherinelc321:
https://issues.redhat.com/browse/OSD-8113

### DIFF
--- a/cmd/account/mgmt/account-unassign.go
+++ b/cmd/account/mgmt/account-unassign.go
@@ -48,6 +48,17 @@ type accountUnassignOptions struct {
 	genericclioptions.IOStreams
 }
 
+type accountCleanOptions interface {
+	deleteLoginProfile() error
+	deleteAccessKeys(user string) error
+	deleteSigningCert(user string) error
+	deletePolicies(user string) error
+	deleteAttachedPolicies(user string) error
+	deleteRoles(newAWSClient awsprovider.Client) error
+	deleteGroups(user string) error
+	deleteUser(user string) error
+}
+
 func newAccountUnassignOptions(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *accountUnassignOptions {
 	return &accountUnassignOptions{
 		flags:      flags,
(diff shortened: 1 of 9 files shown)


cmd/account/mgmt/account-unassign.go
```diff
@@ -128,57 +132,119 @@ func (o *accountUnassignOptions) run() error {
...
 	}
 
-	// Delete login profile
-	err = o.deleteLoginProfile(accountUsername)
-	if err != nil {
-		return err
+	for _, userName := range allUsers {
+
+		// Delete login profile
+		err = o.deleteLoginProfile(userName)
```
georgettica, line 169: This could be simplified in a for loop that takes the fun too s as a param.and runs for each runs the check
This way was done in https://github.com/openshift/gcp-project-operator/blob/58ea50451037cda552783adc4b17b1300eba376c/pkg/controller/projectclaim/projectclaim_controller.go#L117

cmd/account/mgmt/account-unassign.go
```diff
@@ -128,57 +132,119 @@ func (o *accountUnassignOptions) run() error {
...
 	}
-	// Delete policies
-	err = o.deletePolicies(accountUsername)
-	if err != nil {
-		return err
+	newAwsClientInput := &awsprovider.AwsClientInput{
+		AccessKeyID:     *result.Credentials.AccessKeyId,
+		SecretAccessKey: *result.Credentials.SecretAccessKey,
+		SessionToken:    *result.Credentials.SessionToken,
+		Region:          "us-east-1",
```
georgettica, line 223: Can this be a variable from the default region used? I. Am not sure this will work in all cases (feel free to correct me if I am off base)

georgettica (commented):
Not a qualified aws persona, so just did a code skim and checked for nits, I hope this review is helpful for you

cmd/account/mgmt/account-unassign.go
fahlmant: IAM resources are global, and not locked to a region. The AWS client needs a region though, so us-east-1 is generally "default".

georgettica: resolving as this is is a "aws api issue"

PR #145, opened 2021-09-03
Remove accidentally committed osdctl binary

fahlmant:


georgettica:
Also add it to the gitignore?

PR #148, opened 2021-09-12
chore(makefile): align mockgen versions

georgettica:
this caused a 'go-get' to be called and change the go.mod stuff

this change also changed some generated files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ mockgen: ensure-mockgen
 	@git diff --exit-code -- ./pkg/provider/aws/mock
 
 ensure-mockgen:
-	go get github.com/golang/mock/mockgen@v1.4.4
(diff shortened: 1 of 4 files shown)


georgettica:
the issue looks as follows:
```
go get github.com/golang/mock/mockgen@v1.4.4
go get: installing executables with 'go get' in module mode is deprecated.
	To adjust and download dependencies of the current module, use 'go get -d'.
	To install using requirements of the current module, use 'go install'.
	To install ignoring the current module, use 'go install' with a version,
	like 'go install example.com/cmd@latest'.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
go get: downgraded github.com/golang/mock v1.5.0 => v1.4.4
go get: removed github.com/openshift/gcp-project-operator v0.0.0-20210818135501-58ea50451037
```

PR #149, opened 2021-09-13
Refactor unassign cmd to delete all aws access

katherinelc321:
https://issues.redhat.com/browse/OSD-8113


cmd/account/mgmt/account-unassign.go
```diff
@@ -131,54 +127,124 @@ func (o *accountUnassignOptions) run() error {
...
+		if err != nil {
+			fmt.Println(err)
+		}
+		// Delete signing certificates
+		err = o.deleteSigningCert(userName)
+		if err != nil {
+			fmt.Println(err)
+		}
+		// Delete policies
+		err = o.deletePolicies(userName)
```
iamkirkbater, line 179: I'm still not convinced that we need the username for these parameters.  We should be deleting _all_ user created policies, all user created login profiles, all users, all user-created roles, etc.  So each one of these functions would be something like:

```
func o.deleteUsers() error {
    all_users := aws.ListUsers()
    for user in all_users {
        aws.DeleteUser(user.Arn)
    }
}

func o.deleteRoles() error {
    all_roles := aws.ListRoles()
    for role in all_roles {
        if role.Name == "OrganizationAccountAccessRole" {
            continue
        }
        aws.DeleteRole(role.Arn)
    }
}
```

So each of these functions would be a self-contained "list all" and then "remove if not managed by AWS".

cmd/account/mgmt/account-unassign.go
```diff
@@ -373,6 +435,27 @@ func (o *accountUnassignOptions) deleteAttachedPolicies(user string) error {
...
+
+	listRoleInput := &iam.ListRolesInput{}
+	roles, err := newAWSClient.ListRoles(listRoleInput)
+	if err != nil {
+		return err
+	}
+	// delete all roles except OrganizationAccountAccessRole
+	for _, rolename := range roles.Roles {
+
+		if rolename.RoleName != aws.String("OrganizationAccountAccessRole") {
```
iamkirkbater, line 448: This function looks like it's exactly what we are looking for (relating to my previous comment).

However, this specific line I don't think will work the way it's intended, because we're trying to compare pointers.  We'd need to dereference `rolename.RoleName` and compare it to the primitive `"OrganizationAccountAccessRole"` and not what's returned from `aws.String()`.

cmd/account/mgmt/account-unassign_test.go
```diff
@@ -15,6 +16,69 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+func TestAssumeRoleForAccount(t *testing.T) {
```
iamkirkbater, line 19: Tests <3

iamkirkbater (changes requested):
Overall looks like this is very-much headed in the right direction.  Keep up the great work!

cmd/account/mgmt/account-unassign.go
fahlmant: Also, there are other roles that AWS creates that we should NOT delete.

cmd/account/mgmt/account-unassign.go
katherinelc321: all the delete operations are inside a loop that iterates through the list of IAM users created by the user we wanna clean. Instead of doing listUsers -> delete -> listUsers -> delete ...... i did for each user in listUsers -> { perform all the deletions}

cmd/account/mgmt/account-unassign.go
katherinelc321: my understanding is that since we instantiated a new AWS client with AssumeRole creds, when we perfom ListRoles, it should only list the roles created by whatever account id we provided to the assumeRoleForAccount func

fahlmant: Every account (unique account ID) gets a set of roles created by AWS. For example

```
$ aws sts get-caller-identity
{
    "UserId": "AROAU4GZCQTXJOWX2WGWB:tfahlman-sre",
    "Account": "335462696174",
    "Arn": "arn:aws:sts::335462696174:assumed-role/OrganizationAccountAccessRole/tfahlman-sre"
}
```

```
$ aws iam list-roles | jq -r '.Roles[].RoleName'

AccessAnalyzerMonitorServiceRole_SSQKKDWYPI
AWSServiceRoleForAWSLicenseManagerRole
AWSServiceRoleForElastiCache
AWSServiceRoleForElasticLoadBalancing
AWSServiceRoleForGlobalAccelerator
AWSServiceRoleForMarketplaceLicenseManagement
AWSServiceRoleForNetworkFirewall
AWSServiceRoleForOrganizations
AWSServiceRoleForRDS
AWSServiceRoleForSupport
AWSServiceRoleForTrustedAdvisor
BYOCAdminAccess-472x92
BYOCAdminAccess-7f5bbr
BYOCAdminAccess-hwq9l7
BYOCAdminAccess-p7t45f
BYOCAdminAccess-pgt46k
BYOCAdminAccess-r7jr2w
BYOCAdminAccess-sj29s4
BYOCAdminAccess-x7m5gf
customperms-ControlPlane-Role
customperms-Installer-Role
customperms-openshift-cloud-credential-operator-cloud-credential
customperms-openshift-cluster-csi-drivers-ebs-cloud-credentials
customperms-openshift-image-registry-installer-cloud-credentials
customperms-openshift-ingress-operator-cloud-credentials
customperms-openshift-machine-api-aws-cloud-credentials
customperms-Support-Role
customperms-Worker-Role
ManagedOpenShift-ControlPlane-Role
ManagedOpenShift-IAM-Role
ManagedOpenShift-Installer-Role
ManagedOpenShift-openshift-cloud-credential-operator-cloud-crede
ManagedOpenShift-openshift-cluster-csi-drivers-ebs-cloud-credent
ManagedOpenShift-openshift-image-registry-installer-cloud-creden
ManagedOpenShift-openshift-ingress-operator-cloud-credentials
ManagedOpenShift-openshift-machine-api-aws-cloud-credentials
ManagedOpenShift-Support-Role
ManagedOpenShift-Worker-Role
OrganizationAccountAccessRole
rds-monitoring-role
```
That includes both roles that have been user created (by me, and the rosa cli) as well as created by AWS by default when the account was first created. Osdctl should ONLY delete the user created roles

cmd/account/mgmt/account-unassign.go
iamkirkbater: Ooh, I see, this is deleting all of the policies that are created as UserPolicies, but I don't think this also deletes standalone policies.

So in the AWS console, if you go to the Users page and create a policy under a user it will attach inline policies to that user.  These are UserPolicies.  If you go into the Policies page and create a new Policy, this is just a Policy, I believe.

Either way, I'll have to do more research to make sure that my assumption above is correct, because my brain tends to instantly brain-dump IAM stuff as soon as I don't directly need it anymore, I'm sure that's a coping mechanism 😂. With that said, I'll give this PR a test as is and we can clean up lingering IAM standalone policies if they're not getting cleaned up as a separate body of work.

cmd/account/mgmt/account-unassign.go
```diff
@@ -131,54 +127,124 @@ func (o *accountUnassignOptions) run() error {
...
-	err = o.deleteAccessKeys(accountUsername)
-	if err != nil {
-		fmt.Println(err)
+
+	return nil
+}
+
+func (o *accountUnassignOptions) assumeRoleForAccount(account_id string) (awsprovider.Client, error) {
+
+	roleArn := fmt.Sprintf("arn:aws:%s:iam::role/OrganizationAccountAccessRole", account_id)
```
iamkirkbater, line 205: This ARN is not valid.

It should be `arn:aws:iam::%s/role/OrganizationAccountAccessRole`

PR #151, opened 2021-10-15
Use the new `SupportRoleArn` field

drpaneas:
Update osdctl to use the new `SupportRoleArn` field for all account access requests that are via a cluster (basically any osdctl account [cli|console] call that doesn't use the -i [aws account id] flag.

https://issues.redhat.com/browse/OSD-8367

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,10 +15,15 @@ fmt:
 	@gofmt -w -s .
 	@git diff --exit-code .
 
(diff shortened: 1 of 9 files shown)


fahlmant (commented):
Will the support role always be named ManagedOpenshift-Support? As I understand it there's now a field in the AccountClaim called supportRoleArn. Should this rely on that value instead?

iamkirkbater:
Closing in favor of #175.

PR #152, opened 2021-10-17
make mockgen local

georgettica:
as there was an issue with `goreleaser` releases, I moved to using the local pull image instead

- use full path to ./bin so mockgen-works
- make goreleaser local aswell (now not pulling from releases, but
installs)
- force not using global goreleaser: the PATH now forces using the local
  binaries and not the global ones, unless you delete them

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,10 @@ unexport GOFLAGS
 # Add OSDCTL `./bin/` dir to path for goreleaser
 # This will look for goreleaser first in the local bin dir
 # but otherwise can be installed elsewhere on developers machines
-export PATH := bin:$(PATH)
+BASE_DIR=$(shell pwd)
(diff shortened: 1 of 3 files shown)


georgettica:
seems the CI assumes I have a lower version of go?
should we bump them to use a newer one?

georgettica:
build will be fixed in https://github.com/openshift/osdctl/pull/156

seems the `release` is only a part of a go version bump

georgettica:
odd, it shows there are no prow stuff https://prow.ci.openshift.org/pr?query=is%3Apr%20repo%3Aopenshift%2Fosdctl%20author%3Ageorgettica%20head%3Ageorgettica%2Fmake-mockgen-local

georgettica:
ready to go @fahlmant 
squashed it all :)

PR #158, opened 2021-11-03
chore(docs): run `osdctl docs ./docs/command/`

georgettica:
this is not run per-commit, and as such has a large drift to our current
setup.

should we use these docs? or remove them altogether


fahlmant (approved):
/lgtm

fahlmant:
Not sure if these are useful, happy to remove them too

PR #159, opened 2021-11-03
Update cli.go

georgettica:
this way you can run the command
```
export $(osdctl account cli -C ${CLUSTERID} --out env)
```

and the command will work


fahlmant (approved):
/lgtm

PR #162, opened 2021-11-12
fix output

katherinelc321:
Some cmds are redefining the global output flag locally and this is causing osdctl to panic. This PR fixes the issue


fahlmant (commented):
@katherinelc321 Should this be closed in favor of #166?

PR #165, opened 2021-11-19
Dependency inject k8s client

NautiluX:
* Create client less often
* Dependency inject it to commands
* Replace real client with mock in tests


fahlmant (approved):
/approve

PR #166, opened 2021-11-19
Fix output

iamkirkbater:

PR #170, opened 2021-12-13
Remove unused PersistentFlag

NautiluX:
Removing an unused PersistentFlag, that is overridden by at least one subcommand, resulting in the following error:
```
$ aws account cli

panic: unable to redefine 'o' shorthand in "cli" flagset: it's already used for "out" flag
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -33,8 +33,6 @@ var GitCommit string
 // Version is the tag version from the environment
 var Version string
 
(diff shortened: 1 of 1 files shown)


georgettica:
#166 solves this is a general pattern

NautiluX:
/close
this problem seems to be fixed, but it seems -o is used somewhere else, will create another PR.

PR #171, opened 2021-12-14
Lazy-initialized k8s client

NautiluX:
* Allow skipping client creation for commands that don't need it
* Client will throw an error if it cannot be initialized on first use

### DIFF
--- a/pkg/k8s/client.go
+++ b/pkg/k8s/client.go
@@ -3,66 +3,75 @@ package k8s
 import (
 	"context"
 	"fmt"
-	"log"
 
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


pkg/k8s/client.go
```diff
@@ -3,66 +3,75 @@ package k8s
...
 	}
 
-	cli, err := client.New(cfg, client.Options{})
+	s.client, err = client.New(cfg, client.Options{})
 	if err != nil {
-		log.Printf("Can't load KubeConfig, using stub client: %v", err)
-		return &StubClient{}
+		panic(s.err())
 	}
-	return cli
```
georgettica, line 67: i don't like this (the `panic`) but i'll live as this is a good solution to the problem :)

PR #174, opened 2021-12-16
OSD-9325: add clusterID to servicelog post payload

ramonbutter:
This PR removes the requirement to specify the Cluster_UUID as a -p parameter and the ${CLUSTER_UUID} definition in the template. Instead, the cluster identifier (UUID, ID, or display Name) is passed directly after the post. Based on this osdctl adds the values that are required by the ocm API. 

E.g.: 
To send one service log use 
```
  osdctl servicelog post clusterID -t github/path/to/template -p arg1="foo"
```
To send SL to several clusters:
```
  osdctl servicelog post {clusterID1, clusterID2} -t github/path/to/template -p arg1="foo"
```
It also further supports the definition of a query list and cluster list file.

### DIFF
--- a/cmd/servicelog/post.go
+++ b/cmd/servicelog/post.go
@@ -373,6 +373,10 @@ func createPostRequest(ocmClient *sdk.Connection, cluster *v1.Cluster) (request 
 
 	Message.ClusterUUID = cluster.ExternalID()
 	Message.ClusterID = cluster.ID()
+	if subscription := cluster.Subscription(); subscription != nil {
+		Message.SubscriptionID = cluster.Subscription().ID()
+	}
+
 	messageBytes, err := json.Marshal(Message)
 	if err != nil {
 		return nil, fmt.Errorf("cannot marshal template to json: %v", err)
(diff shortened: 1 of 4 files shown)


cmd/servicelog/post.go
```diff
@@ -51,6 +51,22 @@ var postCmd = &cobra.Command{
...
+		if len(args) < 1 && filtersFromFile == "" && clustersFile == "" {
+			log.Fatalf("No cluster identifier has been found.")
+		}
+
+		var queries []string
+		for _, clusterIds := range args {
+			queries = append(queries, generateQuery(clusterIds))
+		}
+
+		if len(queries) != 0 {
```
georgettica, line 63:
```suggestion
		if len(queries) > 0 {
```

cmd/servicelog/post.go
```diff
@@ -51,6 +51,22 @@ var postCmd = &cobra.Command{
...
+			if len(filterParams) != 0 {
```
georgettica, line 64:
```suggestion
			if len(filterParams) > 0 {
```

georgettica:
does that change mean we can remove the `CLUSTER_UUID` field from the servicelog json files?

georgettica:
/approve
/lgtm
/hold

looks good and seems thought out

NautiluX (commented):
/lgtm

PR #175, opened 2021-12-17
Updates references from BYOCAdminAccess to ManagedOpenShift-Support

iamkirkbater:
Resolves [OSD-8367](https://issues.redhat.com/browse/OSD-8367)


dkeohane (approved):
/lgtm

iamkirkbater:
@fahlmant To carry over the discussion on the PR that was closed - https://github.com/openshift/osdctl/pull/151#pullrequestreview-825374351

I think the added cost of looking up the accountclaim (and thus having to be tunneled and logged into a hive) outweighs the cost of updating this command if we change that field again.  I'm not sure if we'll care about STS clusters with the commands that are functionally effected (and not just updating the error log messages), though we can always circle back and make this more dynamic if our needs change.

PR #177, opened 2021-12-22
fix for sending servicelog to cluster by using a query

ramonbutter:


georgettica:
could we merge it with https://github.com/openshift/osdctl/pull/178/files ?

PR #178, opened 2021-12-22
fix(servicelog): bulk commands

georgettica:
now servicelogs can be sent and listed in bulk

### DIFF
--- a/cmd/servicelog/post.go
+++ b/cmd/servicelog/post.go
@@ -108,7 +108,7 @@ var postCmd = &cobra.Command{
 				cluster := ClustersFile.Clusters[i]
 				query = append(query, generateQuery(cluster))
 			}
-			filterParams = query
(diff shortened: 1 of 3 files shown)


cmd/servicelog/post.go
```diff
@@ -51,7 +51,7 @@ var postCmd = &cobra.Command{
 		readFilterFile()      // parse the ocm filters in file provided via '-f' flag
 		readTemplate()        // parse the given JSON template provided via '-t' flag
 
-		if len(args) < 1 && filtersFromFile == "" && clustersFile == "" {
+		if len(args) == 0 && len(filterParams) == 0 && clustersFile == "" {
```
ramonbutter, line 54: filterParams are the query that we use in ocm. The -q parameter directly fills that

ramonbutter (approved):
/lgtm

PR #179, opened 2022-01-09
Ensure kubecli is set for packet-capture cmd

mrbarge:
This PR addresses a `panic` that I observed when attempting to run an `osdctl network packet-capture` call. The `kubeCli` variable was not being set and hence was a nil pointer when first used.

Upon further investigation it was observed that it was being ignored in the `newPacketCaptureOptions` function. This PR ensures that it is set as part of that function call.

Successfully tested on a cluster to ensure that the command now behaves appropriately.


iamkirkbater:
/retest

Just want to make sure this is still good and we can merge it.

dastergon:
@iamkirkbater @mrbarge anything else missing on the PR?

mrbarge:
Should still be good to go, I think. Thanks for checking!

PR #180, opened 2022-01-10
Adds new flags to reset in order to "shred" CRs

iamkirkbater:
This PR adds two new flags to the reset command to be used together in order to create a workflow that allows us to "shred" accounts on an ad-hoc basis.


cmd/account/reset.go
```diff
@@ -44,19 +44,31 @@ func newCmdReset(streams genericclioptions.IOStreams, flags *genericclioptions.C
 		"Skip the prompt check")
 	resetCmd.Flags().BoolVar(&ops.resetLegalEntity, "reset-legalentity", false,
 		`This will wipe the legalEntity, claimLink and reused fields, allowing accounts to be used for different Legal Entities.`)
+	resetCmd.Flags().BoolVarP(&ops.editState, "edit-state", "e", false,
+		"Change only the state of an account.")
+	resetCmd.Flags().StringVar(&ops.stateInput, "state", "",
+		"Select the state to change to. Required when using the --edit-state flag. Valid values are Ready or Failed")
```
NautiluX, line 50: Why do we need 2 parameters here? Wouldn't be setting a string enough to know we should set the state? (e.g. `-e failed`)

cmd/account/reset.go
```diff
@@ -75,6 +87,23 @@ func (o *resetOptions) complete(cmd *cobra.Command, args []string) error {
...
+
+		switch stateIn {
+		case "ready":
+			o.setState = stateReady
+		case "failed":
+			o.setState = stateFailed
+		default:
+			return cmdutil.UsageErrorf(cmd, "Unsupported state.  Supported states are Ready or Failed")
+		}
+	}
```
NautiluX, line 105: Why do we need to check this twice? Can't we move this to the run function so the `if o.editState` is only done once? Translating the string to state could be a function called from there.

NautiluX (changes requested):

PR #184, opened 2022-01-20
OSD-9701 Adding per Account breakdown to Cost command 

dkeohane:
Adding functionality to identify costs per account in an OU, instead of per OU

### DIFF
--- a/cmd/cost/get.go
+++ b/cmd/cost/get.go
@@ -32,7 +32,7 @@ func newCmdGet(streams genericclioptions.IOStreams, globalOpts *globalflags.Glob
 	}
 	getCmd.Flags().StringVar(&ops.ou, "ou", "", "get OU ID")
 	getCmd.Flags().BoolVarP(&ops.recursive, "recursive", "r", false, "recurse through OUs")
-	getCmd.Flags().StringVarP(&ops.time, "time", "t", "", "set time. One of 'LM', 'MTD', 'TYD', '3M', '6M', '1Y'")
+	getCmd.Flags().StringVarP(&ops.time, "time", "t", "", "set time. One of 'LM', 'MTD', 'YTD', '3M', '6M', '1Y'")
 	getCmd.Flags().StringVar(&ops.start, "start", "", "set start date range")
 	getCmd.Flags().StringVar(&ops.end, "end", "", "set end date range")
 	getCmd.Flags().BoolVar(&ops.csv, "csv", false, "output result as csv")
--- a/cmd/cost/get.go
+++ b/cmd/cost/get.go
@@ -31,12 +31,13 @@ func newCmdGet(streams genericclioptions.IOStreams, globalOpts *globalflags.Glob
 			cmdutil.CheckErr(ops.run())
 		},
 	}
-	getCmd.Flags().StringVar(&ops.ou, "ou", "", "get OU ID")
+	getCmd.Flags().StringVar(&ops.ou, "ou", "", "set OU ID")
 	getCmd.Flags().BoolVarP(&ops.recursive, "recursive", "r", false, "recurse through OUs")
 	getCmd.Flags().StringVarP(&ops.time, "time", "t", "", "set time. One of 'LM', 'MTD', 'YTD', '3M', '6M', '1Y'")
 	getCmd.Flags().StringVar(&ops.start, "start", "", "set start date range")
 	getCmd.Flags().StringVar(&ops.end, "end", "", "set end date range")
 	getCmd.Flags().BoolVar(&ops.csv, "csv", false, "output result as csv")
+	getCmd.Flags().BoolVar(&ops.sum, "sum", true, "Hide sum rows")
 
 	return getCmd
 }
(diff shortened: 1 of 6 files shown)


cmd/cost/list.go
```diff
@@ -142,6 +182,87 @@ func listCostsUnderOU(OU *organizations.OrganizationalUnit, awsClient awsprovide
...
+	Unit      string
+}
+
+type OUCost struct {
+	Costs   []AccountCost
+	OU      *organizations.OrganizationalUnit
+	options *listOptions
+}
+
+func (o *OUCost) getCost(awsClient awsprovider.Client) error {
```
georgettica, line 197: Why not return the costs here are pass the array to the print function?

cmd/cost/list.go
```diff
@@ -97,13 +94,56 @@ func (f listCostResponse) String() string {
...
+	} else {
+		ouCost.getCost(awsClient) // Get cost per account, print per account
+		if ops.level == "account" {
+			ouCost.printCostPerAccount()
+		}
+		cost, unit, err = ouCost.getSum() // Sum up account costs
+		if err != nil {
+			log.Fatalln("Error summing up cost of OU:", err)
+		}
+		printCostList(cost, unit, OU, ops, true) // TODO: Update bool here
```
georgettica, line 141: Why does this not have a return value?

cmd/cost/list.go
```diff
@@ -142,6 +182,87 @@ func listCostsUnderOU(OU *organizations.OrganizationalUnit, awsClient awsprovide
...
+
+	return nil
+}
+
+func (o OUCost) getSum() (sum float64, unit string, err error) {
+	sum = 0
+	for _, cost := range o.Costs {
+		sum += cost.Cost
+		if unit == "" {
+			unit = cost.Unit
```
georgettica, line 236: I guess this is a phase 1 thing, and if we get more it'll be implemented better

cmd/cost/list.go
```diff
@@ -142,6 +182,87 @@ func listCostsUnderOU(OU *organizations.OrganizationalUnit, awsClient awsprovide
...
+		fmt.Println("AccountID,Cost,Unit")
+	}
+	for _, accountCost := range o.Costs {
+		resp := listAccountCostResponse{
+			AccountId: accountCost.AccountID,
+			Cost:      accountCost.Cost,
+			Unit:      accountCost.Unit,
+		}
+		if o.options.csv {
+			fmt.Printf("%s,%.2f,(%s)\n", accountCost.AccountID, accountCost.Cost, accountCost.Unit)
```
georgettica, line 259: Shouldn't this be using the o.options value?

georgettica (commented):
Untested, but looks good overall

cmd/cost/list.go
NautiluX: more what? I think the current assumption is there is only USD. This is already an improvement since it will fail if not all numbers have the same currency.

cmd/cost/list.go
NautiluX: which value?

cmd/cost/list.go
NautiluX: We thought a type might be a better way than another function on the options struct and a function with 1 parameter better than a function with 3.

cmd/cost/list.go
NautiluX: you're right, printCostList doesn't return an error, so we should update the function signature.

NautiluX:
Jira: OSD-9701

cmd/cost/list.go
georgettica: And if in some wierd case it's not, a future implementation will sum by currency

But as I said it's phase 2 of this (if we get the error alot)

cmd/cost/list.go
georgettica: The output file
As printf prints to stdout and not to the provided output device

NautiluX: makes sense I think

cmd/cost/list.go
georgettica: The print function will only recieve The costs array
And the get needs to return an error maybe?

NautiluX:
fixes OSD-9687

cmd/cost/list.go
georgettica: Ok

cmd/cost/get.go
```diff
@@ -369,7 +370,7 @@ func (o *getOptions) printCostGet(cost float64, unit string, ops *getOptions, OU
 	}
 
 	if ops.csv { //If csv option specified, print result in csv
-		fmt.Printf("\n%s,%f (%s)\n\n", *OU.Name, cost, unit)
+		fmt.Printf("\n%s,%s (%s)\n\n", *OU.Name, cost.StringFixed(2), unit)
```
georgettica, line 373: This just shows with decimal points? 
Not around a goenv to check

georgettica:
/lgtm
/approve
/hold
Just in case my last comment was problematic

cmd/cost/get.go
NautiluX: it rounds to 2 decimal places, not sure if that answers your question?

NautiluX (commented):
/lgtm

PR #187, opened 2022-01-20
Add support for internal service logs

dustman9000:
This adds support to post and list internal service logs for use with cluster troubleshooting across on-call shifts.

To post an internal message:
```
osdctl servicelog post -i -p MESSAGE='super secret message' -y <cluster id>
```

To list internal messages:
```
osdctl servicelog list -i <cluster id>
```

API reference:
https://api.openshift.com/?urls.primaryName=Service%20logs#/default/post_api_service_logs_v1_cluster_logs

https://issues.redhat.com/browse/OSD-9692

### DIFF
--- a/cmd/servicelog/post.go
+++ b/cmd/servicelog/post.go
@@ -33,6 +33,7 @@ var (
 	isDryRun        bool
 	skipPrompts     bool
 	clustersFile    string
+	internalOnly    bool
 
 	// Messaged clusters
 	successfulClusters = make(map[string]string)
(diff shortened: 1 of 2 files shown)


cmd/servicelog/post.go
```diff
@@ -31,14 +31,16 @@ var (
...
 
 	// Messaged clusters
 	successfulClusters = make(map[string]string)
 	failedClusters     = make(map[string]string)
 )
 
 const (
-	defaultTemplate = ""
+	defaultTemplate        = ""
+	internalMessageSummary = "SREP INTERNAL"
```
dustman9000, line 43: Should this summary be customizable?

dustman9000:
Waiting on resolution of https://issues.redhat.com/browse/SDB-2539 such that internal service logs won't be viewable by cluster owners.

iamkirkbater:
From a usability perspective, I'm wondering if adding the two new flags is necessary?  I'm wondering if `osdctl servicelog post -i -p MESSAGE="my internal message goes here" $cluster_id` makes a bit more sense, as it uses the existing parameters methodology that we're used to today.

dustman9000:
> From a usability perspective, I'm wondering if adding the two new flags is necessary? I'm wondering if `osdctl servicelog post -i -p MESSAGE="my internal message goes here" $cluster_id` makes a bit more sense, as it uses the existing parameters methodology that we're used to today. Maybe we could do something similar with the default internal summary as well, only overwrite it if that parameter is passed through the `-p` flag.

@iamkirkbater - Good suggestion. I've removed the extra parameters and defaulted to "INTERNAL" as the SL summary and now require the user to specify the MESSAGE template parameter.

dustman9000:
I've verified that both email and ocm cluster history events do not show up for non-RH employee cluster owners. This is now ready for review/merge.

iamkirkbater (commented):
/lgtm

PR #189, opened 2022-01-31
feat(owned): add new command

georgettica:
CURRENT STATE:
Works correctly

the command will use the ocm creds only and list clusters based on the
creator (if no args are provided, use the current user)


georgettica:
feel free to review and give me your worst.
currently previewing the Subscription, but can move to show the Cluster resource if it'll look better

cmd/cluster/owner.go
```diff
@@ -0,0 +1,187 @@
...
+	return subscriptionsWithExternalID, nil
+}
+
+func (o *ownerOptions) run() error {
+	//The ocm
+	token := os.Getenv("OCM_TOKEN")
+	if token == "" {
+		ocmToken, err := utils.GetOCMAccessToken()
+		if err != nil {
+			return fmt.Errorf("OCM token not set. Please configure it using the OCM_TOKEN evnironment variable or the ocm cli")
```
shibumi, line 148: Quick question: Do we follow a specific style guide? The Go styleguide suggests that all errors should start with lower case. If we want to follow this Line 148 and Line 156 need a minor change :)

fahlmant: "provided OCM token not set"

shibumi: sounds good!

cmd/cluster/owner.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+		subscriptionsWithExternalID = append(subscriptionsWithExternalID, sub)
+		return true
+	})
+
+	return subscriptionsWithExternalID, nil
+}
+
+func (o *ownerOptions) run() error {
+	//The ocm
```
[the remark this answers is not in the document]
georgettica: yeah, forgot it was there, will remove

cmd/cluster/owner.go
shibumi: We could think about using golangci on the long term: https://github.com/golangci/golangci-lint
It's basically the de facto replacement of golint.

shibumi: I just have no idea how our Jenkins pipelines do work exactly or if it would make sense if we have some pipelines in Github Actions and some in Jenkins... probably we should discuss this with more people.

georgettica: I can explain how our CI works if you wish, but offline (and I think I'll record as it's good stuff even if we only see it once)

georgettica:
fixed all of your comments, hope it's good enough now (IE ready for another round in the review ring)

cmd/cluster/cmd.go
```diff
@@ -17,9 +19,29 @@ func NewCmdCluster(streams genericclioptions.IOStreams, flags *genericclioptions
...
+	k8svalid, err := o.ValidateIdentifiers()
+	if !k8svalid {
+		if err != nil {
+			cmdutil.PrintErrorWithCauses(err, io.ErrOut)
+			return err
+		}
+
+	}
+
+	awsvalid, err := o.Awscloudfactory.ValidateIdentifiers()
```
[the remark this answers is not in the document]
fahlmant: Agree, still needs to be removed

cmd/cluster/cmd.go
```diff
@@ -17,9 +19,29 @@ func NewCmdCluster(streams genericclioptions.IOStreams, flags *genericclioptions
...
 	return clusterCmd
 }
 
 func help(cmd *cobra.Command, _ []string) {
 	cmd.Help()
 }
+
+func CompleteValidation(o *k8spkg.ClusterResourceFactoryOptions, io genericclioptions.IOStreams) error {
+	k8svalid, err := o.ValidateIdentifiers()
+	if !k8svalid {
```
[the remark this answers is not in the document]
fahlmant: If the command isn't calling the kubernetes api at all, this can be removed.

cmd/cluster/owner.go
```diff
@@ -0,0 +1,64 @@
...
+	genericclioptions.IOStreams
+	GlobalOptions *globalflags.GlobalOptions
+}
+
+// newCmdOwner implements the health command to describe number of running instances in cluster and the expected number of nodes
+func newCmdOwner(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags, globalOpts *globalflags.GlobalOptions) *cobra.Command {
+	ops := newOwnerOptions(streams, flags, globalOpts)
+	healthCmd := &cobra.Command{
+		Use:               "health",
+		Short:             "Describes health of cluster nodes and provides other cluster vitals.",
```
fahlmant, line 28: Please update these appropriately

cmd/cluster/owner.go
```diff
@@ -0,0 +1,64 @@
...
+		return err
+	}
+
+	o.output = o.GlobalOptions.Output
+
+	return nil
+}
+
+func (o *ownerOptions) run() error {
+
```
fahlmant, line 62: Currently this doesn't do anything...

fahlmant (changes requested):
Why is cmd/root.go being changed?

georgettica (commented):
forgot to add these comments. will address them soon

cmd/cluster/owner.go
georgettica: Fixed :)

cmd/cluster/cmd.go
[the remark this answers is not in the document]
georgettica: I'll remove that later

georgettica:
so this works for 2 days now. added a small fix today (based on local testing of the previous script)

cmd/cluster/owner.go
```diff
@@ -0,0 +1,154 @@
...
+		fmt.Println("using the current user")
+		response, err := connection.AccountsMgmt().V1().CurrentAccount().Get().
+			Send()
+		if err != nil {
+			return fmt.Errorf("Can't send request: %v", err)
+		}
+		accountName = response.Body().Username()
+		accountID = response.Body().ID()
+	} else {
+		tmpl, err := template.New("test").Parse("id = '{{.}}' or username like '%{{.}}%' or email like '%{{.}}%'")
```
fahlmant, line 96: What is "test" here used for?

cmd/cluster/owner.go
```diff
@@ -0,0 +1,154 @@
...
+		if err != nil {
+			return fmt.Errorf("could not parse template: %w", err)
+		}
+		var tpl bytes.Buffer
```
fahlmant, line 100: This var name isn't really clear to me

cmd/cluster/owner.go
```diff
@@ -0,0 +1,154 @@
...
+
+		err = tmpl.Execute(&tpl, accountName)
+		if err != nil {
+			return fmt.Errorf("could not execute the template with the input %v: %w", accountName, err)
+		}
+		searchString := tpl.String()
+		fmt.Printf("the search query is '%s'\n", searchString)
```
fahlmant, line 107: This seems like it would be better behind a verbose -v option

cmd/cluster/owner.go
```diff
@@ -0,0 +1,154 @@
...
+			Send()
+
+		if err != nil {
+			return fmt.Errorf("Can't send request: %v", err)
+		}
+
+		if response.Total() != 1 {
+			fmt.Println("the users that were found are:")
+			v1.MarshalAccountList(response.Items().Slice(), os.Stdout)
+			fmt.Println()
```
fahlmant, line 119: Please remove

cmd/cluster/owner.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+		accountID = response.Items().Get(0).ID()
+
+	}
+
+	if accountID == "null" || accountID == "" {
+		return fmt.Errorf("could not extract the accountID")
+	}
+
+	fmt.Printf("the user is '%s' and his ID is '%s'\n", accountName, accountID)
```
fahlmant, line 130:
```suggestion
	fmt.Printf("the user is '%s' with ID '%s'\n", accountName, accountID)
```

fahlmant: Also would be good for verbose mode

cmd/cluster/owner.go
```diff
@@ -0,0 +1,154 @@
...
+
+	searchString := fmt.Sprintf("creator.id = '%s' and status != 'Deprovisioned' and status != 'Archived'", accountID)
```
fahlmant, line 132: Should the static part of this be a string defined outside this function? or maybe a const?

cmd/cluster/owner.go
```diff
@@ -0,0 +1,154 @@
...
+	response, err := connection.AccountsMgmt().V1().Subscriptions().List().Parameter("search", searchString).
+		Send()
+
+	fmt.Printf("the search query is '%s'\n", searchString)
```
fahlmant, line 136: Verbose mode candidate

cmd/cluster/owner.go
```diff
@@ -0,0 +1,154 @@
...
+	fmt.Printf("the user '%s' has '%d' owned clusters\n\n", accountName, response.Total())
+	if response.Total() == 0 {
+		return nil
+	}
+
+	items := []string{}
+	for _, i := range response.Items().Slice() {
+		items = append(items, i.ExternalClusterID())
+	}
+	fmt.Printf("'%v' are the clusters externalid's for this user\n", items)
```
fahlmant, line 151:
```suggestion
	fmt.Printf("'User %s owns the following clusters\n", accountName)
	for _, i := range response.Items().Slice() {
		fmt.Println(i.ExternalClusterID())
	}
```

fahlmant (changes requested):
Left some feedback for the main change. For future reference, please create a separate PR for the minor changes you make in other files. It clutters up review when there are so many unrelated changes all in a single commit

cmd/cluster/owner.go
georgettica: Sure, will do!

cmd/cluster/owner.go
georgettica: Newline is required as the template print is not printing a newline at the end.
Adding this makes reading the object easier

cmd/cluster/owner.go
georgettica: It's not, basically also an empty string will work.
It's how go templates work (annoying but required in this case)

fahlmant: Then can you change it to a blank string?

cmd/cluster/owner.go
fahlmant: Can you add a comment explaining that?

cmd/cluster/owner.go
fahlmant: The code should still read well if the var name is descriptive.

cmd/cluster/owner.go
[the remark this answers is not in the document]
fahlmant: There's no reason to allocate an array to store something if we can just print them all out immediately. Also, if I want to copy and paste one of the cluster IDs from the output, the newline is going to help easily distinguish each cluster ID individually.

cmd/cluster/owner.go
georgettica: Of course, that is the plan

cmd/cluster/owner.go
```diff
@@ -0,0 +1,160 @@
...
+		fmt.Printf("the search query is '%s'\n", searchString)
+	}
+
+	if err != nil {
+		return fmt.Errorf("Can't send request: %v", err)
+	}
+
+	fmt.Printf("'User %s owns the following clusters (total %d):\n", accountName, response.Total())
+
+	if response.Total() == 0 {
```
fahlmant, line 151: Why not do this before 149?

cmd/cluster/owner.go
```diff
@@ -0,0 +1,160 @@
...
+	o.output = o.GlobalOptions.Output
+
+	return nil
+}
+
+func createConnection() (*sdk.Connection, error) {
+	connection, err := ocm.NewConnection().Build()
+	if err != nil {
+		// TODO: this should be tracked by bubbling an error up and, exposing so we can compare it here.
+		if strings.Contains(err.Error(), "Not logged in, run the") {
```
fahlmant, line 67: Is there an error code we can check here instead? Also we can combine 68 with 70, start with a base error string and append "please run ocm login" to the message if the error code is hit

cmd/cluster/owner.go
```diff
@@ -0,0 +1,160 @@
...
+
+		response, err := connection.AccountsMgmt().V1().Accounts().List().Parameter("search", searchString).
+			Send()
+
+		if err != nil {
+			return fmt.Errorf("Can't send request: %v", err)
+		}
+
+		if response.Total() != 1 {
+			fmt.Println("the users that were found are:")
```
fahlmant, line 120:
```suggestion
			fmt.Println("Found users:")
```

cmd/cluster/owner.go
georgettica: no, they just return a `fmt.Errorf` with the message.

georgettica: K, let's do that then :/

georgettica: This should be better.. TAL?

PR #190, opened 2022-02-03
[OSD-9246] Fix account reset with resetlegalentityid

dkeohane:
Previously if you attempted to reset an account and move it up an OU it was already part of, AWS would throw a DuplicateAccountException error. We were not handling that case. These changes ensure that we're not trying to move an account to an OU it is already part of. 

This issue was encountered as part of research into dangling account crs [[OSD-9246]](https://issues.redhat.com/browse/OSD-9246)


georgettica:
I guess this also needs 
/approve
so enjoy :)

PR #191, opened 2022-03-03
feat(goreleaser): add binary hardening + update

shibumi:
This commit adds binary hardening for the released osdctl binaries.
Furthermore, it upgrades goreleaser to the newest version
and uses the Github Action for provisioning the GoReleaser.


.goreleaser.yml
```diff
@@ -1,14 +1,11 @@
-# This is an example .goreleaser.yml file with some sane defaults.
```
shibumi, line 1: Little comment to this one:

I am confident that we can remove this line and file from the secrets, because Github is exposing the secret via `${{ secret.GITHUB_TOKEN }}` on default.

georgettica:
looks good in general. 
will test the build locally awswell :)

shibumi:
@georgettica  I am running a test build right now: https://github.com/shibumi/osdctl/actions/runs/1944625906

shibumi:
It builds locally now:

```
❯ goreleaser build --rm-dist --skip-validate
   • building...      
   • loading config file       file=.goreleaser.yml
   • loading environment variables
   • getting and validating git state
      • building...               commit=74450663c19bef7a0891ec019b23b7acfd88e38a latest tag=v0.9.2-test.4
      • pipe skipped              error=validation is disabled
   • parsing tag      
   • running before hooks
      • running                   hook=go mod tidy
   • setting defaults 
   • checking distribution directory
      • --rm-dist is set, cleaning it up
   • loading go mod information
   • build prerequisites
   • writing effective config file
      • writing                   config=dist/config.yaml
   • generating changelog
      • writing                   changelog=dist/CHANGELOG.md
   • building binaries
      • building                  binary=dist/osdctl_linux_amd64/osdctl
      • building                  binary=dist/osdctl_darwin_amd64/osdctl
   • storing release metadata
      • writing                   file=dist/artifacts.json
      • writing                   file=dist/metadata.json
   • build succeeded after 3.03s
'goreleaser build --rm-dist --skip-validate' time: 3.095s, cpu: 340%
```

I removed the `gomod` parameter in the .goreleaser file again, because it created some trouble with a few of our dependencies:

```
release failed after 27.71s error=failed to proxy module: exit status 1: go: github.com/shibumi/osdctl@v0.9.2-test.3 requires
	github.com/openshift/hive@v1.0.5 requires
	sigs.k8s.io/cluster-api-provider-aws@v0.0.0: reading sigs.k8s.io/cluster-api-provider-aws/go.mod at revision v0.0.0: unknown revision v0.0.0
```

@georgettica I think we are ready to go now. What I could not test is the release process itself.. my pipeline had some trouble releasing it: https://github.com/shibumi/osdctl/runs/5446607949?check_suite_focus=true

georgettica:
Nice!
We can at least speak to the `hive` team
But this can be the next step :)

fahlmant:
Why do we need to talk to the hive team?

fahlmant:
Why do we need to talk to the hive team?

fahlmant:
@shibumi Can you clarify a bit here? Why is the download step being removed? Why are there now two steps with the same name, `Run GoReleaser`, with slightly different configurations?

georgettica:
@fahlmant as this change with the gomod wasn't able to pull a version (v0.0.0)
If this can be versioned on their side it'll solve this dep issue I think

fahlmant:
The version we have pinned for hive is really old. I'd say lets update that instead.

shibumi:
> @shibumi Can you clarify a bit here? Why is the download step being removed? Why are there now two steps with the same name, Run GoReleaser, with slightly different configurations?

@fahlmant  The download step is not necessary, because we can directly use the GitHub Action provided by the GoReleaser team: https://github.com/goreleaser/goreleaser-action

The different step name is an error. Good catch! :) I should rename the first one to "check", but it might makes more sense to remove the check step in the releaser configuration, because `goreleaser release ...` invokes check, already.

@georgettica  Right Now, I have no hive issue, because I simply do not use the `go proxy` functionality from Goreleaser. See here: https://goreleaser.com/customization/gomod/

shibumi:
FYI: I have updated the goreleaser binary in the Action workflow to 1.6.1 and removed the unnecessary check. The check is included in `goreleaser release --rm-dist`, hence not necessary in our release pipeline, but we still need the check for validating the configuration in the validate_goreleaser pipeline (in case we want to enhance the configuration in the future).

PR #192, opened 2022-03-09
Create loggingCheck comamnd to show logging support for a target cluster

fahlmant:
This command takes a specified cluster ID and looks at all external labels to determine if the logging stack is SREP supported or not


sam-nguyen7 (approved):
/lgtm

PR #194, opened 2022-03-09
feat: implement latest version via Github API

shibumi:
This PR implements the latest tag via callling the Github API.

Unfortunately, I had to introduce a new `version` subcommand, because the cobra in-built version flag `--version` is not capable of running a function as a hook before rootCmd initialization.

Furthermore, I have modified the version output a little bit. We are now printing valid JSON (like kubectl does):

```
❯ ./osdctl version
{
  "commit": "",
  "version": "",
  "latest": "v0.9.3"
}
```

This PR closes: [OSD-10247](https://issues.redhat.com/browse/OSD-10247)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,8 +1,6 @@
 package cmd
 
 import (
-	"fmt"
-
 	routev1 "github.com/openshift/api/route/v1"
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	gcpv1alpha1 "github.com/openshift/gcp-project-operator/pkg/apis"
@@ -26,12 +24,6 @@ import (
 	"github.com/openshift/osdctl/pkg/k8s"
 )
 
-// GitCommit is the short git commit hash from the environment
-var GitCommit string
-
-// Version is the tag version from the environment
-var Version string
-
 func init() {
 	_ = awsv1alpha1.AddToScheme(scheme.Scheme)
 	_ = routev1.AddToScheme(scheme.Scheme)
@@ -44,7 +36,6 @@ func NewCmdRoot(streams genericclioptions.IOStreams) *cobra.Command {
 	globalOpts := &globalflags.GlobalOptions{}
 	rootCmd := &cobra.Command{
 		Use:               "osdctl",
-		Version:           fmt.Sprintf("%s, GitCommit: %s", Version, GitCommit),
 		Short:             "OSD CLI",
 		Long:              `CLI tool to provide OSD related utilities`,
 		DisableAutoGenTag: true,
@@ -77,8 +68,12 @@ func NewCmdRoot(streams genericclioptions.IOStreams) *cobra.Command {
 	templates.ActsAsRootCommand(rootCmd, []string{"options"})
 	rootCmd.AddCommand(newCmdOptions(streams))
 
-	//Add cost command to use AWS Cost Manager
+	// Add cost command to use AWS Cost Manager
 	rootCmd.AddCommand(cost.NewCmdCost(streams, globalOpts))
+
+	// Add version subcommand. Using the in-build --version flag does not work with cobra
+	// because there is no way to hook a function to the --version flag in cobra.
+	rootCmd.AddCommand(versionCmd)
 
 	return rootCmd
 }
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -1,0 +1,89 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"time"
+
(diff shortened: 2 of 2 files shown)


shibumi:
If you are interested in the exact JSON we get from GitHub. Do the following commands:

```
$ curl https://api.github.com/repos/openshift/osdctl/releases/latest | jq 
```

If find other fields that might be from interest feel free to ping me and I will add them. I thought about adding the latest git tag or making the version output prettier. In General.

cmd/version.go
```diff
@@ -0,0 +1,79 @@
...
+var versionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Display the version",
+	Long:  "Display version of osdctl",
+	Run:   version,
+}
+
+func version(cmd *cobra.Command, args []string) {
+	latest, _ := getLatestVersion() // let's ignore this error, just in case we have no internet access
+	ver, _ := json.MarshalIndent(&versionResponse{
```
shibumi, line 40: @georgettica  Do you think I can just ignore the possible json error here? Or should I panic or output the error instead?

cmd/root.go
```diff
@@ -41,10 +35,11 @@ func init() {
 
 // NewCmdRoot represents the base command when called without any subcommands
 func NewCmdRoot(streams genericclioptions.IOStreams) *cobra.Command {
+	latest, _ := getLatestVersion() // let us ignore errors here. The "general" functionality is more important than getting the latest version
 	globalOpts := &globalflags.GlobalOptions{}
 	rootCmd := &cobra.Command{
 		Use:               "osdctl",
-		Version:           fmt.Sprintf("%s, GitCommit: %s", Version, GitCommit),
+		Version:           fmt.Sprintf("%s, GitCommit: %s, Latest: %s", Version, GitCommit, latest),
```
shibumi, line 42: Oh I have to remove this line... will be fixed in the next push

cmd/version.go
```diff
@@ -0,0 +1,79 @@
...
+	Commit  string `json:"commit`
+	Version string `json:"version`
+	Latest  string `json:"latest`
+}
+
+var versionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Display the version",
+	Long:  "Display version of osdctl",
+	Run:   version,
```
georgettica, line 35: this can be changed to RunE which does allow to return an error
so maybe that solves your problem

cmd/version.go
georgettica: added a review note on the `versionCmd` that might solve your issue

cmd/version.go
shibumi: Yes it does. Thank you!

cmd/version.go
```diff
@@ -0,0 +1,83 @@
...
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
+// GitCommit is the short git commit hash from the environment
+var GitCommit string
+
+// Version is the tag version from the environment
+var Version string
```
georgettica, line 17: create a var block for these variables?

georgettica:
not much is problematic for this PR, do this tiny nit and I will approve :)

cmd/version.go
shibumi: done

shibumi:
@georgettica  done

georgettica (commented):
/approve
/lgtm
I think we have done it multiple times, and maybe this should be written down for the next CLI wanting to check the latest release

Also found a bug in the GH releases that seems to not be addressed
/hold
If you think what I said is ok to proceed feel free to unhold

shibumi:
@georgettica So, what you are talking about is making a Go package out of the functionality that we could re-use in other clients? Is this correct?

georgettica:
just an SOP / mention of it somewhere.

the hold was more on the bug there is in github release that a latest release is decided by the last release that has a push in it) so in most cases it's ok but it can cause problems, see https://github.com/vmware-tanzu/velero/issues/4733

shibumi:
@georgettica I don't know if the GitHub API works the same way. It's also mentionable that the GitHub API only seems to return stable releases:

❯ curl https://api.github.com/repos/openshift/osdctl/releases/latest | jq
```json
{
  "url": "https://api.github.com/repos/openshift/osdctl/releases/61377308",
  "assets_url": "https://api.github.com/repos/openshift/osdctl/releases/61377308/assets",
  "upload_url": "https://uploads.github.com/repos/openshift/osdctl/releases/61377308/assets{?name,label}",
  "html_url": "https://github.com/openshift/osdctl/releases/tag/v0.9.3",
  "id": 61377308,
.....
}
```

shibumi:
@georgettica  for your information I am working on this SOP here: https://github.com/openshift/ops-sop/pull/1804

PR #195, opened 2022-03-09
chore(gomod): bump deps

georgettica:
related to
- CVE-2021-42576
- CVE-2015-3627


fahlmant (commented):
Could you clarify the changes a bit here? As far as I can tell this isn't changing the version of any current dependencies; this is adding completely new dependencies. Looking at the CVEs I don't see specific resolution laid out. Could you clarify what adding these packages does?

georgettica:
these packages were inner deps of some libraries we use. github suggested updating them.

this will resolve CVE's we don't currently have but it's better than waiting on a PR for a third party IMO (it can happen in parralel but still)

georgettica:
ran `go mod tidy` now so it would be clearer what happened

georgettica:
so this doesn't solve it like I hoped (now that I am looking into the go.sum)

so I will close and file an issue to the packages that introduced it

PR #196, opened 2022-03-10
implement in-place upgrades for osdctl

shibumi:
This PR implements in-place upgrades for osdctl. For doing this there are a few requirements:

1. We are getting the latest release URL via the GitHub API
2. Our releases should not change and **must** follow the same patterns: osdctl_0.9.3_Linux_x86_64.tar.gz, osdctl_0.9.3_Darwin_x86_64.tar.gz

This is an early POC (it works) and is based on one of my other PRs that implements the `version` subcommand.

To test this PR:

1.  goreleaser build --skip-validate --rm-dist
2. cd dist/osdctl_linux_amd64 
3. ./osdctl version
4. ./osdctl upgrade
5. ./osdctl version

Expected Output:

```
❯ ./osdctl version
{
  "commit": "ccce3bc",
  "version": "0.9.2",
  "latest": "v0.9.3"
}

❯ ./osdctl upgrade

❯ ./osdctl --version
osdctl version 0.9.3, GitCommit: aaa1cd7
```

Note the old `--version` flag for 0.9.3, that's because my new version subcommand is not yet on the main branch.

This PR closes: [OSD-10248](https://issues.redhat.com/browse/OSD-10248)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -15,7 +15,14 @@ builds:
     goarch:
       - amd64
     ldflags:
-      - -s -w -X github.com/openshift/osdctl/cmd.GitCommit={{.ShortCommit}} -X github.com/openshift/osdctl/cmd.Version={{.Version}}
+      # the following line will inject the current git commit and git tag in the binary during the build process.
+      # .ShortCommit and .Version are template variables that will be set during the GoReleaser run
+      # The "-X" go flag injects the strings into the two global variables GitCommit and Version
+      # See also: https://pkg.go.dev/cmd/link
+      - -s
+      - -w 
+      - -X github.com/openshift/osdctl/cmd.GitCommit={{.ShortCommit}}
+      - -X github.com/openshift/osdctl/cmd.Version={{.Version}}
       - "-extldflags=-zrelro" # binary hardening: For further explanation look here: https://www.redhat.com/en/blog/hardening-elf-binaries-using-relocation-read-only-relro
       - "-extldflags=-znow"
 
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -1,0 +1,145 @@
+package cmd
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"time"
+
+	"github.com/coreos/go-semver/semver"
+	"github.com/spf13/cobra"
+)
+
+var upgradeCmd = &cobra.Command{
+	Use:           "upgrade",
+	Short:         "Upgrade osdctl",
+	Long:          "Fetch latest osdctl from GitHub and replace the running binary",
+	RunE:          upgrade,
+	SilenceErrors: true,
+}
+
+func upgrade(cmd *cobra.Command, args []string) error {
+	// rootName ensures that the upgrade will fail if we ever decide to rename osdctl
+	// between releases :-)
+	rootName := cmd.Root().Name()
+
+	latest, err := getLatestVersion()
+	if err != nil {
+		return err
+	}
+	latestWithoutPrefix := strings.TrimPrefix(latest, "v")
+	// check if an upgrade is necessary
+	currentSemVer := semver.New(Version)
+	latestSemVer := semver.New(latestWithoutPrefix)
+	if !currentSemVer.LessThan(*latestSemVer) {
+		return fmt.Errorf("we are already up to date")
+	}
+	// upgrade necessary
+	client := http.Client{
+		Timeout: time.Second * 60,
+	}
+
+	addr := fmt.Sprintf(versionAddressTemplate,
+		latestWithoutPrefix,
+		latestWithoutPrefix,
+		parseGOOS(runtime.GOOS),
+		parseGOARCH(runtime.GOARCH))
+
+	req, err := http.NewRequest(http.MethodGet, addr, nil)
+	if err != nil {
+		return err
+	}
+
+	res, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+
+	gzf, err := gzip.NewReader(res.Body)
+	if err != nil {
+		return err
+	}
+
+	tr := tar.NewReader(gzf)
+	for {
+		f, err := tr.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return err
+		}
+		if f.Name != rootName {
+			continue
+		}
+		// For replacing a running executable we have to use the syscall "rename".
+		// "rename" can only be called on executables (old/new destination/name)
+		// that are stored on the same filesystem. This is the reason, why we cannot
+		// use a directory on ramfs here (f.e. /tmp/). Instead, we are creating a
+		// temp dir in the user's $HOME.
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			return err
+		}
+
+		dir, err := ioutil.TempDir(homeDir, ".*")
+		if err != nil {
+			return err
+		}
+
+		defer os.RemoveAll(dir)
+
+		tmpFilePath := filepath.Join(dir, rootName)
+
+		tmpFile, err := os.OpenFile(tmpFilePath, os.O_CREATE|os.O_RDWR, 0755)
+		if err != nil {
+			return err
+		}
+
+		_, err = io.Copy(tmpFile, tr)
+		if err != nil {
+			return err
+		}
+
+		// get path of current executable
+		exe, err := os.Executable()
+		if err != nil {
+			return err
+		}
+
+		os.Rename(tmpFilePath, filepath.Join(filepath.Dir(exe), rootName))
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func parseGOOS(goos string) string {
+	switch goos {
+	case "linux":
+		return "Linux"
+	case "darwin":
+		return "Darwin"
+	case "windows":
+		return "Windows"
+	default:
+		return ""
+	}
+}
+
+func parseGOARCH(goarch string) string {
+	switch goarch {
+	case "amd64":
+		return "x86_64"
+	default:
+		return ""
+	}
+}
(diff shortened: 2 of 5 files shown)


shibumi:
This PR is based on: https://github.com/openshift/osdctl/pull/194. We could either merge this one directly or merge the latest version first and I rebase.

cmd/upgrade.go
```diff
@@ -0,0 +1,127 @@
...
+	// check if an upgrade is necessary
+	currentSemVer := semver.New(Version)
+	latestSemVer := semver.New(latestWithoutPrefix)
+	if currentSemVer.LessThan(*latestSemVer) {
+		// upgrade necessary
+		client := http.Client{
+			Timeout: time.Second * 60,
+		}
+
+		addr := fmt.Sprintf(versionAddressTemplate, latestWithoutPrefix, latestWithoutPrefix, parseGOOS(runtime.GOOS), parseGOARCH(runtime.GOARCH))
```
georgettica, line 42: Line too long, make this a multiline statement

cmd/upgrade.go
```diff
@@ -0,0 +1,127 @@
...
+			if err != nil {
+				return err
+			}
+			if f.Name == "osdctl" {
+				homeDir, err := os.UserHomeDir()
+				if err != nil {
+					return err
+				}
+
+				dir, err := ioutil.TempDir(homeDir, ".*")
```
georgettica, line 74: Could you give an example to where this is created?

cmd/upgrade.go
```diff
@@ -0,0 +1,127 @@
...
+					return err
+				}
+
+				_, err = io.Copy(tmpFile, tr)
+				if err != nil {
+					return err
+				}
+
+				// get path of current executable
+				exe, err := os.Executable()
```
georgettica, line 92: That's cool!
Good to know :)

cmd/upgrade.go
```diff
@@ -0,0 +1,127 @@
...
+	case "darwin":
+		return "Darwin"
+	case "windows":
+		return "Windows"
+	default:
+		return ""
+	}
+}
+
+func parseGOARCH(goarch string) string {
```
georgettica, line 120: Not sure all the users have a GOARCH of amd64

cmd/version.go
```diff
@@ -10,6 +10,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	versionAPIEndpoint     = "https://api.github.com/repos/openshift/osdctl/releases/latest"
+	versionAddressTemplate = "https://github.com/openshift/osdctl/releases/download/v%s/osdctl_%s_%s_%s.tar.gz" // version, version, GOOS, GOARCH
+
```
georgettica, line 16: Trailing newline

georgettica (commented):
Overall looks good and cool, this might fail if you don't have admin perms, right?

georgettica:
@shibumi , you should have in this commit only the changes for this PR.
It's annoying you need to have many branches but that is what we need in case we want to merge this first

cmd/upgrade.go
shibumi: I only added amd64, because we have only amd64 releases right now

cmd/upgrade.go
shibumi: It creates a temporary directory in the user's HOME directory. First, I went for a temporary directory in the ramfs (f.e. `/tmp`), but the problem was that it is impossible to replace the binary this way. 

For replacing a binary/executable on POSIX systems (MacOS/Linux/BSD) we have to do the move syscall.
The move syscall has one strict requirement: The old file location and the new file location must be on the same filesystem... that's why I went for the HOME directory instead.

But this also means: If you are running osdctl from a different filesystem, the whole operation may fail.

Long story, short: it's complicated :/

shibumi: One addition: Using a hard copy operation does not work. We have to copy the binary first to the target filesystem and then do a "move" or "rename" operation.

shibumi:
@georgettica oh, this PR only consists of the necessary version.. it's just based on my other version PR, because I need that PR as dependency... As soon as we merge the other PR we master of osdctl should be updated and this PR should only show the necessary changes for this feature in particular.

cmd/upgrade.go
shibumi: @georgettica  I have added an explanation as comment in the code. Good catch! This definitely needed more explanation on why we are doing this, this way! :tada:

shibumi: @georgettica done

cmd/upgrade.go
```diff
@@ -0,0 +1,133 @@
...
+func upgrade(cmd *cobra.Command, args []string) error {
+	latest, err := getLatestVersion()
+	if err != nil {
+		return err
+	}
+	latestWithoutPrefix := strings.TrimPrefix(latest, "v")
+	// check if an upgrade is necessary
+	currentSemVer := semver.New(Version)
+	latestSemVer := semver.New(latestWithoutPrefix)
+	if currentSemVer.LessThan(*latestSemVer) {
```
ramonbutter, line 36:
```suggestion
	if !currentSemVer.LessThan(*latestSemVer) {
	   fmt.Println("Osdctl is already in newest version")
	   return nil
	}
	...
```
removes one level of nesting

cmd/upgrade.go
```diff
@@ -0,0 +1,133 @@
...
+		tr := tar.NewReader(gzf)
+		for {
+			f, err := tr.Next()
+			if err == io.EOF {
+				break
+			}
+			if err != nil {
+				return err
+			}
+			if f.Name == "osdctl" {
```
ramonbutter, line 69:
```suggestion
			if f.Name != "osdctl" {
                               continue
                        }
                        ....
```
removes one level of nesting

cmd/upgrade.go
```diff
@@ -0,0 +1,133 @@
...
+				// get path of current executable
+				exe, err := os.Executable()
+				if err != nil {
+					return err
+				}
+
+				os.Rename(tmpFilePath, filepath.Join(filepath.Dir(exe), "osdctl"))
+				if err != nil {
+					return err
+				}
```
ramonbutter, line 106: remove the temp Dir is not required?

cmd/upgrade.go
shibumi: Good catch!

cmd/upgrade.go
```diff
@@ -0,0 +1,145 @@
...
+	// between releases :-)
+	rootName := cmd.Root().Name()
+
+	latest, err := getLatestVersion()
+	if err != nil {
+		return err
+	}
+	latestWithoutPrefix := strings.TrimPrefix(latest, "v")
+	// check if an upgrade is necessary
+	currentSemVer := semver.New(Version)
```
fahlmant, line 39: Just curious. where does Version come from?

cmd/upgrade.go
```diff
@@ -0,0 +1,145 @@
...
+	latestSemVer := semver.New(latestWithoutPrefix)
+	if !currentSemVer.LessThan(*latestSemVer) {
+		return fmt.Errorf("we are already up to date")
```
fahlmant, line 42:
```suggestion
                fmt.Println("Already up to date, nothing to do!")
		return nil
```

cmd/upgrade.go
shibumi: Version is one of the variables we inject during the build process via Go build flags: https://github.com/openshift/osdctl/pull/196/files/ea42f9e990dbfa2857709584806749f22758536c#diff-7cb26babce207ed9c17e9dd3ead0b66d204da3cf4d11649412c197defbb3029cL18

shibumi:
@fahlmant done.

cmd/upgrade.go
shibumi: With Go 18 this kind of information is automatically embedded in `debug.GitVersion`, IIRC. This might be a nice task for the future :) Would make our build pipeline a little bit simpler

shibumi:
The build pipeline fails because it needs Go 1.7.0 (that needs Go 1.18). Looks like we are setting latest somewhere... I will fix thix.

cmd/upgrade.go
georgettica: @fahlmant this line
https://github.com/openshift/osdctl/blob/master/.goreleaser.yml#L18
Also @shibumi , can you add a comment here and there regarding this voodoo?

fahlmant: Yes, a comment would be much appreciated :)

shibumi:
@georgettica  I can't comment on your last comment.. but I fixed it. Have another look :)

cmd/upgrade.go
shibumi: @fahlmant  done

.goreleaser.yml
```diff
@@ -15,6 +15,9 @@ builds:
     goarch:
       - amd64
     ldflags:
+      # the following line will inject the current git commit and git tag in the binary during the build process.
+      # .ShortCommit and .Version are template variables that will be set during the GoReleaser run
+      # The "-X" go flag injects the strings into the two global variables GitCommit and Version
       - -s -w -X github.com/openshift/osdctl/cmd.GitCommit={{.ShortCommit}} -X github.com/openshift/osdctl/cmd.Version={{.Version}}
```
georgettica, line 21: this is great, but a crosslink  to this would make it better: the lines I want to have the crosslinks in
https://github.com/openshift/osdctl/blob/1447240f5a5f8b38288ee6b41a971d34e762d34d/cmd/version.go#L20-L24

if we are already here, maybe split the ldflags to
```
- -s
- -w
- -X ...
- -X ...
```
this way you can comment on each of them for the change

georgettica: the link to the -X is https://pkg.go.dev/cmd/link

shibumi: done

georgettica:
/approve
/lgtm
:)

georgettica:
I'll live with the consequences of this :)
We should cut a release for this too

shibumi:
Little not to this PR:

1. The version template must be stable (this is ensured by GoReleaser)
2. We don't have automated builds/releases right now (correct me if wrong @fahlmant )
3. We don't have another layer of security (for instance, signed releases... I am working on an ADR for that. More people looking into that would be awesome!)

fahlmant:
@georgettica feel free to cut a release

PR #197, opened 2022-03-13
chore(post): show args passed

georgettica:
this is good in case you accidentally provide too many args by mistake

### DIFF
--- a/cmd/servicelog/list.go
+++ b/cmd/servicelog/list.go
@@ -35,7 +35,7 @@ var listCmd = &cobra.Command{
 		}()
 
 		if len(args) != 1 {
-			log.Infof("The amount of arguments passed are: %d and they are '%v'", len(args), args)
+			log.Infof("Too many arguments. Expected 1 got %d", len(args))
 		}
 
 		// Use the OCM client to retrieve clusters
--- a/cmd/servicelog/list.go
+++ b/cmd/servicelog/list.go
@@ -34,6 +34,10 @@ var listCmd = &cobra.Command{
 			}
 		}()
 
+		if len(args) != 1 {
+			log.Infof("The amount of arguments passed are: %d and they are '%v'", len(args), args)
+		}
+
 		// Use the OCM client to retrieve clusters
 		clusters := getClusters(ocmClient, args)
 
--- a/cmd/servicelog/post.go
+++ b/cmd/servicelog/post.go
@@ -57,7 +57,7 @@ var postCmd = &cobra.Command{
 
 		var queries []string
 		if len(args) != 1 {
-			log.Infof("The amount of arguments passed are: %d and they are '%v'", len(args), args)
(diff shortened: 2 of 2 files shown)


georgettica:
related to [OSD-9903](https://issues.redhat.com/browse/OSD-9903)

georgettica:
related to OSD-9903

shibumi:
@georgettica  would it make sense to only enable this when a `--verbose` flag is being set?

shibumi:
@georgettica  Also: I don't see where your code is related to the Issue you are working on. Can you add a comment or an explanation how your code solves the issue described in: https://issues.redhat.com/browse/OSD-9903 ?

georgettica:
Seems ok to me, but I am already printing the command sent.
Once we move it to verbose, no on will use them

cmd/servicelog/list.go
```diff
@@ -34,6 +34,10 @@ var listCmd = &cobra.Command{
 			}
 		}()
 
+		if len(args) != 1 {
+			log.Infof("The amount of arguments passed are: %d and they are '%v'", len(args), args)
```
fahlmant, line 38:
```suggestion
			log.Infof("Too many arguments. Expected 1m got %d", len(args))
```

cmd/servicelog/post.go
```diff
@@ -56,6 +56,9 @@ var postCmd = &cobra.Command{
 		}
 
 		var queries []string
+		if len(args) != 1 {
+			log.Infof("The amount of arguments passed are: %d and they are '%v'", len(args), args)
```
fahlmant, line 60:
```suggestion
			log.Infof("Too many arguments. Expected 1m got %d", len(args))
```

cmd/servicelog/list.go
shibumi: Ah this clarifies my question I had!

shibumi: @fahlmant  shouldn't it be:"Too many arguments. Expected 1, but got %d"? Is the `1m` a typo?

fahlmant: yep, 1m is a typo. Fixed. And imo we should exit

cmd/servicelog/list.go
```diff
@@ -34,6 +34,10 @@ var listCmd = &cobra.Command{
 			}
 		}()
 
+		if len(args) != 1 {
+			log.Infof("Too many arguments. Expected 1 got %d", len(args))
```
shibumi, line 38: @georgettica  do we want to exit when the args don't match or is this not necessary?

georgettica: No, I accepted the comment but meant for this to be informational, in case we hit this later on

shibumi:
Oh! I didn't know that approve would "automerge" this. Sorry @fahlmant

georgettica:
Yeah, this is the automation we should have had in CAD, not there yet

PR #199, opened 2022-03-25
[OSD-10363] Add support for the AWS GovCloud partition

mjlshen:
- In the "front-end" CLI, I've abstracted out the partition such that it should work for any partition
- In the "back-end", I've also abstracted out the partition, but added support specifically for the GovCloud partition. As more partitions are added, `GetFederationEndpointUrl` and `GetConsoleUrl` can grow.
  - Previously the backend assumed us-east-1 as a safe default; I [added a retry](https://github.com/openshift/osdctl/compare/master...mjlshen:OSD-10363?expand=1#diff-a6eb9c64557660244a5986e0e7a0bfc80590bab12c6ad1393bc25bdfc0f473bdR185) for us-gov-west-1 (the default GovCloud region) if it fails.

- klog --> klog/v2
- Seems like prow was failing due to GoReleaser's latest version using `strings.Cut` that is part of Go 1.18. Unsure if we want to bump this project up to Go 1.18, so pinning GoReleaser for now

### DIFF
--- a/pkg/k8s/clusterresourcefactory.go
+++ b/pkg/k8s/clusterresourcefactory.go
@@ -4,22 +4,21 @@ import (
 	"context"
 	"errors"
 	"fmt"
-
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/service/sts"
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
+	"github.com/openshift/osdctl/cmd/common"
+	awsprovider "github.com/openshift/osdctl/pkg/provider/aws"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/klog/v2"
-
-	"github.com/openshift/osdctl/cmd/common"
-	awsprovider "github.com/openshift/osdctl/pkg/provider/aws"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // ClusterResourceFactoryOptions defines the struct for running list account command
@@ -86,7 +85,7 @@ func (factory *ClusterResourceFactoryOptions) GetCloudProvider(verbose bool) (aw
 	// only initialize kubernetes client when account name is set or cluster ID is set
 	if factory.AccountName != "" || factory.ClusterID != "" {
 		factory.KubeCli = NewClient(factory.Flags)
-		factory.Awscloudfactory.Region = "us-east-1"
+		factory.Awscloudfactory.Region = endpoints.UsEast1RegionID
 	}
 
 	supportRoleDefined := false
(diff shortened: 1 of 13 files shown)


pkg/provider/aws/client.go
```diff
@@ -122,37 +120,49 @@ type AwsClient struct {
...
-	_, err := sess.Config.Credentials.Get()
-
-	if aerr, ok := err.(awserr.Error); ok {
-		switch aerr.Code() {
-		case "NoCredentialProviders":
-			return nil, errors.Wrap(err, "Could not create AWS session")
-		default:
-			return nil, errors.Wrap(err, "Could not create AWS session")
+	if _, err := sess.Config.Credentials.Get(); err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
```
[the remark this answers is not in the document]
mjlshen: Ah, I tabbed it over so that `err` wouldn't get overwritten on previous line 167, similar thing to this going on in new line 177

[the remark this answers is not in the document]
dustman9000: Ah got it, the GH diff threw me off. It made it look like the previous whitespace was space, not tabs.

pkg/k8s/clusterresourcefactory.go
```diff
@@ -142,11 +141,21 @@ func (factory *ClusterResourceFactoryOptions) GetCloudProvider(verbose bool) (aw
 		klog.Error("Fail to get caller identity. Could you please validate the credentials?")
 		return nil, err
 	}
-	if verbose {
```
dustman9000, line 145: Was there a reason to remove this for verbose output?

pkg/provider/aws/client.go
```diff
@@ -164,20 +174,21 @@ func NewAwsClient(profile, region, configFile string) (Client, error) {
...
-				klog.Error("Credentials provided to osdctl are invalid. Please validate them and try again. Note this does not indicate a problem with the cluster's credentials")
+	if _, err := awsClient.GetCallerIdentity(nil); err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case "InvalidClientTokenId":
+				if region == endpoints.UsGovEast1RegionID || region == endpoints.UsGovWest1RegionID {
+					return nil, fmt.Errorf("failed `aws sts get-caller-identity` validation: %v", err)
+				}
+				klog.Infoln("credentials provided invalid, trying GovCloud by setting region to us-gov-west-1.")
+				return NewAwsClient(profile, endpoints.UsGovWest1RegionID, configFile)
```
dustman9000, line 185: This might have the potential for infinite recursion if the profile or configFile is invalid, or is a missing region the only reason for the InvalidClientTokenId error?

dustman9000:
Good stuff! I only had a couple questions, but nothing that would block this PR. Feel free to cancel the hold when ready.
/lgtm
/hold

pkg/provider/aws/client.go
mjlshen: An incorrect region (since it defaults to us-east-1) isn't the only reason; invalid values in the provided AWS credentials profile or configFile are other reasons as you mentioned.

However, if it runs into `InvalidClientTokenId` again after the recursive call, it would get caught by line 181 the next time it comes around and exit out since this recursive call explicitly sets it to us-gov-west-1. If not, it'll error out of the default case in line 187.

dustman9000: Ah gotcha, good point. Thanks!

pkg/k8s/clusterresourcefactory.go
mjlshen: It just wasn't helpful when debugging and felt like clutter that was leftover. That said there were effectively 0 additional logs emitted by verbose output on the entire `osdctl account console/cli` happy path, so I'm for adding more verbose logging to osdctl in general because it took quite a bit of digging for me to piece things together.

I think it should be more along the lines of "successful response from sts GetCallerIdentity" or "detected this is in GovCloud", etc.

PR #200, opened 2022-03-28
Enable darwin_arm64 builds

cblecker:
This enables builds for darwin_arm64 (M1-based macOS machines)

PR #201, opened 2022-03-29
fix: add missing arm64 GOOS for auto upgrades

shibumi:
This adds a missing GOOS case to the new upgrade command.

@georgettica  you might want to cut a new release.

TBH, we should remove the override in the goreleaser here:
```
archives:
  - replacements:
      darwin: Darwin
      linux: Linux
      amd64: x86_64
```

If we remove this section, we could get rid of the switch statements in the upgrade command.... because due to this replacements we have to maintain a mapping between the GOOS and GOARCH and our custom replacements. 
This leads to problems.

@cblecker's PR is related to this one: https://github.com/openshift/osdctl/pull/200 (I don't blame him.. he couldn't know that we have introduced the new upgrade command... we really should think about generating a changelog on the long term).

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -139,6 +139,8 @@ func parseGOARCH(goarch string) string {
 	switch goarch {
 	case "amd64":
 		return "x86_64"
+	case "arm64":
+		return "arm64"
 	default:
 		return ""
 	}


shibumi:
@georgettica  my suggestions:


1. Remove the archive renaming
2. Remove the mapping of GOOS/GOARCH in upgrade.go
3. Figure out a way how we can create changelogs. With GitHub actions we could easily generate one, but with prow I don't see an option atm. Do you have an idea? :(

georgettica:
I'll ask @cblecker who wanted the arm64 to TAL so it might have been solved already :)

georgettica:
I'm fine with removing the replacement, but making sure this change doesn't break current dev workflows

Also I am running goreleaser locally, so we can just add the sections and fill the changelog using the release command

shibumi:
If we decide to remove the replacement and mapping this will break the current upgrades from 0.9.5 to newer version, hence it might make sense to cut a major release 0.10.0 after doing it.

I am just not sure if changing the release tarball names breaks existing workflows for somebody.

cblecker:
I don't use (or really like) self-upgrading binaries. I maintain a homebrew tap for my own (and anyone that wants to use it) use: https://github.com/cblecker/homebrew-tap/blob/master/Formula/osdctl.rb

Because of the updated goreleaser config, I was able to modify my formula to include arm64 support: https://github.com/cblecker/homebrew-tap/commit/06e9b1fb16e24f0e20f827ecfc29215d5fd5ed86

georgettica:
Sry, this comment was required for the other PR #202

georgettica:
This i think can just be merged as is (if you want to verify @cblecker just for feature completeness)
But it looks ok by me
/lgtm
/approve
Also doing a
/hold
In case @fahlmant or @rafael-azevedo have objections

shibumi:
We have three options:

1. We can merge this PR here and ignore https://github.com/openshift/osdctl/pull/202
2. We can merge this PR and later merge https://github.com/openshift/osdctl/pull/202
3. We can not merge this PR and just merge https://github.com/openshift/osdctl/pull/202

Merging https://github.com/openshift/osdctl/pull/202 would "break" the `osdctl upgrade` command, because the URL to the release tarballs will change.


@cblecker small note: We can add brew support in GoReleaser.. then you wouldn't need to maintain a separate brew package and you could just directly get a brew package from GitHub. GoReleaser has full brew tap support: https://goreleaser.com/customization/homebrew/

cblecker:
I don't think this change will break anything on my end. No preferences on the other PR either.

@shibumi It does, but the way goreleaser does it, it takes the already pre-built binaries and just puts them in the right spot. I prefer to maintain my own formula and compile locally.

dastergon:
@shibumi any updates?

shibumi:
Sorry, I don't know why I missed this in my flood of GitHub notifications. I will have a look at this now.

shibumi:
@dastergon I will let this merge and close the other PR

PR #202, opened 2022-03-29
fix: remove custom replacements and mapping

shibumi:
This PR removes the custom archive replacements and custom GOOS/GOARCH mappings in our upgrade command.
I do suggest, that we cut a 0.10.0 release, because this feature is  a breaking change in the current upgrade behavior.


shibumi:
closed for now for the sake of compatibility

PR #203, opened 2022-04-06
Add pagination to account management commands

jharrington22:
This PR adds pagination to AWS API calls to ensure we get all results. 

Opened to fix the listing of accounts only returning a subset of accounts for a given organization

Before:

```
osdctl account mgmt list -p osd-staging-2  | grep Username | wc -l
16
```

After:

```
dist/osdctl_linux_amd64/osdctl account mgmt list -p osd-staging-2  | grep Usern
ame | wc -l                                                                                              
73 
```

I took the opportunity to fix up some other calls around listing users and policies where there could be 100's of results.

### DIFF
--- a/cmd/account/mgmt/account-assign.go
+++ b/cmd/account/mgmt/account-assign.go
@@ -159,29 +159,37 @@ func (o *accountAssignOptions) findUntaggedAccount(rootOu string) (string, error
 	input := &organizations.ListAccountsForParentInput{
 		ParentId: &rootOu,
 	}
-	accounts, err := o.awsClient.ListAccountsForParent(input)
+	var accounts []*organizations.ListAccountsForParentOutput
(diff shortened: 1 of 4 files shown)


rafael-azevedo:
@jharrington22 I believe the unit tests are failing.

jharrington22:
> @jharrington22 I believe the unit tests are failing.

Yeap I'll sort these out thanks.

georgettica:
saw a review of me was requested, but it's still not ready (tests are suggested to be added)
it seems it would be the correct thing to do is to
/hold
until it's ready

georgettica (changes requested):
as written before, add the corresponding tests and make checks pass

dastergon:
@jharrington22 do you think we should find a new owner for the PR?

PR #205, opened 2022-04-12
[OSD-10470] Now displaying flags and global flags in help

Nikokolas3270:
The help for `osdctl cost -h` after that:

```                           
The cost command allows for cost management on the AWS platform (other 
platforms may be added in the future)

Usage:
  osdctl cost [command]

Available Commands:
  create      Create a cost category for the given OU
  get         Get total cost of a given OU
  list        List the cost of each Account/OU under given OU
  reconcile   Checks if there's a cost category for every OU. If an OU is missing a cost category, creates the cost category

Flags:
  -a, --aws-access-key-id string       AWS Access Key ID
  -c, --aws-config string              specify AWS config file path
  -p, --aws-profile string             specify AWS profile
  -g, --aws-region string              specify AWS region (default "us-east-1")
  -x, --aws-secret-access-key string   AWS Secret Access Key
  -h, --help                           help for cost

Global Flags:
      --alsologtostderr                  log to standard error as well as files
      --as string                        Username to impersonate for the operation
      --cluster string                   The name of the kubeconfig cluster to use
      --context string                   The name of the kubeconfig context to use
      --insecure-skip-tls-verify         If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
      --kubeconfig string                Path to the kubeconfig file to use for CLI requests.
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory
      --logtostderr                      log to standard error instead of files
  -o, --output string                    Invalid output format: Valid formats are ['', 'json', 'yaml']
      --request-timeout string           The length of time to wait before giving up on a single server request. Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests. (default "0")
  -s, --server string                    The address and port of the Kubernetes API server
      --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
  -v, --v Level                          log level for V logs
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging

Use "osdctl cost [command] --help" for more information about a command.
```

Sub-command `options` at root level is still there but is not displayed anymore when shooting `osdctl -h`. Remark that as discussed in below issue on `minikube`, a solution could be to explicitly implement an `options` sub-command at root level:
https://github.com/kubernetes/minikube/issues/5036
Here the code for that `options` sub-command:
https://github.com/kubernetes/minikube/pull/6144/files
Not sure we really want to go in that way as `osdctl options` still displays the same before and after my change; also [Cobra](https://github.com/spf13/cobra) is perfectly able to display in a nicely manner all flags & global flags available for each command in the tree.
Hope I am not breaking anything by removing the call to `templates.ActsAsRootCommand`. Please tell me if that's the case.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -9,7 +9,6 @@ import (
 
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/kubectl/pkg/util/templates"
 
 	"github.com/openshift/osdctl/cmd/aao"
 	"github.com/openshift/osdctl/cmd/account"
@@ -67,7 +66,6 @@ func NewCmdRoot(streams genericclioptions.IOStreams) *cobra.Command {
 	rootCmd.AddCommand(newCmdCompletion(streams))
 
 	// add options command to list global flags
-	templates.ActsAsRootCommand(rootCmd, []string{"options"})
 	rootCmd.AddCommand(newCmdOptions(streams))
 
 	// Add cost command to use AWS Cost Manager


Nikokolas3270:
/assign @iamkirkbater
Hello Kirk, could you please review this change and share your feedback on it?
Many thanks! Nico.

shibumi:
@Nikokolas3270 can you provide some information about ActsAsRootCommand and what this is doing and why we are removing it?

Nikokolas3270:
@shibumi: here is the code for that function:
https://github.com/kubernetes/kubectl/blob/ef04426c7bdb5f533703dcadb28a61339a5484d1/pkg/util/templates/templater.go#L36
It is not clear what it is supposed to do. For sure googling `ActsAsRootCommand` does not yield many results.

shibumi:
> @shibumi: here is the code for that function: https://github.com/kubernetes/kubectl/blob/ef04426c7bdb5f533703dcadb28a61339a5484d1/pkg/util/templates/templater.go#L36 It is not clear what it is supposed to do. For sure googling `ActsAsRootCommand` does not yield many results.

Yes, it just looks like a template for help output etc...

The PR looks fine for me.

/lgtm

georgettica:
does what it is supposed to 
/approve
/lgtm
/hold in case the `ActsAsRootCommand` is required for someone
you have the power to unhold @Nikokolas3270

Nikokolas3270:
Hello,

Following above discussion, here is my reply to the following question: what does `ActsAsRootCommand` exactly do?

We already know that the only purpose of this function is to redefine the help & usage computed by Cobra.
This is done using [text/template](https://pkg.go.dev/text/template#Template.Funcs) package with the following format:
https://github.com/kubernetes/kubectl/blob/652881798563c00c1895ded6ced819030bfaa4d7/pkg/util/templates/templates.go#L24
As you can see line 51, options/flags are only displayed if they are visible or exposed:
- Visible options are the options which are local to the command node and which are not propagated to the descendants nodes.
- Exposed options must be explicitly defined as such with a proper call to the `ExposeFlags` method on the templater/`FlagExposer` returned by `ActsAsRootCommand`. We will see later how to that, but first let's understand in which case we may have visible options.

Visible options are the options which are local to the command, this means that `-p` option defined in `osdctl cost` command cannot be visible in `osdctl cost create` as it is not local in this sub-command (it is inherited from the parent command).
Also the `-p` option is not visible in `osdctl cost` command as it is defined as a `PersistentFlags()` and therefore propagates to the descendant commands.
**In other words, whatever the command node you consider: persistent flags cannot be visible options**.

I guess this choice was made because knowing if a persistent flag also applies to the command in which it is defined requires checking whether or not the is a target associated to that command (i.e. if the `Run` member is not `nil`) and this may be something not easy to do where the check is made (but for sure it is not impossible to do).
That said a persistent flag applies to all descendant commands for sure and I do not understand why they are ignored there. Because of that "bug", factorising flags definition in a parent command is no longer possible unless you expose the flag as follow:
```
templates.ActsAsRootCommand(costCmd, []string{"options"}).ExposeFlags(costCmd, "aws-profile")
```
I you add above instruction in `NewCmdCost` function of `cmd/cost/cost.go` file, `osdctl cost -h` will now print as follow:
```
The cost command allows for cost management on the AWS platform (other 
platforms may be added in the future)

Available Commands:
  create      Create a cost category for the given OU
  get         Get total cost of a given OU
  list        List the cost of each Account/OU under given OU
  reconcile   Checks if there's a cost category for every OU. If an OU is missing a cost category, creates the cost
category

Options:
  -p, --aws-profile='': specify AWS profile

Use "osdctl cost <command> --help" for more information about a given command.
```
Of course you will need to do that for every `PersistentFlags()` registered there to have all of them correctly printed in the help.
Note that it is also possible to do that in all sub-commands of `osdctl cost` (like `osdctl cost create`); that way:
- `osdctl cost` still does not list any option which make sense as the command has no target and is only there to group sub-commands
- And only the sub-commands are listing the options in their help. Of course this is not great from a factorisation perspective.

To sum up, I would say:
1. Yes, it is possible to keep using `ActsAsRootCommand` if persistent flags/options are all exposed.
2. But it would be better to fix `ActsAsRootCommand` so it always consider visible persistent flags as visible (unless there is no target for the action to consider; remark that defining a non-persistent flag for an action without a target does not make sense)
3. If we are not willing to make the effort of fixing `ActsAsRootCommand`, I propose to let the pull request as it it. I let you debate in the comments.

Finally:
- About the second parameter of the `ActsAsRootCommand` function; that is to say `[]string{"options"}`. This parameter is only there to tell that the `config` (implicit) subcommand should be categorised as belonging to no command group. This could be documented.
- I think that the `ActsAsRootCommand` function is not particularly well named, it could be renamed... but for sure it could have a better documentation.

shibumi:
/assign @fahlmant 

Taylor can you have a look and merge this if it's okay?

cmd/root.go
```diff
@@ -9,7 +9,6 @@ import (
 
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/kubectl/pkg/util/templates"
```
fahlmant, line 12: With this removed, can you run `go mod tidy` to see if anything gets removed?

Nikokolas3270: Good idea but sadly nothing is removed in those files as module `k8s.io/kubectl` is used in bunch of other files.

fahlmant (approved):
/approve

PR #206, opened 2022-04-12
[OSD-10928] Added CLUSTER_ID as param of 'osdctl servicelog post' in help

Nikokolas3270:
After that change `osdctl servicelog post -h` will output as follows:
```
Send a servicelog message to a given cluster

Options:
  -c, --clusters-file='': Read a list of clusters to post the servicelog to. the format of the file is:
{"clusters":["$CLUSTERID"]}
  -d, --dry-run=false: Dry-run - print the service log about to be sent but don't send it.
  -p, --param=[]: Specify a key-value pair (eg. -p FOO=BAR) to set/override a parameter value in the template.
  -q, --query=[]: Specify a search query (eg. -q "name like foo") for a bulk-post to matching clusters.
  -f, --query-file=[]: File containing search queries to apply. All lines in the file will be concatenated into a single
query. If this flag is called multiple times, every file's search query will be combined with logical AND.
  -t, --template='': Message template file or URL
  -y, --yes=false: Skips all prompts.

Usage:
  osdctl servicelog post CLUSTER_ID [flags] [options]

Use "osdctl options" for a list of global command-line options (applies to all commands).
```

Please tell me if this is okay for you.
Thanks!
   Nico

### DIFF
--- a/cmd/servicelog/post.go
+++ b/cmd/servicelog/post.go
@@ -43,7 +43,7 @@ const (
 
 // postCmd represents the post command
 var postCmd = &cobra.Command{
-	Use:   "post",
+	Use:   "post CLUSTER_ID",
 	Short: "Send a servicelog message to a given cluster",
 	Run: func(cmd *cobra.Command, args []string) {
 


cmd/servicelog/post.go
```diff
@@ -43,7 +43,7 @@ const (
 
 // postCmd represents the post command
 var postCmd = &cobra.Command{
-	Use:   "post",
+	Use:   "post CLUSTER_ID",
```
shibumi, line 46: I think C0bra has an automated way to generate this first argument. You might want to look for "Positional arguments" with c0bra :thinking:

Nikokolas3270: Indeed this would make sense and be better

Nikokolas3270: I based my work on this example found in Cobra code base:
https://github.com/spf13/cobra/blob/master/doc/cmd_test.go
In which they do not use positional args... but they exist, so lets use them. Thanks!

Nikokolas3270: Hum, actually even if it is possible to give constraints on the number of arguments, it is not possible to name them:
https://stackoverflow.com/questions/69167939/named-positional-arguments-in-cobra
Here the API to constraint the number of args (something we could use as at most one cluster id is allowed):
https://github.com/spf13/cobra/blob/45e521ea93e2270ef7333487f0a87e3887de2da9/args.go#L60

This is also discussed there:
https://github.com/spf13/cobra/issues/395#issuecomment-293634704

georgettica:
for now we can merge this
/approve
/lgm
/hold unless you want to wait for upstream

same here, @Nikokolas3270 you have the unhold power

Nikokolas3270:
Merging pull request as discussed with @georgettica et @shibumi 
Once below issue is addressed, we can always can back to modify the code to use the upcoming named arguments feature of Cobra.
https://github.com/spf13/cobra/issues/1666

georgettica:
PS, you were supposed to unhold and the automation would merge it
`/unhold`
We reserve manual merge for extreme cases from what I reckon

PR #207, opened 2022-04-14
OSD-10993 - Split user params only once to avoid truncation

tnierman:
Fixes https://issues.redhat.com/browse/OSD-10993.

---

User parameters for servicelog templates will be truncated if they contain more than on `=`. This was found to be because the `parseUserParameters` function only takes the second token after dividing along every `=` separator in the user parameter string.

By only dividing the string a single time, the full parameter is successfully added to the template.


fahlmant (approved):
/approve

sam-nguyen7 (approved):
/lgtm

PR #210, opened 2022-04-22
Do not allow passing -u *and* -i

bergmannf:
Right now when passing -u and -i the -u flag will take precedence.

This is very confusing, so instead of just unassigning everything, don't
allow passing both flags at once.


dkeohane (approved):

fahlmant (approved):
/approve

PR #211, opened 2022-04-27
OSD-11287 Don't auto-retry GovCloud and improve the error message

mjlshen:
Retrying GovCloud and printing out the info line is confusing for users not intending to access a GovCloud account and having invalid credentials in the local AWS profile is very common, so I've improved the error message with more actionable steps a user can take before reaching out.

It now looks something like this:

```
❯ ./osdctl account console -i 12345 -p myProfile -o env
error: `aws sts get-caller-identity --profile myProfile --region us-east-1` failed - ensure the credentials in the profile are valid and the region is in the AWS partition you expect
```


fahlmant (approved):
/approve
/lgtm

PR #213, opened 2022-04-29
OSD-11287 Fix bug where region is being overriden in clusterresourcefactory

### LINKED ISSUE #212: region is always defaulting to UsEast1RegionID
I'm using osdctl to access account console using -r option, however it's always defaulting to us-east-1 
This is getting set in https://github.com/openshift/osdctl/blob/master/pkg/k8s/clusterresourcefactory.go#L155 it should really be respecting the -r option

mjlshen:
Fixes #212 , after this fix osdctl account console/credentials respects the `-r` flag again when specifying an account id with the `-i` flag


fahlmant (approved):
/approve

PR #214, opened 2022-04-29
Not explicitly running the help target anymore for commands packing sub-commands

Nikokolas3270:
This does not change the way those commands behave when called without a sub-command.
This only changes the help displayed for those commands: the help no more encourage running those commands without a sub-command.
This generalizes hcnage done in [2124a930b143b27d5f92fbe14acc1f9c23fe871c](https://github.com/openshift/osdctl/commit/2124a930b143b27d5f92fbe14acc1f9c23fe871c) to all commands.


georgettica:
Seems quick and easy
/lgtm
/approve

PR #215, opened 2022-05-06
Adding ability to specify which AWS Account ID is to be assigned

dkeohane:
Ticket ref: https://issues.redhat.com/browse/OSD-11010

Previously we could only be assigned account ids found by the command , now we can specify an AWS Account ID we'd like to be assigned. 

Added a unit test to cover this which tests the command at the Cobra CLI level - so it feels more like an integration test to me, but allows the flexibility of mocking when desired, this is open for scrutiny :)   To do so, I had to remove the `CheckErr` function wrapping the `complete` and `run` functions (as this would exit out of the test if an error was encountered)  and the cobra.Command parameter was changed to `RunE` instead of `Run` for testability purposes (now actually returning an error).

### DIFF
--- a/cmd/account/mgmt/account-assign.go
+++ b/cmd/account/mgmt/account-assign.go
@@ -30,6 +30,7 @@ type accountAssignOptions struct {
 	awsClient    awsprovider.Client
 	username     string
 	payerAccount string
+	accountID    string
 	output       string
 
 	flags      *genericclioptions.ConfigFlags
@@ -66,13 +67,12 @@ func newCmdAccountAssign(streams genericclioptions.IOStreams, flags *genericclio
 		Run: func(cmd *cobra.Command, args []string) {
 			cmdutil.CheckErr(ops.complete(cmd, args))
 			cmdutil.CheckErr(ops.run())
-
 		},
 	}
 	ops.printFlags.AddFlags(accountAssignCmd)
 	accountAssignCmd.Flags().StringVarP(&ops.payerAccount, "payer-account", "p", "", "Payer account type")
 	accountAssignCmd.Flags().StringVarP(&ops.username, "username", "u", "", "LDAP username")
-
+	accountAssignCmd.Flags().StringVarP(&ops.accountID, "account-id", "i", "", "(optional) Specific AWS account ID to assign")
 	return accountAssignCmd
 }
 
@@ -106,6 +106,7 @@ func (o *accountAssignOptions) run() error {
 	} else {
 		return fmt.Errorf("invalid payer account provided")
 	}
+
 	//Instantiate aws client
 	awsClient, err := awsprovider.NewAwsClient(o.payerAccount, "us-east-1", "")
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/account/mgmt/account-assign.go
```diff
@@ -63,16 +65,18 @@ func newCmdAccountAssign(streams genericclioptions.IOStreams, flags *genericclio
 		Use:               "assign",
 		Short:             "Assign account to user",
 		DisableAutoGenTag: true,
-		Run: func(cmd *cobra.Command, args []string) {
-			cmdutil.CheckErr(ops.complete(cmd, args))
-			cmdutil.CheckErr(ops.run())
-
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := ops.complete(cmd, args)
```
bergmannf, line 69: Might be useful to at least mention in the commit message that `CheckErr` was removed and the parameter was changed to `RunE` for testability purposes.

cmd/account/mgmt/account-assign_test.go
```diff
@@ -1,18 +1,89 @@
...
+	awsprovider "github.com/openshift/osdctl/pkg/provider/aws"
 	"github.com/openshift/osdctl/pkg/provider/aws/mock"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
 )
 
+func TestAccountAassignCommand(t *testing.T) {
+
+	mocks := setupDefaultMocks(t, []runtime.Object{})
```
iamkirkbater, line 23: I see that you call out having to change up the command instantiation to test this, but I'm wondering if there was any reason why you couldn't just do something like:

```
o := accountAssignOptions{}
o.awsClient = mockAwsClient
o.username = "myuser"
o.accountid = "12345"

err := o.run()
```

instead of running this big integration-test like unit test?

dkeohane: Yup, I could have done exactly that! Honestly this mainly just stemmed from curiosity to see if I could find an interesting way to run integration tests, so I'm happy to create a `o.run()` unit test too if you'd like?

dkeohane: I've moved the existing test into an integration test file, and created a unit test for isOwned

> Maybe use that to test raw input handling but not specifically the functionality of the operations in the command?
If we're just testing the input and not the command operations then we're kinda just writing unit tests for the Cobra command library then?

iamkirkbater: Yes and no.  What I imagine is in conditions like where we need to specify one flag OR another, we would have a few test cases that are testing the input validation function we have:

1. Neither flag is passed - test that command errors
2 & 3. one and only one of each flag is passed - test that command succeeds
4. Both flags passed - test that command errors

Though I guess the way we structure this we could handle this with a function that can also get unit tested as well, instead of doing a full integration test.

So yeah, just unit testing here would work.  I'm not going to ask you to do _more_ work for little to no real gain :)

cmd/account/mgmt/account-assign.go
iamkirkbater: We can probably re-introduce the `CheckErr` wrapper here now, right?

iamkirkbater (changes requested):
Let's just reimplement the CheckErr to remain consistent and then I think this is ready to go.

iamkirkbater:
Thanks for your work on this Dan!

/lgtm

PR #216, opened 2022-05-09
[OSD-11388] Fix for osdctl can assign suspended accounts

dkeohane:
Ticket Ref: https://issues.redhat.com/browse/OSD-11388

Adding check to account mgmt assign to ensure we only assign active accounts. 
Added a unit test

This PR still remains untested with actual account assignment


iamkirkbater:
I think you should just need to run `make mockgen` and commit the updated mock files to get the lint job to pass.

cmd/account/mgmt/account-assign.go
```diff
@@ -192,8 +200,14 @@ func (o *accountAssignOptions) findUntaggedAccount(rootOu string) (string, error
 		}
 
 		if !isOwned {
-			accountAssignID = *a.Id
-			break
+			isSuspended, err := isSuspended(*a.Id, o.awsClient)
+			if err != nil {
+				return "", err
+			}
```
fahlmant, line 206: If the account is suspended, would it make sense here to `continue` to the next loop to find an account that isn't suspended?

fahlmant: Hmm.. sorry I'm not following. Won't this return at line 178, and stop this function from running?

fahlmant: Ah duh, this only returns on err, not on isSuspended

iamkirkbater (commented):
/lgtm

PR #223, opened 2022-06-28
fix: gosec warnings

shibumi:
This commit fixes all gosec warnings either via ignoring the error, where it makes sense
or via fixing it. I have fixed multiple implicit memory aliasing errors in loops.
These can be dangerous. For more information: https://stackoverflow.com/questions/62446118/implicit-memory-aliasing-in-for-loop

Fixes: OSD-11463


cmd/account/get/legal-entity.go
```diff
@@ -101,12 +101,18 @@ func (o *getLegalEntityOptions) run() error {
...
+			err := outputflag.PrintResponse(o.output, resp)
+			if err != nil {
+				fmt.Println("Error while printing response: ", err.Error())
+			}
 		}
 	}
 
 	// matched account not found
-	fmt.Fprintf(o.IOStreams.Out, "Account matched for AWS Account ID %s not found\n", o.accountID)
+	_, err := fmt.Fprintf(o.IOStreams.Out, "Account matched for AWS Account ID %s not found\n", o.accountID)
```
boranx, line 112: afaik, Fprintf() or Sprintf() does only formatting without spewing the outcome to the stdout
If we omit the first param, would there be a value for using it?

shibumi: There is no value of using the first parameter. It's just an integer stating how much bytes have been written

pkg/provider/aws/iam.go
```diff
@@ -179,6 +179,7 @@ func RefreshIAMPolicy(awsClient Client, federatedRole *awsv1alpha1.AWSFederatedR
 
 	// Detach all removed polices from role
 	for _, policyArn := range policyArnsToDetach {
+		policyArn := policyArn // fix for implicit memory aliasing
```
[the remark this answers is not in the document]
shibumi: Yes, I tried the fix beneath and this led to a failing unit test. The error was an out-of-bounds error, so we tried to access an element at an index, that did not exist. I did not investigate this further and just opted for this solution here, because I do not wanted to change the tests.

Changing the tests, might change the desired behavior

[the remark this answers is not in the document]
fahlmant: What about changing the later for loop to use this pattern?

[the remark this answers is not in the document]
shibumi: This should work

cmd/account/get/legal-entity.go
```diff
@@ -101,12 +101,18 @@ func (o *getLegalEntityOptions) run() error {
 				Id:   account.Spec.LegalEntity.ID,
 			}
 
-			outputflag.PrintResponse(o.output, resp)
+			err := outputflag.PrintResponse(o.output, resp)
+			if err != nil {
+				fmt.Println("Error while printing response: ", err.Error())
+			}
```
fahlmant, line 107:
```suggestion
			err := outputflag.PrintResponse(o.output, resp)
			if err != nil {
				fmt.Println("Error while printing response: ", err.Error())
				return err
			}
```

shibumi: Done

cmd/account/mgmt/account-list.go
```diff
@@ -136,7 +136,10 @@ func (o *accountListOptions) run() error {
 			Accounts: value,
 		}
 
-		outputflag.PrintResponse(o.output, resp)
+		err := outputflag.PrintResponse(o.output, resp)
+		if err != nil {
+			fmt.Println("Error while printing response: ", err.Error())
+		}
```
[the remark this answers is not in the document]
shibumi: Done

cmd/cost/get.go
```diff
@@ -125,7 +125,10 @@ func (o *getOptions) run() error {
 		}
 	}
 
-	o.printCostGet(cost, unit, o, OU)
+	err = o.printCostGet(cost, unit, o, OU)
+	if err != nil {
+		log.Println("Error calling printCostGet(): ", err.Error())
```
fahlmant, line 130: return err

cmd/cost/get.go
```diff
@@ -379,7 +382,10 @@ func (o *getOptions) printCostGet(cost decimal.Decimal, unit string, ops *getOpt
 		fmt.Println("Cost of all accounts under OU:")
 	}
 
-	outputflag.PrintResponse(o.output, resp)
+	err := outputflag.PrintResponse(o.output, resp)
+	if err != nil {
+		fmt.Println("Error calling PrintResponse(): ", err.Error())
```
fahlmant, line 387: return err

cmd/cost/list.go
```diff
@@ -260,7 +260,10 @@ func (o OUCost) getSum() (sum decimal.Decimal, unit string, err error) {
 }
 
 func (o OUCost) printCostPerAccount(awsClient awsprovider.Client) {
-	o.getCost(awsClient)
+	err := o.getCost(awsClient)
+	if err != nil {
+		fmt.Println("Error while calling getCost(): ", err.Error())
```
fahlmant, line 265: return

cmd/cost/list.go
```diff
@@ -273,7 +276,10 @@ func (o OUCost) printCostPerAccount(awsClient awsprovider.Client) {
 			fmt.Printf("%s,%s,%s,%s\n", *o.OU.Id, accountCost.AccountID, accountCost.Cost.StringFixed(2), accountCost.Unit)
 			continue
 		}
-		outputflag.PrintResponse(o.options.output, resp)
+		err := outputflag.PrintResponse(o.options.output, resp)
+		if err != nil {
+			fmt.Println("Error while printing response: ", err.Error())
```
fahlmant, line 281: return

cmd/cost/list.go
```diff
@@ -305,5 +311,8 @@ func printCostList(cost decimal.Decimal, unit string, OU *organizations.Organiza
 		return
 	}
 
-	outputflag.PrintResponse(ops.output, resp)
+	err := outputflag.PrintResponse(ops.output, resp)
+	if err != nil {
+		fmt.Println("Error while printing response: ", err.Error())
```
fahlmant, line 316: return

cmd/env/env.go
```diff
@@ -94,7 +95,10 @@ func (e *OcEnv) RunCommand(cmd *cobra.Command, args []string) {
 		e.Options.Alias = args[0]
 	}
 	if e.Options.ClusterId == "" && e.Options.Alias == "" {
-		cmd.Help()
+		err := cmd.Help()
+		if err != nil {
+			fmt.Println("could not print help")
```
fahlmant, line 100: return

cmd/env/env.go
```diff
@@ -141,13 +145,20 @@ func (e *OcEnv) Start() {
...
-	file, err := os.Open(e.Path + "/.ocenv")
+	path := filepath.Clean(e.Path + "/.ocenv")
+	file, err := os.Open(path)
 	if err != nil {
 		log.Fatal(err)
 	}
-	defer file.Close()
+	defer func() {
+		if err := file.Close(); err != nil {
+			fmt.Println("Error closing file: ", path)
```
fahlmant, line 159: return

cmd/env/env.go
```diff
@@ -175,7 +188,12 @@ func (e *OcEnv) killChildren() {
 		}
 		log.Fatalf("Failed to read file .killpids: %v", err)
 	}
-	defer file.Close()
+	defer func(file *os.File) {
+		err := file.Close()
+		if err != nil {
+			fmt.Println("Error while closing file: ", path)
```
fahlmant, line 194: return

cmd/env/env.go
```diff
@@ -198,15 +216,18 @@ func (e *OcEnv) killChildren() {
...
 		log.Printf("failed to delete .killpids, you may need to clean it up manually: %v\n", err)
 	}
 
 }
 func (e *OcEnv) Delete() {
 	fmt.Printf("Cleaning up OpenShift environment %s\n", e.Options.Alias)
-	os.RemoveAll(e.Path)
+	err := os.RemoveAll(e.Path)
+	if err != nil {
+		fmt.Println("Error while calling os.RemoveAll", err.Error())
```
fahlmant, line 229: return

cmd/env/env.go
```diff
@@ -234,14 +255,24 @@ PATH=` + e.Path + `/bin:` + os.Getenv("PATH") + `
 	if err != nil {
 		log.Fatal(err)
 	}
-	defer direnvfile.Close()
+	defer func(direnvfile *os.File) {
+		err := direnvfile.Close()
+		if err != nil {
+			fmt.Println("Error while calling direnvFile.Close(): ", err.Error())
```
fahlmant, line 261: return

cmd/env/env.go
```diff
@@ -234,14 +255,24 @@ PATH=` + e.Path + `/bin:` + os.Getenv("PATH") + `
...
 	zshenvfile := e.ensureFile(e.Path + "/.zshenv")
 	_, err = zshenvfile.WriteString("source .ocenv")
 	if err != nil {
 		log.Fatal(err)
 	}
-	defer direnvfile.Close()
+	defer func(direnvfile *os.File) {
+		err := direnvfile.Close()
+		if err != nil {
+			fmt.Println("Error while calling direnvFile.Close(): ", err.Error())
```
fahlmant, line 273: return

cmd/env/env.go
```diff
@@ -319,16 +350,21 @@ func (e *OcEnv) getLoginScript() string {
...
 func (e *OcEnv) createBin(cmd, content string) {
-	filepath := e.binPath() + "/" + cmd
-	scriptfile := e.ensureFile(filepath)
-	defer scriptfile.Close()
+	path := filepath.Join(e.binPath(), cmd)
+	scriptfile := e.ensureFile(path)
+	defer func(scriptfile *os.File) {
+		err := scriptfile.Close()
+		if err != nil {
+			fmt.Println("Error closing file: ", path)
```
fahlmant, line 358: return

cmd/env/env.go
```diff
@@ -319,16 +350,21 @@ func (e *OcEnv) getLoginScript() string {
...
 	_, err := scriptfile.WriteString(content)
 	if err != nil {
-		panic(fmt.Errorf("error writing to file %s: %v", filepath, err))
+		panic(fmt.Errorf("error writing to file %s: %v", path, err))
 	}
-	err = os.Chmod(filepath, 0744)
+	err = os.Chmod(path, 0700) //#nosec G302 -- Expect file permissions to be 0600 or less, not applicable here, because it's an executable
 	if err != nil {
-		log.Fatalf("Can't update permissions on file %s: %v", filepath, err)
+		log.Fatalf("Can't update permissions on file %s: %v", path, err)
```
fahlmant, line 367: return

cmd/federatedrole/cmd.go
```diff
@@ -21,5 +22,8 @@ func NewCmdFederatedRole(streams genericclioptions.IOStreams, flags *genericclio
 }
 
 func help(cmd *cobra.Command, _ []string) {
-	cmd.Help()
+	err := cmd.Help()
+	if err != nil {
+		fmt.Println("Error while calling cmd.Help()", err.Error())
```
fahlmant, line 27: return

cmd/network/cmd.go
```diff
@@ -20,5 +21,8 @@ func NewCmdNetwork(streams genericclioptions.IOStreams, flags *genericclioptions
 }
 
 func help(cmd *cobra.Command, _ []string) {
-	cmd.Help()
+	err := cmd.Help()
+	if err != nil {
+		fmt.Println("Error while calling cmd.Help(): ", err.Error())
```
fahlmant, line 26: return

cmd/options.go
```diff
@@ -15,7 +16,10 @@ func newCmdOptions(streams genericclioptions.IOStreams) *cobra.Command {
 		DisableAutoGenTag:     true,
 		DisableFlagsInUseLine: true,
 		Run: func(cmd *cobra.Command, args []string) {
-			cmd.Usage()
+			err := cmd.Usage()
+			if err != nil {
+				fmt.Println("Error while calling cmd.Usage(): ", err.Error())
```
fahlmant, line 21: return

cmd/servicelog/cmd.go
```diff
@@ -9,7 +10,10 @@ func NewCmdServiceLog() *cobra.Command {
 		Use:   "servicelog",
 		Short: "OCM/Hive Service log",
 		Run: func(cmd *cobra.Command, args []string) {
-			cmd.Help()
+			err := cmd.Help()
+			if err != nil {
+				fmt.Println("Error calling cmd.Help(): ", err.Error())
```
fahlmant, line 15: return

cmd/sts/cmd.go
```diff
@@ -21,5 +22,8 @@ func NewCmdSts(streams genericclioptions.IOStreams, flags *genericclioptions.Con
 }
 
 func help(cmd *cobra.Command, _ []string) {
-	cmd.Help()
+	err := cmd.Help()
+	if err != nil {
+		fmt.Println("Cannot print help")
```
fahlmant, line 27: return

cmd/env/env.go
```diff
@@ -198,15 +219,19 @@ func (e *OcEnv) killChildren() {
...
+		return
 	}
 
 }
 func (e *OcEnv) Delete() {
 	fmt.Printf("Cleaning up OpenShift environment %s\n", e.Options.Alias)
-	os.RemoveAll(e.Path)
+	err := os.RemoveAll(e.Path)
+	if err != nil {
+		fmt.Println("Error while calling os.RemoveAll", err.Error())
```
fahlmant, line 233: Looks like this is still missing a return

fahlmant (commented):
Generally lgtm, however
/hold
It might be good to consult the security team on the exceptions

shibumi:
@fahlmant I fixed the last two issues + I will ask the security team for a review.

cmd/env/env.go
```diff
@@ -141,13 +146,21 @@ func (e *OcEnv) Start() {
 	shell := os.Getenv("SHELL")
 
 	fmt.Print("Switching to OpenShift environment " + e.Options.Alias + "\n")
-	cmd := exec.Command(shell)
+	// ignore the following line in linter, only way to fix this is via setting a
+	// constant string for the exec.Command
+	cmd := exec.Command(shell) //nolint:gosec
```
jaybeeunix, line 151: for these few items w/o a specific test disabled, is that because it wouldn't allow you to disable the specific test?

thx for adding the specific tests to the nosecs, too!

shibumi: @jaybeeunix  I have updated the 3 nolints and replaced them with a //#nosec line.. this allows us pinning to one specific issue

pkg/provider/aws/sts_test.go
```diff
@@ -302,7 +302,7 @@ func TestGetSignInToken(t *testing.T) {
 			title: "Server returns 200 but bad body format",
 			handler: func(w http.ResponseWriter, r *http.Request) {
 				w.WriteHeader(200)
-				w.Write([]byte("foo"))
+				w.Write([]byte("foo")) //#nosec G104 -- Errors unhandled, just a test
```
dastergon, line 305: Can't we just ignore G104 for all tests? Is there a direct to add at the top instead of adding this line in all occurrences?

.golangci.yaml
```diff
@@ -0,0 +1,17 @@
...
+  exclude-use-default: false
+
+# individual linter configs go here
+linters-settings:
+
+# default linters are enabled `golangci-lint help linters`
+linters:
+  disable-all: true
+  enable:
+    - gosec
```
dastergon, line 17: needs an extra empty line

cmd/account/cmd.go
```diff
@@ -38,5 +39,9 @@ func NewCmdAccount(streams genericclioptions.IOStreams, flags *genericclioptions
 }
 
 func help(cmd *cobra.Command, _ []string) {
-	cmd.Help()
+	err := cmd.Help()
+	if err != nil {
+		fmt.Println("Error while calling cmd.Help(): ", err.Error())
+		return
```
Nikokolas3270, line 45: `return` is not needed as this is the last instruction of the function.
Also this would be consistent with `aao.help` function in which there is no `return` in case of error.

cmd/account/get/account.go
```diff
@@ -117,10 +117,10 @@ func (o *getAccountOptions) run() error {
 			return err
 		}
 
-		for _, a := range accounts.Items {
+		for i, a := range accounts.Items {
 			if a.Spec.AwsAccountID == o.accountID {
 				accountCRName = a.Name
-				account = &a
+				account = &accounts.Items[i]
```
Nikokolas3270, line 123: I guess the idea here is to have `account` point an item of the list and not a copy of an item. This is great for consistency even if this was not causing any issue (except for `gosec`).
However each item of the list is still iteratively copied anyway in `a`; also code is not very consistent, what about:
```
for i := range accounts.Items {
    a := &accounts.Items[i]
    if a.Spec.AwsAccountID == o.accountID {
        accountCRName = a.Name
        account = &a
```

Nikokolas3270:
Code seems fine, thanks for improving the code quality 👍 
For the 2 points I reported, there are several occurrences of them in the code. Please fix them all.
I guess that in the future we want to enable `gosec` linter in the builder to keep sure the code remains correct for this linter. Is there some ticket tracking that?

pkg/provider/aws/sts_test.go
shibumi: Yes, we can ignore all G104 errors via the golangci-lint configuration file.

cmd/account/cmd.go
shibumi: This makes sense, I just followed Taylor's advice here to include it

shibumi:
> I guess that in the future we want to enable gosec linter in the builder to keep sure the code remains correct for this linter. Is there some ticket tracking that?

I think there is an epic for gosec preparations.. if this is complete the plan is to enforce gosec for all projects.

.golangci.yaml
shibumi: fixed.

shibumi:
/unhold

Please lgtm or hold again if you are unhappy

fahlmant (approved):
/approve
/lgtm

cmd/account/get/account.go
Nikokolas3270: Sorry didn't see your reply till now.
Yes I agree this change was mostly cosmetic, so not a must have indeed and anyway the PR has been merged without it.
I did not test if the code I proposed was compiling but if it was `*a.Name` and `a.Name` would point the same string so I do not see the risk.

PR #226, opened 2022-07-06
Initial cluster context command

sam-nguyen7:
PR introduces a new command. `osdctl cluster context` which takes in either an internal clusterID, external clusterID, or clustername and returns whether the cluster is fully supported or in limited support. 

Moving forward we will be adding more functionalities to this command such as reporting on triggered alerts, Sevicelogs, and more.


fahlmant:
@sam-nguyen7 Could this be folded into `osdctl cluster health`?

clcollins:
@fahlmant this is intended to be a quick-reference of state based on external tools.  We're planning to extend it over time.

fahlmant (approved):
/lgtm
/approve

PR #227, opened 2022-07-12
Added error checking, extracted token and connetion builders to use t…

luis-falcon:
Added error checking, extracted token and connetion builders to use the utils function, all support commands can use any cluster identifyer, etc

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -5,7 +5,16 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+
+	sdk "github.com/openshift-online/ocm-sdk-go"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 )
+
+type lmtSprReasonItem struct {
+	ID      string
+	Summary string
+	Details string
+}
 
 func GetOCMAccessToken() (*string, error) {
 	// Get ocm access token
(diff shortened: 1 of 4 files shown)


pkg/utils/utils.go
```diff
@@ -19,3 +28,125 @@ func GetOCMAccessToken() (*string, error) {
...
+		Send()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get limited Support Reasons: %s", err)
+	}
+
+	lmtReason := limitedSupportReasons.Items()
+
+	var clusterLmtSprReasons []*lmtSprReasonItem
+
+	lmtReason.Each(func(lmtSprReason *cmv1.LimitedSupportReason) bool {
```
fahlmant, line 140: Everything looks good here, just curious what this Each is doing?

fahlmant: Why was this resolved?

fahlmant, line 140: What is this each doing?

luis-falcon: you asked a question. I answered, what else was there to do to? Am I missing something here? is that not how it is supposed to be done? Or is the Conversation starting party the one that should be "resolving"?  :thinking:

fahlmant: It was resolved before you actually posed the response :) 

Screenshot from 2022-07-13 11-41-46

luis-falcon: ahhh. Ok cool :/ is good to know

luis-falcon: Hey @fahlmant this is because we get a list when we get the limited support reasons. This is because a cluster can have many reasons why it is on limited support. Thus, for each of those reasons (1 or many) on that limited support list, we create a reason object and append them to the slice we will return. We do this Right here instead of let's say just return the full list to the code calling this function because we also want to get only the pertinent properties (the ones we care about) for the limited support reasons and because that way if we need this code at some point we don't need to parse that list, we just get the slice of reasons we need and job done. And if we want more or less at some point we can modify this struct and the .each part of the code.

So In summary, we looping through each limited support reason and extracting only the info we need.

pkg/utils/utils.go
```diff
@@ -19,3 +28,125 @@ func GetOCMAccessToken() (*string, error) {
...
+		clusterLmtSprReason := lmtSprReasonItem{
+			ID:      lmtSprReason.ID(),
+			Summary: lmtSprReason.Summary(),
+			Details: lmtSprReason.Details(),
+		}
+		clusterLmtSprReasons = append(clusterLmtSprReasons, &clusterLmtSprReason)
+
+		return true
+	})
```
fahlmant, line 149:
```suggestion
	for _, resp := range lmtReason {
		clusterLmtSprReason := lmtSprReasonItem{
			ID:      resp.ID(),
			Summary: resp.Summary(),
			Details: resp.Details(),
		}
		clusterLmtSprReasons = append(clusterLmtSprReasons, &clusterLmtSprReason)
	}
```

I think putting this directly in a loop is better than using an anonymous function if we're only ever going to return true

luis-falcon: I tried like this, this will not work, the range function is trying then to iterate through a pointer which is what the lmtReason is.

fahlmant: Looks like you can use `UnmarshalLimitedSupportReasonList` to get a list of limited support reasons, and then iterate through that. 

 https://github.com/openshift-online/ocm-sdk-go/blob/main/clustersmgmt/v1/limited_support_reason_list_type_json.go#L56

fahlmant (approved):
/approve
/lgtm

PR #228, opened 2022-07-12
Added support for external cluster id and cluster name to -C flag

lnguyen1401:
Now osdctl command can use external cluster ID and cluster name in addition to internal cluster ID when passing via -C flag. for example:  osdctl account cli -C $(cluster internal_id/external_id/display_name) -p $profile

Jira ticket: https://issues.redhat.com/browse/OSD-10463


iamkirkbater:
/hold

Just in case anyone else wants to take a look.

pkg/k8s/clusterresourcefactory.go
```diff
@@ -41,7 +42,7 @@ func (factory *ClusterResourceFactoryOptions) AttachCobraCliFlags(cmd *cobra.Com
 		"The namespace to keep AWS accounts. The default value is aws-account-operator.")
 	cmd.Flags().StringVarP(&factory.AccountName, "account-name", "a", "", "The AWS account CR we need to create a temporary AWS console URL for")
 	cmd.Flags().StringVarP(&factory.AccountID, "account-id", "i", "", "The AWS account ID we need to create AWS credentials for -- This argument will not work for CCS accounts")
-	cmd.Flags().StringVarP(&factory.ClusterID, "cluster-id", "C", "", "The Internal Cluster ID from Hive to create AWS console URL for")
+	cmd.Flags().StringVarP(&factory.ClusterID, "cluster-identifier", "C", "", "The Internal Cluster ID/External Cluster ID/ Cluster Name from Hive to create AWS console URL for")
```
iamkirkbater, line 45: Actually, I think we need to add an alias here if we're going to change from `cluster-id` to `cluster-identifier` as that would constitute a breaking change and could break any scripts that use this flag.

iamkirkbater (changes requested):

iamkirkbater:
/unapprove

Jumped the gun here a bit XD

pkg/k8s/clusterresourcefactory.go
lnguyen1401: Hey Kirk! would it be ok if i can leave the cluster-id as is ?

iamkirkbater:
/lgtm

Only being "held" to allow others time to review as well.  Please feel free to remove when appropriate.

fahlmant (approved):
/approve
/lgtm
/unhold

PR #230, opened 2022-07-20
[OSD-12483]Adding SL output to cluster context command

dkeohane:
Ticket Ref: https://issues.redhat.com/browse/OSD-12483

Pulling Servicelogs as part of the cluster context command
- Only displaying the Error Servicelogs
- Shows SLs for the past 30 days by default, but the days value is configurable
- Shows only the SL summary by default, will show Summary, Description, Creation Timestamp and Severity when viewed in verbose mode.

### DIFF
--- a/cmd/cluster/context.go
+++ b/cmd/cluster/context.go
@@ -2,12 +2,17 @@ package cluster
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log"
 	"os"
+	"time"
 
+	"github.com/openshift-online/ocm-cli/pkg/dump"
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/osdctl/cmd/servicelog"
+	sl "github.com/openshift/osdctl/internal/servicelog"
 	"github.com/openshift/osdctl/internal/utils/globalflags"
 	"github.com/openshift/osdctl/pkg/printer"
 	"github.com/openshift/osdctl/pkg/utils"
@@ -20,6 +25,7 @@ type statusOptions struct {
 	output    string
 	verbose   bool
 	clusterID string
+	days      int
 
 	genericclioptions.IOStreams
 	GlobalOptions *globalflags.GlobalOptions
@@ -31,8 +37,8 @@ type limitedSupportReasonItem struct {
 	Details string
 }
 
-// newCmdcontext implements the context command to show the current context of a cluster
-func newCmdcontext(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags, globalOpts *globalflags.GlobalOptions) *cobra.Command {
+// newCmdContext implements the context command to show the current context of a cluster
+func newCmdContext(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags, globalOpts *globalflags.GlobalOptions) *cobra.Command {
 	ops := newStatusOptions(streams, flags, globalOpts)
 	statusCmd := &cobra.Command{
 		Use:               "context",
@@ -45,6 +51,7 @@ func newCmdcontext(streams genericclioptions.IOStreams, flags *genericclioptions
 		},
 	}
 	statusCmd.Flags().BoolVarP(&ops.verbose, "verbose", "", false, "Verbose output")
+	statusCmd.Flags().IntVarP(&ops.days, "days", "d", 30, "Command will display X days of Error SLs sent to the cluster. Days is set to 30 by default")
 
 	return statusCmd
 }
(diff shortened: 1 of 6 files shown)


cmd/cluster/context.go
```diff
@@ -130,6 +159,8 @@ func checkSupportStatus(response *cmv1.LimitedSupportReasonsListResponse) error
 		return true
 	})
 
+	fmt.Printf("\n============================================================\n Cluster Limited Support Status \n============================================================\n")
```
NautiluX, line 162: I feel like Println could make this more readable

cmd/cluster/context.go
```diff
@@ -147,3 +178,76 @@ func checkSupportStatus(response *cmv1.LimitedSupportReasonsListResponse) error
...
+
+		// If the days since the SL was sent exceeds o.days days, we're not interested
+		if (time.Since(serviceLog.CreatedAt).Hours() / 24) > float64(o.days) {
+			continue
+		}
+
+		erroredServiceLogs = append(erroredServiceLogs, serviceLog)
+	}
+
+	fmt.Printf("\n============================================================\n Service Logs with Error Severity sent in the past %d Days:\n============================================================\n", o.days)
```
NautiluX, line 230: I feel like Println could make this more readable

cmd/cluster/context.go
```diff
@@ -102,20 +106,45 @@ func (o *statusOptions) run() error {
...
-	response, err := resource.List().SendContext(ctx)
+	lsResponse, err := getLimitedSupportReasons(connection, ctx, o.clusterID)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Can't retrieve cluster limited support reasons: %v\n", err)
 		os.Exit(1)
 	}
 
 	// Check support status of cluster
-	checkSupportStatus(response)
+	checkSupportStatus(lsResponse)
```
NautiluX, line 116:
```suggestion
	printSupportStatus(lsResponse)
```
I think the current function name is a bit misleading, it doesn't return anything but get and print the supportstatus

cmd/cluster/context.go
```diff
@@ -102,20 +106,45 @@ func (o *statusOptions) run() error {
...
+
+	// Retrieve and Print the Servicelogs for this cluster
+	err = o.getServiceLogs()
```
NautiluX, line 119: I think the current function name is a bit misleading, it doesn't return anything but get and print the SLs
```suggestion
	err = o.printServiceLogs()
```

cmd/cluster/context.go
```diff
@@ -147,3 +178,76 @@ func checkSupportStatus(response *cmv1.LimitedSupportReasonsListResponse) error
...
+	cluster := clusters[0]
+
+	var erroredServiceLogs []sl.ServiceLogShort
+	serviceLogListAllMessagesFlag := false
+
+	// Now get the SLs for the cluster
+	slResponse, err := sendRequest(servicelog.CreateListSLRequest(ocmClient, cluster.ExternalID(), serviceLogListAllMessagesFlag))
+	if err != nil {
+		return err
+	}
```
NautiluX, line 205: isn't this whole section necessary for the `serviclelog list` command as well? Is it possible to reuse/unify it?

dkeohane: Yeah it could be - the main difference between the 2 is that `serviclelog list` gets the response and just dumps it out, so all SLs. I'm filtering the SLs here. The `dump.pretty` adds colour to the `servicelog list` command, but I haven't been able to get it to work for the cluster context cmd.

macgregor: I think hes saying you could have something like:
```
func fetchServiceLogs(string clusterId, bool listAllMessagesFlag) sl.ServiceLogShortList, error {
        // Create OCM client to talk to cluster API
	ocmClient := utils.CreateConnection()
	defer func() {
		if err := ocmClient.Close(); err != nil {
			fmt.Printf("Cannot close the ocmClient (possible memory leak): %q", err)
		}
	}()

        ...

        var serviceLogs sl.ServiceLogShortList
	err = json.Unmarshal(slResponse.Bytes(), &serviceLogs)
	if err != nil {
		fmt.Printf("Failed to unmarshal the SL response %q\n", err)
		return err
	}
        return serviceLogs
}
```

And call that in both places. Each caller can format the output however they want.

dkeohane: done

cmd/cluster/context.go
```diff
@@ -61,6 +68,10 @@ func (o *statusOptions) complete(cmd *cobra.Command, args []string) error {
 		return cmdutil.UsageErrorf(cmd, "Provide exactly one cluster ID")
 	}
 
+	if o.days < 1 {
```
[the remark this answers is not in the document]
dkeohane: I felt it was overkill, leaving it to the discretion of people not to try break this - famous last words ;)

sam-nguyen7 (approved):
/lgtm
/approve

Appreciate the refactor as well. Thanks!

PR #231, opened 2022-07-21
Unit test for cluster support command

luis-falcon:
Created unit test for the cluster support command functions.

### DIFF
--- a/cmd/cluster/support/delete.go
+++ b/cmd/cluster/support/delete.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/openshift-online/ocm-cli/pkg/arguments"
 	sdk "github.com/openshift-online/ocm-sdk-go"
-	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/openshift/osdctl/internal/support"
 	"github.com/openshift/osdctl/internal/utils/globalflags"
 	ctlutil "github.com/openshift/osdctl/pkg/utils"
@@ -105,7 +104,7 @@ func (o *deleteOptions) run() error {
 		os.Exit(1)
 	}
 
-	deleteRequest, err := createDeleteRequest(connection, cluster, o.limitedSupportReasonID)
+	deleteRequest, err := createDeleteRequest(connection, cluster.ID(), o.limitedSupportReasonID)
 	if err != nil {
 		fmt.Printf("failed post call %q\n", err)
 	}
@@ -123,11 +122,11 @@ func (o *deleteOptions) run() error {
 }
 
 // createDeleteRequest sets the delete API and returns a request
-func createDeleteRequest(ocmClient *sdk.Connection, cluster *v1.Cluster, reasonID string) (request *sdk.Request, err error) {
+func createDeleteRequest(client SDKConnectionClient, clusterID string, reasonID string) (request *sdk.Request, err error) {
 
-	targetAPIPath := "/api/clusters_mgmt/v1/clusters/" + cluster.ID() + "/limited_support_reasons/" + reasonID
+	targetAPIPath := "/api/clusters_mgmt/v1/clusters/" + clusterID + "/limited_support_reasons/" + reasonID
 
-	request = ocmClient.Delete()
+	request = client.Delete()
 	err = arguments.ApplyPathArg(request, targetAPIPath)
 	if err != nil {
 		return nil, fmt.Errorf("cannot parse API path '%s': %v", targetAPIPath, err)
--- a/cmd/cluster/support/delete_test.go
+++ b/cmd/cluster/support/delete_test.go
@@ -1,0 +1,39 @@
+package support
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestCreateDeleteRequest(t *testing.T) {
+	g := NewGomegaWithT(t)
+
+	testCases := []struct {
+		title          string
+		clusterID      string
+		lmtSprReasonID string
+		client         *Client
+		errExpected    bool
+	}{
+		{
+			title:          "Delete Request creation",
+			clusterID:      "5a5a5a5a-5a5a-5a5a-5a5a-5a5a5a5a5a5a",
+			lmtSprReasonID: "FAFAFAFA-FAFAFAFA-FAFAFAFA-FAFAFAFA",
+			client: &Client{
+				name: "fakeSDKClient",
+			},
+			errExpected: false,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.title, func(t *testing.T) {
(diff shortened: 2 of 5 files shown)


cmd/cluster/support/delete.go
```diff
@@ -105,7 +104,7 @@ func (o *deleteOptions) run() error {
 		os.Exit(1)
 	}
 
-	deleteRequest, err := createDeleteRequest(connection, cluster, o.limitedSupportReasonID)
+	deleteRequest, err := createDeleteRequest(connection, cluster.ID(), o.limitedSupportReasonID)
```
fahlmant, line 107: Does this need to be changed to be tested?

cmd/cluster/support/delete.go
```diff
@@ -123,11 +122,11 @@ func (o *deleteOptions) run() error {
 }
 
 // createDeleteRequest sets the delete API and returns a request
-func createDeleteRequest(ocmClient *sdk.Connection, cluster *v1.Cluster, reasonID string) (request *sdk.Request, err error) {
+func createDeleteRequest(client SDKConnectionClient, clusterID string, reasonID string) (request *sdk.Request, err error) {
```
fahlmant, line 125: ocmClient seemed more specific, why did you change it? 

Also does this work on a real cluster? I think deleting by external ID doesn't work, but I see that's what your test is running. Can this be changed back so we aren't altering the behavior we're testing as well?

cmd/cluster/support/delete_test.go
```diff
@@ -0,0 +1,39 @@
...
+		title          string
+		clusterID      string
+		lmtSprReasonID string
+		client         *Client
+		errExpected    bool
+	}{
+		{
+			title:          "Delete Request creation",
+			clusterID:      "5a5a5a5a-5a5a-5a5a-5a5a-5a5a5a5a5a5a",
+			lmtSprReasonID: "FAFAFAFA-FAFAFAFA-FAFAFAFA-FAFAFAFA",
```
fahlmant, line 22: Do these match the pattern that the API uses? Can you show an example?

cmd/cluster/support/delete_test.go
```diff
@@ -0,0 +1,39 @@
...
+			client: &Client{
+				name: "fakeSDKClient",
+			},
+			errExpected: false,
```
fahlmant, line 26: Can you also create a tests that expects an error?

cmd/cluster/support/post.go
```diff
@@ -144,11 +146,11 @@ func (o *postOptions) run() error {
 
 // createPostRequest create and populates the limited support post call
 // swagger code gen: https://api.openshift.com/?urls.primaryName=Clusters%20management%20service#/default/post_api_clusters_mgmt_v1_clusters__cluster_id__limited_support_reasons
-func createPostRequest(ocmClient *sdk.Connection, cluster *v1.Cluster) (request *sdk.Request, err error) {
+func createPostRequest(client SDKConnectionClient, clusterID string) (request *sdk.Request, err error) {
```
fahlmant, line 149: Same questions here as above; why change to client, and does this work as expected with external IDs?

cmd/cluster/support/post_test.go
```diff
@@ -0,0 +1,131 @@
...
+		title       string
+		option      *postOptions
+		errExpected bool
+	}{
+		{
+			title: "No template",
+			option: &postOptions{
+				GlobalOptions: &globalflags.GlobalOptions{},
+			},
+			errExpected: true,
```
fahlmant, line 23: I'm not familiar enough with these templates to know why they would fail. Can you leave a short comment explaining why this should fail? Can you also add a success case?

cmd/cluster/support/post_test.go
```diff
@@ -0,0 +1,131 @@
...
+		path        string
+		errExpected bool
+	}{
+		{
+			title:       "File not existant",
+			path:        "/tmp/fakeFile.json",
+			errExpected: true,
+		},
+		{
+			title:       "File existant",
```
fahlmant, line 52:
```suggestion
			title:       "File exists",
```

cmd/cluster/support/post_test.go
```diff
@@ -0,0 +1,131 @@
...
+		client      *Client
+		errExpected bool
+	}{
+		{
+			title:     "Post Request creation",
+			clusterID: "5a5a5a5a-5a5a-5a5a-5a5a-5a5a5a5a5a5a",
+			client: &Client{
+				name: "fakeSDKClient",
+			},
+			errExpected: false,
```
fahlmant, line 84: Can you also add a test that's expected to fail?

cmd/cluster/support/tests_helpers.go
```diff
@@ -0,0 +1,21 @@
+package support
+
+import sdk "github.com/openshift-online/ocm-sdk-go"
+
+//interface that must be met to be a SDK connection
+type SDKConnectionClient interface {
```
fahlmant, line 6: This package is nice, but is it required for these tests to run? Can this refactor be done in a separate PR?

fahlmant (changes requested):
Given that the title states that this is adding unit tests, I'm a bit surprised to find so many other changes. Is it possible to spin out the code changes to another PR, and then have this PR follow on? 

Also, would you mind adding a small comment for each tests to explain why it should fail or succeed? Its not clear just looking at the tests right now.

cmd/cluster/support/delete.go
fahlmant: You cannot delete by external ID in OCM, only by internal ID. So I think leaving this as `cluster *v1.Cluster` is better here because we KNOW a v1.Cluster has the internal ID as part of it, where as with a string, there's no guarantee about the data passed in.

cmd/cluster/support/tests_helpers.go
fahlmant: Why is it in a file called `test_helpers` then? Can that be clarified? 

Also, as mentioned earlier, could this PR be done separately so that this refactor of how our functions work isn't lumped together with unit tests?

cmd/cluster/support/delete.go
fahlmant: think leaving this as cluster *v1.Cluster is better here because we KNOW a v1.Cluster has the internal ID as part of it, where as with a string, there's no guarantee about the data passed in.

cmd/cluster/support/delete_test.go
luis-falcon: Yes they do. But to make sure I added a validating function with a regEx, in order to make sure they match that allowed me to create a test that fails when i submit a bad ID. To be sumited with other changes shortly

cmd/cluster/support/delete_test.go
luis-falcon: Also done to be submited shortly.

PR #233, opened 2022-07-27
OSD-11989 mock interface and refactors.

luis-falcon:
mock sdk connection basic requests (post, delete), refactor code to accommodate api mocking for unit tests, added validation to clusterID format before sending requests to API.


clcollins:
/lgtm
/approve

Thanks for the run-through, too,, @luis-falcon!

PR #235, opened 2022-07-29
Handle error from FetchServiceLogs properly

fahlmant:
Small fix to handle the error from FetchServiceLogs before handling the error from the Pretty print in `ocm servicelog list`


dkeohane (approved):
/lgtm :taco:

PR #236, opened 2022-07-29
Remove unused metrics cmd

sam-nguyen7:
`osdctl metrics` used to display aws account operator metrics. This is deprecated as we should be using ocm-backplane to view cluster metrics.


fahlmant (approved):
/approve
/lgtm

fahlmant (approved):
/approve
/lgtm

PR #237, opened 2022-08-01
Create capability add and remove commands

fahlmant:
This PR adds two new commands, 
```
osdctl capability add [hibernation|autoscaling|ovn|upgradeChannelChange] [-g $OrgID | -b $SubscriptionID]

osdctl capability remove [hibernation|autoscaling|ovn|upgradeChannelChange] [-g $OrgID | -b $SubscriptionID]
```
This allows for adding and removing a specific capability to an OCM Organization or Subscription. The capabilities initially supported are: hibernation, autoscaling, ovn, and upgradeChannelChange.

Ref: https://issues.redhat.com/browse/OSD-12627

### DIFF
--- a/cmd/capability/add.go
+++ b/cmd/capability/add.go
@@ -1,0 +1,110 @@
+package capability
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/openshift/osdctl/pkg/utils"
+	"github.com/spf13/cobra"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+type addOptions struct {
+	OrganizationID string
+	SubscriptionID string
+}
+
+func newAddCmd() *cobra.Command {
+	ops := addOptions{}
+	addCmd := &cobra.Command{
+		Use:   "add [capability] -g [organization ID]",
+		Short: "adds a specific capability to a specific OCM organization.\nAvailable capabilities: hibernation, autoscaling, ovn, upgradeChannelChange",
+		Args:  cobra.ExactArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(ops.complete(cmd, args))
+			cmdutil.CheckErr(ops.run(cmd, args[0]))
+		},
+	}
+
+	addCmd.Flags().StringVarP(&ops.OrganizationID, "organization-id", "g", "", "Specify an OCM Organization to apply a capability to")
+	addCmd.Flags().StringVarP(&ops.SubscriptionID, "subscription-id", "b", "", "Specify a Subscription to apply a capability to")
+
+	return addCmd
+}
+
+func (o *addOptions) complete(cmd *cobra.Command, args []string) error {
+	if len(args) == 0 {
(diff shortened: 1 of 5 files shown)


cmd/capability/cmd.go
```diff
@@ -0,0 +1,69 @@
...
+	// Add subcommands
+	capabilityCmd.AddCommand(newAddCmd())
+	capabilityCmd.AddCommand(newRemoveCmd())
+
+	return capabilityCmd
+}
+
+func getCapabilityKey(capability string) string {
+
+	switch capability {
```
iamkirkbater, line 57: This caught me up at first, thinking these were somehow magic strings.  Would this be easier to maintain if it is depicted as a map[string]string and checking if a key exists?

fahlmant: This is indeed basically a fake map. I tried making a var of type map[string]string at the package level, however Go doesn't allow this. In order to create a map that could be used, we'd need a function that builds the map and returns it to the caller, then the caller would need to attempt to get the value out with the key. This is a shortcut that recreates the functionality of a map, using constant string values to make sure there's no misspellings. I'll leave a comment explaining this

iamkirkbater (commented):
/lgtm

cmd/capability/remove.go
```diff
@@ -0,0 +1,102 @@
...
+	defer func() {
+		if err := ocmClient.Close(); err != nil {
+			fmt.Printf("Cannot close the ocmClient (possible memory leak): %q", err)
+		}
+	}()
+
+	// Build the base of the request path
+	var base string
+	if o.OrganizationID != "" {
+		// If the organization
```
sam-nguyen7, line 80: If the organization ID is set

cmd/capability/remove.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	// Ensure that the capability selected is valid
+	availableCapabilities := []string{hibernation, autoscaling, ovn, upgradeChannelChange}
+	for i, s := range availableCapabilities {
+		if args[0] == s {
+			break
+		}
+		if i == len(availableCapabilities)-1 {
+			return fmt.Errorf("invalid capability specified. Avaialbe capabilites are hibernation, autoscaling, ovn and upgradeChannelChange")
```
sam-nguyen7, line 50: Available

cmd/capability/add.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	// Ensure that the capability selected is valid
+	availableCapabilities := []string{hibernation, autoscaling, ovn, upgradeChannelChange}
+	for i, s := range availableCapabilities {
+		if args[0] == s {
+			break
+		}
+		if i == len(availableCapabilities)-1 {
+			return fmt.Errorf("invalid capability specified. Avaialbe capabilites are hibernation, autoscaling, ovn and upgradeChannelChange")
```
sam-nguyen7, line 51: Available

sam-nguyen7 (commented):
/lgtm

cmd/capability/add.go
clcollins: Fixed in #239 :smile:

PR #241, opened 2022-08-08
OSD-11989 Added unit test for support post command functions

luis-falcon:
Unit tests to pertinent functions on the post command.

### DIFF
--- a/cmd/cluster/support/post_test.go
+++ b/cmd/cluster/support/post_test.go
@@ -1,0 +1,93 @@
+package support
+
+import (
+	"testing"
+
+	"github.com/openshift/osdctl/internal/support"
+)
+
+func TestValidateBadResponse(t *testing.T) {
+
+	//good json format
+	gMsg := "{\"field1\":\"data1\",\"field2\":\"data2\"}"
+
+	//bad json format
+	bMsg := "{\"field1\":\"data1\",\"field2\"\"data2\"}"
+
+	testCases := []struct {
+		title       string
+		body        []byte
+		errExpected bool
+		errReason   string
+	}{
+		{
+			title:       "Validate good Json Response",
+			body:        []byte(gMsg),
+			errExpected: false,
+		},
+		{
+			title:       "Validate bad Json Response",
+			body:        []byte(bMsg),
+			errExpected: true,
+			errReason:   "Server returned invalid JSON",
+		},
+	}
+	for _, tc := range testCases {
+		_, result := validateBadResponse(tc.body)
+		if tc.errExpected {
+			if result == nil {
+				t.Fatalf("Test %s failed. Expected error %s, but got none", tc.title, tc.errReason)
+			}
+			if result.Error() != tc.errReason {
+				t.Fatalf("Test %s failed. Expected error %s, but got %s", tc.title, tc.errReason, result.Error())
+			}
+		}
+		if !tc.errExpected && result != nil {
+			t.Fatalf("Test %s failed. Expected no errors, but got %s", tc.title, result.Error())
+		}
+	}
+}
+
+func TestValidateGoodResponse(t *testing.T) {
+
+	//good json format
+	gMsg := "{\"field1\":\"data1\",\"field2\":\"data2\"}"
+	//bad json format
+	bMsg := "{\"field1\":\"data1\",\"field2\"\"data2\"}"
+
+	testCases := []struct {
+		title        string
+		body         []byte
+		lmtSprReason support.LimitedSupport
+		errExpected  bool
+		errReason    string
+	}{
+		{
+			title:        "Validate good Json Response",
+			body:         []byte(gMsg),
+			lmtSprReason: support.LimitedSupport{},
+			errExpected:  false,
+		},
+		{
+			title:        "Validate bad Json Response",
+			body:         []byte(bMsg),
+			lmtSprReason: support.LimitedSupport{},
+			errExpected:  true,
+			errReason:    "Server returned invalid JSON",
+		},
+	}
+	for _, tc := range testCases {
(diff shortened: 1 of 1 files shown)


cmd/cluster/support/post_test.go
```diff
@@ -0,0 +1,115 @@
...
+	//good json format
+	gMsg := "{\"field1\":\"data1\",\"field2\":\"data2\"}"
+
+	//bad json format
+	bMsg := "{\"field1\":\"data1\",\"field2\"\"data2\"}"
+
+	testCases := []struct {
+		title       string
+		body        []byte
+		errExpected bool
```
fahlmant, line 53: Since there are two different errors returned from the function, making sure that the error returned is the one expected may be good to add here.

cmd/cluster/support/post_test.go
```diff
@@ -0,0 +1,115 @@
...
+	//good json format
+	gMsg := "{\"field1\":\"data1\",\"field2\":\"data2\"}"
+	//bad json format
+	bMsg := "{\"field1\":\"data1\",\"field2\"\"data2\"}"
+
+	testCases := []struct {
+		title        string
+		body         []byte
+		lmtSprReason support.LimitedSupport
+		errExpected  bool
```
fahlmant, line 90: Same as above, since there are two different errors returned from the function, making sure that the error returned is the one expected may be good to add here.

cmd/cluster/support/post_test.go
```diff
@@ -0,0 +1,115 @@
...
+		{
+			title:       "File not existant",
+			path:        "/tmp/fakeFile.json",
+			errExpected: true,
+		},
+		{
+			title:       "File existant",
+			path:        "./post.go",
+			errExpected: false,
+		},
```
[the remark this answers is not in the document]
luis-falcon: That would be in my opinion a unit test built for the utils package not for the support. We would want to test the "IsValidUrl" Function in the utils package Not this one. Making that test on this function would go curl the actual URL every time we run a test. I do not think this is the place to check for that.

cmd/cluster/support/post_test.go
luis-falcon: Same as above. This would be if we were maybe testing for the URL. But I do not test for the URL path in here.

luis-falcon (commented):
Check comments for answers.

cmd/cluster/support/post_test.go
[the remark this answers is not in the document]
fahlmant: We need to test all valid inputs of this function if we're going to unit test it here. This function was the one stripping the / out of the URL, not IsValidURL

cmd/cluster/support/post_test.go
fahlmant: I'm not referring to URLs here at all. There are two different errors that this function can return, 
https://github.com/openshift/osdctl/blob/master/cmd/cluster/support/post.go#L248
and
https://github.com/openshift/osdctl/blob/master/cmd/cluster/support/post.go#L251

cmd/cluster/support/post_test.go
fahlmant: Same as above, this function returns two different errors

fahlmant (changes requested):
Looking at this cluster, I believe testing `accessFile` isn't helpful as the actual logic being tested is whether the file path is valid or not, which is specifically from the output of FileExists. The unti tests should instead be on that function then, rather accessFile

cmd/cluster/support/post_test.go
fahlmant: In fact, there's a good example on how to check this here: https://github.com/openshift/osdctl/blob/master/cmd/capability/capability_test.go#L120

fahlmant: If there's no way to trigger the specific errors that's fine. If we're adding unit tests for a function, testing as many cases as possible is the goal.

luis-falcon: I do not agree with this. If you see on the code, we are basically testing the "Complete function" of the cobra command, not particular functions on our code. Which is basically testing that we pass parameters and if we dont that we get the "proper error", and The way it is doing it is basically telling me hey if is not nil I should get this string i hardcoded on the test. I do not think this is adding any value to the testing any function. As a test I don't care which string is not running. I care that all of a sudden I change code and the unit test is not passing.

example this part here :
```
 {
			Name:          "Providing no arguments",
			ErrorExpected: true,
			ErrorReason:   "capability was not provided, please specifiy which capability to add",
			Args:          []string{},
			RemoveOptions: removeOptions{OrganizationID: "myOrganization"},
		},
```
		
		the error reason has the string that we supposed to get as an error from the complete if we do not pass the right capability to the command itself. Do I care where the function failed? Not really, because I am not creating a test that should accommodate a function, I am trying to code a function that Should pass a test.
		
		the test is saying hey if it is not nil then it should have been this particular string on this particular part of the code. I dont know maybe I am failing to see something here. I will try to Recreate this code.
		
		In addition this is not using gomega at all which i believe was new standard for all tests we created.

fahlmant: The main goal of unit tests is testing specific function's logic. For a specific input, does the function return the expected output? That's what the linked code is doing. For example, if a wrong capability is given with a valid organization, but the error returned is "Invalid organization", then there's something wrong in the logic of the function. A unit test should cover as many possible cases as possible and that includes error states as there are often edge cases around errors that may not be caught correctly, or otherwise cause unintended behavior. If it doesn't make sense in this particular case then that's ok, but testing that a specific input causes a specific error is a valid use of unit tests imo.

luis-falcon (commented):
made the changes to the for loop to account for different errors.

luis-falcon (commented):
changes made

PR #245, opened 2022-08-23
[OSD-12763] capture different interface for ovn and sdn

bmeng:
update the capture subcommand to detect the correct interface for tcpdump based on network type


mrbarge:
I've successfully tested this on both OVN and SDN clusters and confirmed that it operates correctly on both.
/lgtm

bmeng:
@clcollins @fahlmant @iamkirkbater any of you could help review this? Thanks!

fahlmant (approved):
/approve
/lgtm

PR #246, opened 2022-08-26
new command for transfering cluster owner

RaphaelBut:
This PR tracks implementation of [OSD-12576](https://issues.redhat.com//browse/OSD-12576) Automating the ocm information update of a cluster ownership transfer.

### DIFF
--- a/cmd/cluster/transferowner.go
+++ b/cmd/cluster/transferowner.go
@@ -1,0 +1,430 @@
+package cluster
+
+import (
+	"encoding/json"
+	"fmt"
+	"regexp"
+
+	"github.com/openshift-online/ocm-cli/pkg/arguments"
+	sdk "github.com/openshift-online/ocm-sdk-go"
+	amv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
+	"github.com/openshift/osdctl/internal/utils/globalflags"
+	"github.com/openshift/osdctl/pkg/utils"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+// transferOwnerOptions defines the struct for running transferOwner command
+type transferOwnerOptions struct {
+	output            string
+	clusterID         string
+	newOwnerName      string
+	newOrganizationId string
+	oldOwnerName      string
+	oldOrganizationId string
+	dryrun            bool
+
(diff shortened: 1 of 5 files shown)


cmd/cluster/transferowner.go
```diff
@@ -0,0 +1,430 @@
...
+	if !ok {
+		return fmt.Errorf("cluster has no console url")
+	}
+
+	displayName, ok := subscription.GetDisplayName()
+	if !ok {
+		return fmt.Errorf("subscription has no displayName")
+	}
+
+	ok = validateOldOwner(o, subscription, oldOwnerAccount)
```
RaphaelBut, line 159: This is to catch wrong input. It is allowed to fail, so that if for example the subscription got patched but then the  command failed, the rerun of the command does not get blocked when checking against the already updated subscription.

cmd/cluster/transferowner.go
```diff
@@ -0,0 +1,430 @@
...
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(ops.complete(cmd, args))
+			cmdutil.CheckErr(ops.run())
+		},
+	}
+	// can we get cluster-id from some context maybe?
+	transferOwnerCmd.Flags().StringVarP(&ops.clusterID, "cluster-id", "C", "", "The Internal Cluster ID/External Cluster ID/ Cluster Name")
+	transferOwnerCmd.Flags().StringVar(&ops.newOwnerName, "new-owner", ops.newOwnerName, "The new owners username to transfer the cluster to")
+	transferOwnerCmd.Flags().StringVar(&ops.newOrganizationId, "new-organization-id", ops.newOrganizationId, "Organization of the new owner")
```
RaphaelBut, line 47: The organization ids could be taken from the account resources and are not strictly needed as input. 

The SOP says organization ids will be provided in the ticket so I think for safeguards it might be a good idea to require them.

iamkirkbater:
@RaphaelBut - since this was tested successfully today does this still need to be a draft PR?

Can we also squash these commits down to a single commit or a few meaningful milestones if there are any?

sam-nguyen7 (commented):
/lgtm
/hold
This is great. Thanks @RaphaelBut . I'ved verified that it works on prod. I can help merge this PR once squashed and no longer in draft.

RaphaelBut:
@sam-nguyen7  Ty, squashed and ready to merge.

PR #248, opened 2022-08-30
[OSD-11315] Add Cloudtrail events to cluster context command

dkeohane:
Ticket ref: https://issues.redhat.com/browse/OSD-11315

Added pulling cloudtrail events to cluster context command.
Currently just blindly pulling the last 40 pages of events, and doing minor filtering on the events after the fact, as Cloudtrail limits the number of attributes you can filter on to 1 per request. Filtering out read-only event types. 

Example use:
./osdctl cluster context {clusterID} -p rhcontrol

### DIFF
--- a/pkg/osdCloud/aws.go
+++ b/pkg/osdCloud/aws.go
@@ -12,6 +12,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/service/sts"
 
+	sdk "github.com/openshift-online/ocm-sdk-go"
 	"github.com/openshift/osdctl/pkg/provider/aws"
 	"github.com/openshift/osdctl/pkg/utils"
 )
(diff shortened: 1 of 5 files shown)


mmazur:
Notes:
1. Default run of the `context` command must not assume a user is logged into a particular cluster. People have different workflows and this would break some.
2. Therefore anything that does assume such additional env requirements should be behind a flag. (`--full` to get all the possible data?)

pkg/osdCloud/aws.go
```diff
@@ -178,3 +179,93 @@ func GenerateRoleSessionName(client aws.Client) (string, error) {
 
 	return fmt.Sprintf("RH-SRE-%s", username), nil
 }
+
+func GenerateAWSClient(awsProfile string, clusterID string) (aws.Client, error) {
```
fahlmant, line 183:
```suggestion
func GenerateAWSClientForCluster(awsProfile string, clusterID string) (aws.Client, error) {
```

pkg/osdCloud/aws.go
```diff
@@ -178,3 +179,93 @@ func GenerateRoleSessionName(client aws.Client) (string, error) {
...
+
+		awsClient, err = aws.NewAwsClientWithInput(&aws.AwsClientInput{
+			AccessKeyID:     *assumedRoleCreds.AccessKeyId,
+			SecretAccessKey: *assumedRoleCreds.SecretAccessKey,
+			SessionToken:    *assumedRoleCreds.SessionToken,
+			Region:          clusterRegion,
+		})
+		if err != nil {
+			return nil, err
+		}
```
fahlmant, line 245: I think making this into a function `BuildCCSClusterClient()` might be helpful, in case this logic ever needs to be ran on its own outside of larger function here.

fahlmant (approved):
/approve
/lgtm

PR #250, opened 2022-09-08
Added 'access' subcommand and associated unit tests

tnierman:
Completes: https://issues.redhat.com/browse/OSD-12757

---

Adds `break-glass` and `break-glass cleanup` subcommands to the `osdctl cluster` command.

- `break-glass` takes a single cluster-identifier argument, which is used to evaluate the cluster type and execute the appropriate procedure to grant the user emergency access to the cluster. Users must be logged onto the cluster's hive shard for this command to function properly. 

  For PrivateLink clusters, a temporary pod is spun up in the cluster's namespace on hive. These pods are configured to expire after 8 hours, however they can be terminated immediately using the `break-glass cleanup` subcommand.

  For non-PrivateLink clusters, the Kubeconfig is retrieved from hive, saved to the local tmp filesystem, and the user is prompted as to whether they would like to spawn a new terminal using the Kubeconfig. It is the responsibility of the SRE calling this command to delete the Kubeconfig file when finished. 

  In all cases, the user is prompted to elevate permissions if they have not impersonated the correct user using the global `--as` flag.

- `break-glass cleanup` takes a single cluster-identifier argument, which is used to evaluate the type of cluster who's access is being dropped, and execute the appropriate procedure to do so. For PrivateLink clusters, users must be logged into the cluster's hive shard for this command to function properly. For non-PrivateLink clusters, this command acts as a glorified `unset KUBECONFIG`.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -12,6 +12,7 @@ import (
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	amv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
 )
 
 type lmtSprReasonItem struct {
@@ -261,4 +262,35 @@ func ConfirmSend() error {
 		log.Fatalf("Exiting...")
 	}
 	return nil
+}
+
+// streamPrintln appends a newline then prints the given msg using the provided IOStreams
+func StreamPrintln(stream genericclioptions.IOStreams, msg string) {
+	stream.Out.Write([]byte(fmt.Sprintln(msg)))
+}
+
+// streamPrint prints the given msg using the provided IOStreams
+func StreamPrint(stream genericclioptions.IOStreams, msg string) {
+	stream.Out.Write([]byte(msg))
+}
+
+// streamPrint prints the given error msg using the provided IOStreams
+func StreamErrorln(stream genericclioptions.IOStreams, msg string) {
+	stream.ErrOut.Write([]byte(fmt.Sprintln(msg)))
+}
+
+// StreamRead retrieves input from the provided IOStreams up to (and including) the delimiter given
+func StreamRead(stream genericclioptions.IOStreams, delim byte) (string, error) {
+	reader := bufio.NewReader(stream.In)
+	return reader.ReadString(delim)
+}
+
+// Contains returns true if the given key is present in the provided list
+func Contains(list []string, key string) bool {
+	for _, item := range list {
+		if item == key {
+			return true
+		}
+	}
+	return false
 }
--- a/README.md
+++ b/README.md
@@ -285,6 +285,19 @@ osdctl federatedrole apply -u <URL>
 osdctl federatedrole apply -f <yaml file>
 ```
 
+### Cluster break-glass access
+#### Access cluster
+```bash
+# Login to the cluster's hive shard
+osdctl cluster break-glass <cluster identifier> --as backplane-cluster-admin
+```
+
+#### Drop cluster access
+```bash
+osdctl cluster break-glass cleanup <cluster identifier>
+# Non-PrivateLink - remove any Kubeconfig files saved locally in /tmp/
+```
+
 ### Send a servicelog to a cluster
 
 #### List servicelogs
--- a/cmd/cluster/access/access.go
+++ b/cmd/cluster/access/access.go
@@ -1,0 +1,451 @@
+package access
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	fpath "path/filepath"
+	"strings"
+	"time"
+
+	clustersmgmtv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/osdctl/pkg/k8s"
+	osdctlutil "github.com/openshift/osdctl/pkg/utils"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+
+	corev1 "k8s.io/api/core/v1"
+	kerr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	clientcmdapiv1 "k8s.io/client-go/tools/clientcmd/api/v1"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	kclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// impersonateUser represents the user SREs are allowed to impersonate in order to retrieve a cluster's kubeconfig.
+	impersonateUser     = "backplane-cluster-admin"
+	kubeconfigSecretKey = "kubeconfig"
+
+	// PrivateLink "jump pod" configuration
+	jumpImage         = "image-registry.openshift-image-registry.svc:5000/openshift/cli:latest"
+	jumpContainerName = "jump"
+	jumpPodLabelKey   = "automated-break-glass-access/cluster"
+
+	// Lifespan for jump pods in seconds. Currently, PrivateLink jump pods will expire after 8 hours
+	jumpPodLifespan = 28800
+)
+
+var (
+	jumpPodPollInterval = 5 * time.Second
+	jumpPodPollTimeout  = 5 * time.Minute
+)
+
+// NewCmdCluster implements the 'cluster access' subcommand
+func NewCmdAccess(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
+	accessCmd := &cobra.Command{
+		Use:               "break-glass <cluster identifier>",
+		Short:             "Emergency access to a cluster",
+		Long:              "Obtain emergency credentials to access the given cluster. You must be logged into the cluster's hive shard",
+		Args:              cobra.ExactArgs(1),
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(accessCmdComplete(cmd, args))
+			// Prior to creating k8s client, verify the user has elevated permissions
+			cmdutil.CheckErr(verifyPermissions(streams, flags))
+			client := k8s.NewClient(flags)
+			clusterAccess := newClusterAccessOptions(client, streams, flags)
+			cmdutil.CheckErr(clusterAccess.Run(cmd, args))
+		},
+	}
+	accessCmd.AddCommand(newCmdCleanup(streams, flags))
+
+	return accessCmd
+}
+
+// clusterCmdComplete verifies the command's invocation, returning an error if the usage is invalid
+func accessCmdComplete(cmd *cobra.Command, args []string) error {
+	if len(args) != 1 {
+		return cmdutil.UsageErrorf(cmd, "Exactly one cluster identifier was expected")
+	}
+	return osdctlutil.IsValidClusterKey(args[0])
+}
+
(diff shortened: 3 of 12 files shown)


cmd/access/common.go
```diff
@@ -0,0 +1,33 @@
+package access
```
dustman9000, line 1: Can you also add a `common_test.go` for future unit testing?

cmd/access/cluster.go
```diff
@@ -0,0 +1,446 @@
...
+	if err != nil {
+		return err
+	}
+	c.Println(fmt.Sprintf("Kubeconfig Secret: %s", kubeconfigSecret.Name))
+
+	if cluster.AWS().PrivateLink() {
+		c.Println("")
+		c.Println("Cluster is PrivateLink, and is only accessible via a jump pod on Hive")
+		return c.createJumpPodAccess(cluster, kubeconfigSecret)
+	} else {
```
sam-nguyen7, line 178: else not needed

cmd/access/cluster.go
```diff
@@ -0,0 +1,446 @@
...
+	if len(args) != 1 {
+		return cmdutil.UsageErrorf(cmd, "Exactly one cluster identifier was expected")
+	}
+	return osdctlutil.IsValidClusterKey(args[0])
+}
+
+// verifyPermissions determines if the user has supplied the correct permissions in order to retrieve a KubeConfig secret from hive.
+// If the user attempts to impersonate an unexpected account, an error is returned.
+// If the user hasn't attempted to impersonate anyone, it prompts whether they would like to do so automatically.
+func verifyPermissions(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) error {
```
sam-nguyen7, line 80: Instead of checking for all possible permission sets, can we just error out if backplane cluster admin is not used?

cmd/access/cluster.go
```diff
@@ -0,0 +1,446 @@
...
+	a := clusterAccessOptions{
+		IOStreams:   streams,
+		ConfigFlags: flags,
+		Client:      client,
+	}
+	return a
+}
+
+// Println appends a newline then prints the given msg using the clusterAccessOptions' IOStreams
+func (c *clusterAccessOptions) Println(msg string) {
```
sam-nguyen7, line 123: These functions looks like they should be in utils, but I'm not sure I understand why this is being used, could you explain?

cmd/access/cluster.go
```diff
@@ -0,0 +1,446 @@
...
+
+	c.Println(fmt.Sprintf("Jump pod created. Waiting for it to start"))
+	c.Println("")
+
+	err = c.waitForJumpPod(pod, jumpPodPollInterval, jumpPodPollTimeout)
+	if err != nil {
+		c.Errorln("Timed out waiting for pod to start.")
+		c.Println(fmt.Sprintf("You can check the status of the pod using\n\n    oc describe pods %s -n %s\n", pod.Name, pod.Namespace))
+		c.Println("Once the pod is running:")
+	} else {
```
sam-nguyen7, line 204: else not needed

cmd/access/cluster.go
```diff
@@ -0,0 +1,446 @@
...
+		}
+
+		c.Println(fmt.Sprintf("File has been written to '%s' for manual use", kubeconfigFilePath))
+		return fmt.Errorf("Could not parse cluster's kubeconfig Secret")
+	}
+
+	// Determine if cluster utilizes a Private API
+	listening, listeningOK := cluster.API().GetListening()
+	if !listeningOK {
+		c.Errorln("\nFailed to determine if the cluster is private.\nIf you're not able to access the cluster, try modifying the resulting kubeconfig according to the SOP: https://github.com/openshift/ops-sop/blob/master/v4/howto/break-glass-kubeadmin.md#for-clusters-with-private-api")
```
sam-nguyen7, line 241: is return needed?

cmd/access/cluster.go
```diff
@@ -0,0 +1,446 @@
...
+	} else if listening == clustersmgmtv1.ListeningMethodInternal {
```
sam-nguyen7, line 242:
```suggestion
	} 
	if listening == clustersmgmtv1.ListeningMethodInternal {
```

sam-nguyen7: if return above

sam-nguyen7 (changes requested):
Great start Trevor! I left some comments but here are the highlights that I see so far:
- Can we change the structure of these to be sub-commands of `osdctl cluster`?
- There are a lot of places where `else` is being used. In most cases, they can be removed without changing the logic and makes it easier to read/review.
- cluster.go has access to osctlutil directly, so they don't need to be methods of clusterAccessOptions object. 
- I reviewed some places with streams and println. Wherever we don't need customer input (eg. line 335-344 in cmd/access/cluster), we can condense these to one Println statement.

cmd/access/cluster.go
tnierman: Yeah, that's the current implementation of this function, with the added step of, if no impersonation user is given, the user is prompted to impersonate `backplane-cluster-admin`. 

Is the request here to remove the user prompts?

cmd/access/cluster.go
```diff
@@ -0,0 +1,446 @@
...
+// verifyPermissions determines if the user has supplied the correct permissions in order to retrieve a KubeConfig secret from hive.
+// If the user attempts to impersonate an unexpected account, an error is returned.
+// If the user hasn't attempted to impersonate anyone, it prompts whether they would like to do so automatically.
+func verifyPermissions(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) error {
+	if flags.Impersonate != nil && *flags.Impersonate != "" {
+		if *flags.Impersonate != impersonateUser {
+			osdctlutil.StreamPrintln(streams, "")
+			return fmt.Errorf("Unauthorized impersonation as user '%s'. Only requests to impersonate '%s' are allowed.", *flags.Impersonate, impersonateUser)
+		}
+	} else {
```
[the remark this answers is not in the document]
tnierman: Else is required in the case where the `flags.Impersonate` field is already set to the correct `impersonateUser` (`backplane-cluster-admin`); removing it would run the prompt logic when the correct user has already been set.

cmd/access/cluster.go
tnierman: The else is needed here to conditionally run the block

cmd/access/cluster.go
tnierman: It could be, but I've attempted to make this code as reasonably resistant to failure as possible. In general, this command is only run when things are already going poorly in a cluster or our ops environment, and usually as a last resort. 

With that in mind, I've take the stance in several places that if an error isn't completely fatal, we should just print the error, while still giving a best effort to complete the task at hand. Here, for instance, even if we don't know whether the cluster is private or not, we might still be able to pull it's kubeconfig, which is a requirement for both private & non-private clusters. If the kubeconfig pulled doesn't work, the error message indicates they may have to modify the file, per the SOP, but it at least saves them the effort of having to manually pull the file when we could still do so easily here :). 

(This is also the case when we fail to retrieve the kubeconfig file from within it's secret - we just save the secret locally, that way the SRE can try and manually pull it).

That being said, all of this is very non-obvious, and I definitely should have at least commented something indicating the intentional omission of a return statement here. I've added one now that hopefully clears up confusion in the future.

pkg/utils/utils.go
```diff
@@ -262,3 +264,34 @@ func ConfirmSend() error {
 	}
 	return nil
 }
+
+// streamPrintln appends a newline then prints the given msg using the provided IOStreams
+func StreamPrintln(stream genericclioptions.IOStreams, msg string) {
```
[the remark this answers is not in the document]
tnierman: Sorry, I may need a bit of clarification on this: is the question why use streams? Or why abstract them in these functions? (Or am I misunderstanding completely?)

If "why use streams", my perception was that the `genericclioptions.IOStreams` [passed to the rootCmd](https://github.com/openshift/osdctl/blob/master/main.go#L24) was intended to be the default I/O method for the various subcommands of osdctl. This was primarily inferred from the arguments passed to the subcommands in the [root command](https://github.com/openshift/osdctl/blob/master/cmd/cmd.go#L53-L73): with the minor exceptions of `version`, `check-upgrade`, and `compatability`, they all receive the same IOStream object.

If "why abstract them in these functions" - they're just for convenience and to keep the code DRY.

cmd/access/cluster.go
tnierman: They're just wrappers for the `Stream*` functions that utilize the `clusterAccessOptions`'s `IOStreams`.

Using `c.Println(msg)` or `c.Print(msg)` is more idiomatic than `osdctlutils.StreamPrintln(c.IOStreams, msg)` or `c.IOStreams.Out.Write([]byte{msg})` for every I/O operation

cmd/access/common.go
```diff
@@ -0,0 +1,37 @@
...
+	nsList := corev1.NamespaceList{}
+	labelSelector := metav1.LabelSelector{MatchLabels: map[string]string{hiveNSLabelKey: clusterid}}
+	selector, err := metav1.LabelSelectorAsSelector(&labelSelector)
+	if err != nil {
+		return corev1.Namespace{}, err
+	}
+	err = client.List(context.TODO(), &nsList, &kclient.ListOptions{LabelSelector: selector})
+	if err != nil {
+		return corev1.Namespace{}, err
+	} else if len(nsList.Items) != 1 {
```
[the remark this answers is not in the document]
tnierman: Personally, I think the structure of an `if-else` block makes sense here: the block's purpose is to iterate through the possible failure conditions of the `List()` operation and return an error if one is met. Breaking it up (at least in my mind), implies that there's some other condition we're checking unrelated to the `List()` that may imply failure before returning the cluster namespace.

Does anyone else have strong opinions about this?

cmd/access/cluster.go
tnierman: Yeah, since I don't think we'd necessarily want to return in the above statement, this should remain (that way this block is skipped if `listeningOk` is `false`)

cmd/access/cluster.go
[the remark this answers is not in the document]
tnierman: Same sorta thing I mention [below](https://github.com/openshift/osdctl/pull/250/files#r978965138): I think the structure of an `if-else` block makes sense in this context, since we're switching over the possible cluster access methods based on the properties of the clusters.

Does anyone else have thoughts on this?

cmd/access/cluster.go
```diff
@@ -0,0 +1,446 @@
...
+	jumpPodLifespan = 28800
+)
+
+var (
+	jumpPodPollInterval = 5 * time.Second
+	jumpPodPollTimeout  = 5 * time.Minute
+)
+
+// newCmdCluster implements the 'cluster' access subcommand
+func newCmdCluster(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
```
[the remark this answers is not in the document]
tnierman: Honestly, I think the invocation `osdctl access cluster my-cluster` and `osdctl access drop my-cluster` are more intuitive (and they conform to cobra's `APPNAME VERB NOUN` [recommendation](https://pkg.go.dev/github.com/spf13/cobra#readme-concepts)), but the current convention in osdctl does seem to be `APPNAME NOUN VERB`. Additionally, the `cluster` subcommand already has several cluster-specific subcommands under it.

I've reorganized the structure so the the logic in this PR can be invoked with `osdctl cluster break-glass <cluster identifier>` and `osdctl cluster break-glass cleanup <cluster identifier>`

cmd/access/common.go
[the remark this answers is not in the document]
dustman9000: Typically the standard has been to omit the `else` if the previous `if` block has a return statement. This is a nuance with golang.

dustman9000:
@tnierman - Looks good! Do you mind updating the PR description?

cmd/access/common.go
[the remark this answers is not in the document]
tnierman: Welp, it appears I'm outvoted. I've updated the code to reflect this and the suggestion above

cmd/access/cluster.go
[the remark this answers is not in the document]
tnierman: Based on the conversation below, it appears I've been outvoted on this - I've updated the logic to reflect these changes :+1:

PR #256, opened 2022-09-23
Add resize-control-plane-node command

typeid:
[OSD-8560](https://issues.redhat.com//browse/OSD-8560)
https://github.com/openshift/ops-sop/blob/master/v4/howto/resize-control-plane.md

PR introduces a new command. `osdctl cluster resize-master-node` which takes in an internal `clusterID`, a `node` and an AWS `machine_type` to resize a node's instance.

Example usage:
```
osdctl cluster resize-control-plane-node -C <internal_clusterID> --machine-type m5.xlarge --node ip-10-0-123-123.ec2.internal
```

Resizing process:
1. Retrieve AWS credentials via backplane-api
2. Initialize AWS-cli
3. Drain target node - if this fails, we get the options to skip/force/cancel
4. stop node's instance (this is a blocking call until it's stopped)
5. modify node's instance attribute (size of the node, e.g. m5.xlarge)
6. start node (this is a blocking call until it's started)
7. uncordon node
8. update machine spec


cmd/cluster/resizenode.go
```diff
@@ -0,0 +1,342 @@
...
+func (o *resizeNodeOptions) initAwsCli() *awsprovider.Client {
+	fmt.Println("Attempting to initialize AWS Client. Switching to hive context to get credentials.")
+
+	targetCluster := o.k8sclusterresourcefactory.ClusterID
+	hiveCluster, err := utils.GetHiveCluster(targetCluster)
+	if err != nil {
+		log.Fatalln("Unable to get hive cluster:", err)
+	}
+
+	utils.SwapOCMContext(hiveCluster)
```
Nikokolas3270, line 106: Shouldn't you capture the error here with `err =`?

cmd/cluster/resizenode.go
```diff
@@ -0,0 +1,342 @@
...
+			log.Fatalln(err)
+		}
+	}()
+
+	/*
+		AWS Credentials to the specific cluster are fetched from the hive cluster,
+		therfore we always have to be logged in on the corresponding hive shard initially
+		to initialize the AWS Cli.
+	*/
+	awsClient, err := o.k8sclusterresourcefactory.GetCloudProvider(o.verbose)
```
Nikokolas3270, line 123: As we can see, we need 4 parameter to create an AWS client:
https://github.com/openshift/osdctl/blob/ed23ebad95488c347ab951d3f6f33a47ea466c26/pkg/k8s/clusterresourcefactory.go#L248
- An access key
- A secret access key
- A session token
- The AWS region to target

Playing with command `ocm backplane cloud credentials`, I can see that this command exactly provides this info.
Looking the code for that command I can see that `backplane-api` service is queried for that:
https://gitlab.cee.redhat.com/service/backplane-cli/-/blob/91a2fa5f29f4506a2b07956ae839262f441b3a21/cmd/ocm-backplane/cloud/credentials.go#L149

As you can see, the only parameter is the cluster id (which is an info you have), verb which is used is the `GetCloudCredentials` verb.
Client however is not the OCM client, it is the backplane client which is not yet used in `osdctl`. However creating that client is not hard. All you have to do is:
1. Retrieve backplane URL using OCM client:
https://gitlab.cee.redhat.com/service/backplane-cli/-/blob/91a2fa5f29f4506a2b07956ae839262f441b3a21/cmd/ocm-backplane/cloud/credentials.go#L127
2. Retrieve the OCM access token:
https://gitlab.cee.redhat.com/service/backplane-cli/-/blob/91a2fa5f29f4506a2b07956ae839262f441b3a21/cmd/ocm-backplane/cloud/credentials.go#L136
3. Then create the backplane client:
https://gitlab.cee.redhat.com/service/backplane-cli/-/blob/91a2fa5f29f4506a2b07956ae839262f441b3a21/cmd/ocm-backplane/cloud/credentials.go#L142
4. And finally make the call to retrieve the AWS credentials with backplane client:
https://gitlab.cee.redhat.com/service/backplane-cli/-/blob/91a2fa5f29f4506a2b07956ae839262f441b3a21/cmd/ocm-backplane/cloud/credentials.go#L149

Remark that all first 3 operations require the `utils` package which can be added with the following import:
`gitlab.cee.redhat.com/service/backplane-cli/pkg/utils`
Once added, run `go mod tidy` to update `go.mod` and `go.sum` files.

If you do not want to add this package, you may want to copy some part of the code from this file locally:
https://gitlab.cee.redhat.com/service/backplane-cli/-/blob/91a2fa5f29f4506a2b07956ae839262f441b3a21/pkg/utils/ocmWrapper.go

Nikokolas3270: Forgot to say, but it is normally not required to have `ocm backplane tunnel` up and running to have above code running successfully.

Nikokolas3270: It is however required to be OCM logged (i.e. `ocm login` must have been successful) with the proposed change. That's hardly an issue as many commands have this prerequisite:
https://github.com/openshift/osdctl/search?q=CreateConnection

cmd/cluster/resizenode.go
```diff
@@ -0,0 +1,342 @@
...
+func newCmdResizeNode(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags, globalOpts *globalflags.GlobalOptions) *cobra.Command {
+	ops := newResizeNodeOptions(streams, flags, globalOpts)
+	resizeNodeCmd := &cobra.Command{
+		Use:               "resizeNode",
+		Short:             "Resize a node. Requires previous login to the target cluster via ocm backplane login <target_cluster>",
+		Args:              cobra.NoArgs,
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(ops.complete(cmd, args))
+			cmdutil.CheckErr(ops.run())
```
Nikokolas3270, line 44: It is a bit strange to me to have a `run` method on options object.
Semantically speaking, it would have preferred having some `run` function taking options as parameter

cmd/cluster/resizenode.go
```diff
@@ -0,0 +1,342 @@
...
+	default:
+		fmt.Println("Invalid response, expected 'skip', 'force' or 'cancel' (case-insensitive).")
+		return drainRecoveryDialog()
+	}
+}
+
+func drainNode(nodeID string) {
+	fmt.Println("Draining node", nodeID)
+
+	// TODO: replace subprocess call with API call
```
Nikokolas3270, line 169: Big +1 :)

cmd/cluster/resizenode.go
typeid: As already discussed, breaking the user's oc/kubectl context - even if just momentarily - is also my biggest issue with this PR. 
Unfortunately the ocm-backplane repo, where all of this is already implemented in (as you stated, e.g. used in ocm-backplane cloud console), is in our private gitlab. I wonder what it would take to open source it?

I know that @fahlmant is working on a major refactor of all the outside API we're using in this repository, which also AFAIK includes making it trivial to initialize an aws-cli without a context switch. @fahlmant do you maybe have some input on this PR?

I see the following steps we could take to move forward from here:
- wait for the refactor to happen, rebase this PR with proper aws-cli creation without context swap
- implement the aws-cli creation without context swap in this PR
- keep the context switch for now, fix when the refactor is done

cmd/cluster/resizenode.go
typeid: Semantically I tend to agree, however this seems to be a standard in all our cobra cli projects.
See: [oc cli](https://github.com/openshift/oc/blob/c443e1c0b03ee88b44f237d08572346332c9bbe5/pkg/cli/login/login.go#L51) and other commands in `osdctl`

cmd/cluster/resizenode.go
typeid: Good catch, thank you :)

cmd/cluster/resizenode.go
typeid: As an update to the backplane-api: I was able to quickly implement this PR using backplane-api, this gets rid of the oc context swaps. 
However, the API calls require you to be (as far as I can tell) backplane tunneled. This breaks the whole concept of this tool being open source, as the command would only be usable by tunneling via a closed source tool.

typeid: Pushed the backplane-api refactor.

pkg/utils/aws.go
```diff
@@ -0,0 +1,85 @@
...
+	SecretAccessKey string `json:"SecretAccessKey" yaml:"SecretAccessKey"`
+	SessionToken    string `json:"SessionToken" yaml:"SessionToken"`
+	Region          string `json:"Region" yaml:"Region"`
+	Expiration      string `json:"Expiration" yaml:"Expiration"`
+}
+
+// Creates an AWS client based on a clusterid
+// Requires previous log on to the correct api server via ocm login
+// and tunneling to the backplane
+func CreateAWSClient(clusterID string) (aws.Client, error) {
```
fahlmant, line 37: Given this relies on getting AWS info from a cluster, I think the function name should change to reflect that. There's already an AWS package in osdctl that provides creating a raw AWS client `awsprovider.NewAwsClientWithInput` so to have two functions with similar names and no context may get confusing. At a larger level, moving this util package to something named maybe `clusterAWS` or `clusterCloud` or something along those lines may help clarify.

pkg/utils/ocm.go
```diff
@@ -70,3 +70,84 @@ func CreateConnection() *sdk.Connection {
 	}
 	return connection
 }
+
+//This command implements the ocm describe cluster call via osm-sdk.
```
fahlmant, line 74:
```suggestion
//This command implements the ocm describe cluster call via ocm-sdk.
```

pkg/utils/ocm.go
```diff
@@ -70,3 +70,84 @@ func CreateConnection() *sdk.Connection {
...
+//This call requires the ocm API Token https://cloud.redhat.com/openshift/token be available in the OCM_TOKEN env variable.
+//Example: export OCM_TOKEN=$(jq -r .refresh_token ~/.ocm.json)
+func DescribeCluster(clusterID string) (*v1.Cluster, error) {
```
fahlmant, line 77: To be pedantic, this is the equivalent of `ocm get cluster $clusterID`, rather than `ocm describe cluster $clusterID`. Describe is a GET with some filtering of the data, but this code just does the GET.

pkg/utils/utils.go
```diff
@@ -231,23 +230,25 @@ func GetAccount(connection *sdk.Connection, key string) (account *amv1.Account,
 	return
 }
 
-func ConfirmSend() error {
+func ConfirmSend() bool {
```
fahlmant, line 233: I believe returning an error is better here. Everywhere this is being used, the `false` case is handled by printing and error and exiting anyway. Generally in Go the best practice is to bubble up the error to the calling logic and allowing the caller to log it and then make a decision based on it.

fahlmant (changes requested):
Generally, the logic looks good.  There are a few small things I'm requesting changes on. 

One larger request I have: would you mind changing `master` to `controlPlane`?

cmd/cluster/resizenode.go
iamkirkbater: Lots of the commands in this repo require tunneling, but that's also mostly because that's the only way to access our clusters. If someone had a similar setup and was using our tool for it depending on their networking setup they might not need the tunnel.

While we might have this tool that's open source but also purpose built, that doesn't prevent others from seeing our code and learning from it or borrowing ideas from it, so there's still benefits there.

I don't have any other context around why our other toolset is closed source, but it's probably more due to the sensitive nature of the commands there.

cmd/cluster/transferowner.go
```diff
@@ -32,7 +32,7 @@ type transferOwnerOptions struct {
 func newCmdTransferOwner(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags, globalOpts *globalflags.GlobalOptions) *cobra.Command {
 	ops := newTransferOwnerOptions(streams, flags, globalOpts)
 	transferOwnerCmd := &cobra.Command{
-		Use:               "transferOwner",
+		Use:               "transfer-owner",
```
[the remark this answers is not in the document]
typeid: Done.

cmd/cluster/loggingcheck.go
```diff
@@ -30,7 +30,7 @@ type loggingCheckOptions struct {
 func newCmdLoggingCheck(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags, globalOpts *globalflags.GlobalOptions) *cobra.Command {
 	ops := newloggingCheckOptions(streams, flags, globalOpts)
 	loggingCheckCmd := &cobra.Command{
-		Use:               "loggingCheck",
+		Use:               "logging-check",
```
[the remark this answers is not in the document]
typeid: Done.

pkg/utils/aws.go
typeid: Makes sense, wasn't sure where to put this. 
Moved to a new clustercloud package.

pkg/utils/ocm.go
typeid: Done.

pkg/utils/ocm.go
typeid: Removed this function, as I noticed it was already implemented elsewhere.

pkg/utils/utils.go
typeid: In this case, with this dialog I tend to agree, as we exit if we don't `Continue y/N?`. Changed.

cmd/cluster/resizecontrolplanenode.go
```diff
@@ -0,0 +1,301 @@
...
+	err = utils.ConfirmSend()
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("To finish the node resize, it is suggested to update the machine spec. This requires ***elevated privileges***. Do you want to proceed?")
+	err = utils.ConfirmSend()
+	if err != nil {
+		fmt.Println("Node resized, machine type not patched. Exiting...")
+		return nil
```
fahlmant, line 292: Why is the error not returned here?

fahlmant (commented):
I'm a bit wary of using log.Fatal in functions that aren't part of the control logic. Generally its best practice in go to bubble up the error from functions that do one thing and let the caller handle them instead.

pkg/utils/ocm.go
```diff
@@ -70,3 +70,62 @@ func CreateConnection() *sdk.Connection {
...
+		ProvisionShard().
+		Get().
+		Send()
+
+	var shard string
+
+	if shardPath != nil && err == nil {
+		shard = shardPath.Body().HiveConfig().Server()
+	}
+
```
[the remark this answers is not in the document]
typeid: Agreed, with the current logic we would still get an error from the default initialization of `shard` to an empty string, but we would overwrite the error. Changed, good catch! :)

cmd/cluster/resizecontrolplanenode.go
typeid: Hmm. I believe this is a leftover from when `ConfirmSend` returned the boolean and I forgot to change it to return the error. 
Either way, I think it should have been an error before as well, as we don't finalize the resize as it should be. Changed.

fahlmant (commented):
Overall, it looks good. The only concern I still have the the log.Fatal calls. cc @rafael-azevedo for thoughts on this

typeid:
> The only concern I still have the the log.Fatal calls.
Removed the log.Fatals, changed to error bubbling and merged with the osdCloud changes.

fahlmant (approved):
/lgtm
/approve

PR #257, opened 2022-09-27
Remove duplicate code and fix cluster commands

typeid:
Refactored the following:
- removed the `sdk.NewConnectionBuilder` calls, which break the commands for stage, as this call uses the hardcoded production url `api.openshift.com`. Replaced those calls with the already existing `utils.CreateConnection`
- refactored `context.go` to use the same `utils.GetLimitedSupportReasons` as `status.go`
- fixed some formatting

### DIFF
--- a/cmd/cluster/context.go
+++ b/cmd/cluster/context.go
@@ -4,14 +4,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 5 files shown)


fahlmant (approved):
/approve
/lgtm

PR #258, opened 2022-09-27
Fix AWS console generation bug where role was incorrectly double assumed

fahlmant:
There is a bug in the console generation for AWS accounts where OrganizationAccountAccessRole was assumed once by the inital user, then again by itself (OrganizationAccountAccessRole assumed OrganizationAccountAccessRole). This has silently worked due to AWS allowing implicit trust of a policy assuming itself. However, this behavior is changing and this no longer works with newer accounts. See https://aws.amazon.com/blogs/security/announcing-an-update-to-iam-role-trust-policy-behavior/ for more details. 

This is a stop gap to get this command to work, so please excuse the messiness of the code. All of k8sclusterresourcefactory will be refactored soon, so this will be made much cleaner in an upcoming PR.

### DIFF
--- a/cmd/account/console.go
+++ b/cmd/account/console.go
@@ -3,14 +3,19 @@ package account
 import (
 	"fmt"
 	"net/url"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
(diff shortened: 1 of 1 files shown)


sam-nguyen7 (approved):
/lgtm

PR #259, opened 2022-10-05
Remove k8sclusterresourcefactory from account console and cli

fahlmant:
Removes k8sclusterresourcefactory usage from `osdctl account cli` and `osdctl account console`

With this change, both `-i` and `-C` are usable without requiring hive access. Only proper OCM and AWS credentials are required. To achieve this, some new utility functions are created for common SRE workflows. Specifically 
- Assuming the OrganizationAccountAccessRole
- Performing the Assume Role chain from IAM User -> RH Support Role -> Jump Role -> ManagedOpenShift-Support target role
- Getting a cluster's AWS Account ID from OCM
- Getting a cluster's Support Role ARN for OCM
- Generating the correct session name using an SRE's kerberos ID

This is the first step in a larger refactor to fully remove k8sclusterresourcefactory, and other commands that require logging into hive (`osdctl cluster health` for example) will be able to use these functions instead.

I'm convinced this pattern is the way forward for osdctl, and we should continue to support -C separate from the backplane path as long as it remains achievable through OCM. There's a host of functionality in osdctl it wouldn't make sense to migrate, and needs the same functionality as -C anyway.

The only thing I'm not confident about is how to set the Jump Role Account ID. In the code I'm simply reading from an env var for simplicity. However, this is error prone and potentially requires setting every time, as the Account ID is different between stage and prod. I'm against hard coding as 1) They're static IDs and I'm concerned about security and 2) osdctl should remain as flexible as possible and rely on account layout, not specific accounts. I think the best way forward here may be to create an `~/.osdctl.conf` file where one can set various configs, like `stage_jump_role_account_id` and `prod_jump_role_account_id`. Then in code, the logic could determine which OCM env is being used and select the right thing from the file. This has the benefit of remaining agnostic, and is a one time setting rather than an env var that needs to be set every time and correctly by an SRE (its just too much cognitive load). The downside is everyone using this tool will need to do an extra step, but given its a one time step I think this could be acceptable. Thoughts?

### DIFF
--- a/cmd/account/cli.go
+++ b/cmd/account/cli.go
@@ -3,19 +3,17 @@ package account
 import (
 	"fmt"
 
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/openshift/osdctl/internal/utils/globalflags"
-	k8spkg "github.com/openshift/osdctl/pkg/k8s"
-	awsprovider "github.com/openshift/osdctl/pkg/provider/aws"
+	"github.com/aws/aws-sdk-go/service/sts"
+	osdCloud "github.com/openshift/osdctl/pkg/osdCloud"
+	"github.com/openshift/osdctl/pkg/provider/aws"
+	"github.com/openshift/osdctl/pkg/utils"
 	"github.com/spf13/cobra"
-	"k8s.io/cli-runtime/pkg/genericclioptions"
-	"k8s.io/klog/v2"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 )
 
 // newCmdCli implements the Cli command which generates temporary STS cli credentials for the specified account cr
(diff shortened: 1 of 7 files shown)


fahlmant:
- [ ] Stage non-CCS, console and cli
- [ ] Stage CCS, console and cli
- [ ] Prod non-CCS, console and cli
- [ ] Prod CCS, console and cli
- [ ] Dev account (-i), console and cli

clcollins:
Regarding:

> ~/.osdctl.conf

I like the idea, but would like to push for the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) usage `~/.config/osdctl/` for config files.

mjlshen:
/lgtm

Validated GovCloud usage `osdctl account cli -i 340574339851 -p fedramp -r us-gov-east-1 -oenv` and `osdctl account console -i 340574339851 -p fedramp -r us-gov-east-1`

PR #262, opened 2022-10-06
[OSD-12925] feat(support): Limited Support to allow parameters

Tessg22:
This change enables `osdctl cluster support` command to accept the parameterized input. Doc included.

### DIFF
--- a/cmd/cluster/support/post.go
+++ b/cmd/cluster/support/post.go
@@ -104,7 +104,7 @@ func (o *postOptions) run() error {
 
 	// For every '-p' flag, replace it's related placeholder in the template
 	for k := range userParameterNames {
-		replaceFlags(userParameterNames[k], userParameterValues[k])
+		replaceWithFlags(userParameterNames[k], userParameterValues[k])
 	}
 
 	//if the cluster key is on the right format
@@ -301,7 +301,7 @@ func parseUserParameters() {
 	}
 }
 
-func replaceFlags(flagName string, flagValue string) {
+func replaceWithFlags(flagName string, flagValue string) {
 	if flagValue == "" {
 		log.Fatalf("The selected template is using '%[1]s' parameter, but '%[1]s' flag was not set. Use '-p %[1]s=\"FOOBAR\"' to fix this.", flagName)
 	}
--- a/cmd/cluster/support/post.go
+++ b/cmd/cluster/support/post.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strings"
 
 	"github.com/openshift-online/ocm-cli/pkg/arguments"
 	"github.com/openshift-online/ocm-cli/pkg/dump"
@@ -23,9 +24,10 @@ import (
 )
 
 var (
-	LimitedSupport support.LimitedSupport
-	template       string
-	isDryRun       bool
+	LimitedSupport                                          support.LimitedSupport
+	template                                                string
+	isDryRun                                                bool
+	templateParams, userParameterNames, userParameterValues []string
 )
 
 const (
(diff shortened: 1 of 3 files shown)


cmd/cluster/support/post.go
```diff
@@ -89,13 +92,21 @@ func (o *postOptions) run() error {
...
 	// Check that the cluster key (name, identifier or external identifier) given by the user
 	// is reasonably safe so that there is no risk of SQL injection
 	err := ctlutil.IsValidClusterKey(o.clusterID)
 	if err != nil {
 		return err
 	}
 
+	// For every '-p' flag, replace it's related placeholder in the template
+	for k := range userParameterNames {
+		replaceFlags(userParameterNames[k], userParameterValues[k])
```
typeid, line 107: To me, `replaceFlags` doesn't describe the function well enough, how about:

**suggestion:** Rename this function to `insertFlags`

docs/command/osdctl_cluster_support.md
```diff
@@ -0,0 +1,39 @@
+## osdctl cluster support
```
typeid, line 1: **Question:** do you know how this is generated? I would suppose by `make`, but for some newer commands, this isn't auto generated.

typeid:
All threads opened are non blocking.
/lgtm

cmd/cluster/support/post.go
Tessg22: The reason behind the wording `replaceFlags` is that the template contains the variables such as `${CONFIGURATION}` and that is _replaced_ by the input given in the command. When you drill down, it uses `strings.ReplaceAll`.

docs/command/osdctl_cluster_support.md
Tessg22: It is generated with `make build` as per [osdctl#installation](https://github.com/openshift/osdctl#installation).

cmd/cluster/support/post.go
typeid: Would `replaceWithFlags` fit even better then? Semantically, `replaceFlags` would mean that we replace the flags, not that we replace the variables with flags (which is why I initially suggested `insertFlags`)

docs/command/osdctl_cluster_support.md
typeid: Thank you, I suppose something is broken with the new commands then, as the doc is not generated for them. I will look into it later if I have time.

cmd/cluster/support/post.go
Tessg22: `replaceWithFlags` sounds good.

docs/command/osdctl_cluster_support.md
Tessg22: My bad, I didn't notice that this is a comment for a doc file change.

cmd/cluster/support/post.go
```diff
@@ -23,9 +24,10 @@ import (
 )
 
 var (
```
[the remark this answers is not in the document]
Tessg22: The reason for defining the LimitedSupport variable isn't unit testing but saving a template file that has been provided through the command into a JSON struct.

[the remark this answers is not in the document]
typeid: It's indeed defined to parse the template file to a struct, but it wouldn't have to be a global variable if it wasn't used in the unit tests (`post_test.go`: e.g L68, L74).

If the above was not the case, we could have defined all the global variables:
```
var (LimitedSupport                                          support.LimitedSupport
	template                                                string
	isDryRun                                                bool
	templateParams, userParameterNames, userParameterValues []string
)
```
within the `postOptions` struct.

The benefits would be:
- all variables in one place
- the json marshalling could be done before running any other logic (preemptive validation)
- no polluting of the global `support` namespace

This is out of scope for this PR though, and would require reworking the way we unit tests this. So just a thought :)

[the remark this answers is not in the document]
Tessg22: I understand now. Do you want to create an OSD card for it that can be picked up during the maintenance sprint?

Tessg22:
@typeid could you please add also approve label?

clcollins:
It looks like the SOP for Limited Support beat this change into osdctl.  We should be more careful to hold those PRs to OPS-SOP until the code it references is in place.

I was one of the approvers, so this is on me too, but we need to share this warning the rest of the team as well.

PR #267, opened 2022-10-11
Update osdctl cluster resize-control-plane command to better reflect SOP changes

typeid:
**Why?** 
The resize-control-plane has many steps which are very failure prone because they are often executed when the API server is already under heavy load. To add robustness to the command, we need to be able to have options after every failure.

**What?**
- Added option to retry in case of node drain failure
- Added option to retry in case of force drain failure
- Added option to retry in case of node stop failure
- Added option to retry in case of node instance type change failure
- Added option to retry in case of node start failure
- Added option to retry or skip in case of uncordon failure
- Added option to retry in case of patch machine spec failure
- Added more error logging to better reflect the state of the process to the user.

- Fixed command outputs to enable the user to know what exactly failed, and how to proceed according to the SOP

- Updated output of single steps to be easier to read


rafael-azevedo:
/lgtm 

I really like these additional options !

typeid:
Noticed a small but unhealthy error in the return logic, fixed with the last commit. 

PR should be good to be merged to already improve usability. However, I am working on actually making the dialogs consistant and enable retry/skip options for everything we have after the feedback from @sam-nguyen7 that the API server fails quite a bit, causing errors with the automated resize.

sam-nguyen7 (commented):
/lgtm
/approve

PR #270, opened 2022-10-13
Create osdctl config file

fahlmant:
This PR creates a new config file for `osdctl` to use at `~/.config/osdctl`. It's a key-value format, with the key and value separated by a colon, and each key-value pair on a new line.

This PR has a few other small changes:
- Renames the directory the `osdctl env` code lives from config to envConfig. The package name is the same
- Creates a function that returns the current OCM Environment (prod, stage, int) for a given connection
- Refactors generating credentials for CCS clusters to get the Jumprole account IDs from the new config file


iamkirkbater:
Is there any specific reason to roll our own config file here instead of using something like `viper` which can abstract a lot of this for us and would be a bit more standard between codebases? Another benefit of using a package like `viper` is that it's relatively simple to allow default behavior -> user configs -> env vars in order of precedence so that you can override things during individual command runs, and you get that out of the box as well.

pkg/osdCloud/aws.go
```diff
@@ -125,9 +127,19 @@ func GenerateJumpRoleCredentials(client aws.Client, awsAccountID, region, sessio
 		return nil, err
 	}
 
+	jumpRoleKey := ProdJumproleConfigKey
```
iamkirkbater, line 130: Can we also add a message here noting that using `osdctl` to assume support role credentials will be deprecated in a future version in favor of backplane?

I don't want to obliterate anyone's workflow, but at the same time we should probably start the deprecation process sooner rather than later.

iamkirkbater: (It doesn't have to be _here_ specifically, but inside this function somewhere, or somewhere it makes sense that we're _not_ using the `-i` flag.)

fahlmant: In https://github.com/openshift/osdctl/pull/259 I made changes to the workflow for access AWS accounts for -C that no longer require hive access, so I don't think we should deprecate -C. I've refactored now a number of commands, like osdctl health, to use this flow as it needs the same functionality as -C anyway.

fahlmant:
As for Viper, I wasn't aware of it before. After looking at it, I'm not sure its the best choice as of now, simply because using JSON, YAML, TOML, etc, I think is too complex for what we currently are using this config file for, a key-value store. Even simple changes to data in these formats tend to become "find the formatting issue" rather than "make sure the data is correct", so I wanted to stay away from them unless absolutely needed. So far, the use cases of the config file I've heard are confined to AWS Account IDs and profiles, and a PD api key, and a key value pair with a colon and a new line work for these. Its entirely possible there are use cases that I've missed, and if JSON, YAML or whatever make more sense given those use cases, then I'm happy to refactor this.

pkg/osdctlConfig/osdctlConfig.go
```diff
@@ -0,0 +1,85 @@
...
+
+func GetValueForKey(key string) (string, error) {
+
+	configFileMap, err := generateMapFromConfig()
+	if err != nil {
+		return "", err
+	}
+
+	if value, ok := configFileMap[key]; ok {
+		return strings.TrimSpace(value), nil
```
dkeohane, line 49: We should probably move the trimming of key and value into the generateMapFromConfig after the split, to avoid people accidentally setting an invalid kvs format and messing up

fahlmant: I agree, it's probably best to contain all parsing logic in one place. Changing

pkg/osdCloud/aws.go
iamkirkbater: Hmm, okay.  I thought the reason for the deprecation was mostly so that we only have the backplane access available, but it does make sense to have a backup method in case backplane itself goes down.

iamkirkbater:
Re: Viper

I know that osde2e uses it heavily, and I'm using it in the v2 rewrite of [ocm-container](/iamkirkbater/ocm-container-v2) (which is an all-go rewrite).  I've found it to be extremely simple to add and really like the flexibility (specifically that around the precedence to how flags can be overwritten).  

My _only_ real concern here about using "the simple key-value store" is that if we _do_ need more functionality later that a structured markup language would provide we're more prone to try to get it working within the current system rather than making the migration, leading to an even more complex config parsing code than necessary, and then getting users to migrate from one config file to the other would be a pain unless we build some kind of automatic migration process into the codebase itself, which again is just more bespoke code for us to maintain.

I'll leave the decision up to you, though, your point about the simplicity does make sense overall.

If we do keep the custom config file, can we add some tests with a few use cases so that if we _do_ end up adding more complex validation logic we can at least prevent regressions?

mrbarge:
I'd be another +1 for viper (backplane and ocm-agent also use it) if only because given osdctl is already using cobra for the CLI and viper is a natural companion to it - allowing for things like extremely straightforward binding between CLI parameters to viper config (which is practically a 1-liner). But I'm equally fine the approach already already taken.

pkg/osdCloud/aws.go
fahlmant: Part of the deprecation discussion (at least from my perspective) was the hive access requirement was annoying and backplane was much simpler. Now the osdctl method is also simple and is only dependent on aws credentials and ocm credentials, so I changed my stance on it. For now, I would prefer to keep it just for backwards compatibility and ease of use.

fahlmant:
/hold
I'll do a spike of viper today. I'm all for using the right tools for the job, and I agree that changing the config file will be annoying once established, so I'll give it a fair try.

fahlmant:
I refactored to include Viper and specified YAML as the format, since that seems to be the best compromise between human manageable and easily parseable, and works with the `key: value` format out of the box.

PR #273, opened 2022-10-17
Add kraken to cluster context and trim down output

mrbarge:
This is a minor update to the `cluster context` command that:

- Adds a link to the CCX dashboard page for a cluster
- Groups all the 'external links' entries into a single section to cut down on terminal spammy-ness


clcollins:
/lgtm

:heart: Kraken!

PR #274, opened 2022-10-17
Fix: documentation generation

typeid:
**Why?:** `make docs` is broken because there is also a folder called `docs`. See [here](https://stackoverflow.com/questions/3931741/why-does-make-think-the-target-is-up-to-date). The other issue is that it calls osdctl only with the ouput folder and misses the actual command `docs`.

**What?:** Fix documentation make command and add generated documentation.

---

I also have trouble understanding why we use `@git diff --exit-code` everywhere in the makefile. Do we really want the make commands to fail whenever something is generated that isn't cached/commited yet? 

We should also find a way to have this generated every time someone re-builds, we could just add this to the build step, but it's a bit odd to generate the doc of the tool using the tool itself. 
We also might want to delete docs and re-create it every time. This way, we get rid of old docs that don't exist anymore.


fahlmant (approved):
/approve

PR #275, opened 2022-10-19
[OSD-12879] Fix assume role for rotate and generate secret

wshearn:
@fahlmant @iamkirkbater

### DIFF
--- a/cmd/account/generate-secret.go
+++ b/cmd/account/generate-secret.go
@@ -269,9 +269,29 @@ func (o *generateSecretOptions) generateCcsSecret() error {
 		return err
 	}
 
+	// Get the Jump ARN value
+	JumpARN := cm.Data["support-jump-role"]
+	if JumpARN == "" {
+		return fmt.Errorf("Jump Access ARN is missing from configmap")
+	}
+	// Assume the ARN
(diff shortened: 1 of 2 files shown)


afreiberger:
Minor nit, the aws.Int64(900) variable in the calls is hard to understand w/out referencing documentation outside this codebase.  Any way we might be able to define an "aws token timeout" variable that equals aws.Int64(900) and reference that in the function calls to enhance readability and ensure that future update to one 900 second timeout addresses all token timeouts?

This was a pre-existing pattern, so I realize I'm asking for an enhancement during this functionality add.

wshearn:
I think something like that would be better as an enhancement issue on here or jira. As you pointed out it is a preexisting pattern, so a card to make sure it is common across the entire codebase would be better.

wshearn:
/hold

@fahlmant Looking at their request, it seems that aws.Int for this timeout is really only used in these two files, and will be low effort to change, the thing is the only other place it is used once is with aws.Int(0) for the timeout so not a full change everywhere which makes me not want to put it in the const file.
I am thinking maybe put it in the options struct of each file?

fahlmant:
+1 to making that change

clcollins:
@wshearn is this ready to go?  Anything related to @fahlmant's request holding this up?

PR #276, opened 2022-10-20
Port cluster provisioning delay/failure script to osdctl

fahlmant:
Ports https://github.com/openshift/ops-sop/blob/master/v4/utils/cluster_provisioning_failure to osdctl. The logic also fixes a bug in the script where subnets that don't have an explicitly associated routetable cause an error incorrectly. When a subnet isn't associated with a routetable, AWS defaults to using the Main routetable for a VPC. The logic here will fallback to using the Main routetable for the VPC when a subnet doesn't have an explicit routetable association.

### DIFF
--- a/cmd/cluster/cpd.go
+++ b/cmd/cluster/cpd.go
@@ -1,0 +1,188 @@
+package cluster
+
+import (
+	"fmt"
+
+	awsSdk "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/openshift/osdctl/pkg/osdCloud"
+	"github.com/openshift/osdctl/pkg/provider/aws"
+	"github.com/openshift/osdctl/pkg/utils"
+	"github.com/spf13/cobra"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+type cpdOptions struct {
+	clusterID  string
+	awsProfile string
+}
+
+const (
+	unknownProvisionCode = "OCM3999"
+)
+
+func newCmdCpd() *cobra.Command {
+	ops := cpdOptions{}
+	cpdCmd := &cobra.Command{
+		Use:               "cpd",
+		Short:             "Runs diagnostic for a Cluster Provisioning Delay (CPD)",
+		Args:              cobra.NoArgs,
+		DisableAutoGenTag: true,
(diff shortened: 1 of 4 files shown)


cmd/cluster/cpd.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// Check if the OCM Error code is a known error
+	if clusterInfo.Status().ProvisionErrorCode() != unknownProvisionCode {
+		fmt.Printf("Error code %s is known, customer already received Service Log\n", clusterInfo.Status().ProvisionErrorCode())
+	}
+
+	// If the cluster is GCP, give instructions on how to get console access
+	if clusterInfo.CloudProvider().ID() == "gcp" {
+
+		return nil
```
clcollins, line 68: Should there be a "Not Implemented Yet" message here, at least?

fahlmant: Ah yes, I forgot to do that. Will add

cmd/cluster/cpd.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	// Get AWS credentials for the cluster
+	awsClient, err := osdCloud.GenerateAWSClientForCluster(o.awsProfile, o.clusterID)
+	if err != nil {
+		fmt.Println(err)
```
mjlshen, line 74: This may be dangerous, but assuming credentials are setup correctly by the SRE we could point them to SL: https://github.com/openshift/managed-notifications/blob/master/osd/aws/ROSA_AWS_invalid_permissions.json

However if credentials aren't setup correctly by the SRE they may mistakenly send this

clcollins: We could tell them to verify their credentials are set first, and give them the link.

fahlmant: I can try to add a bit more error handling around this. FWIW, this is what is in the current script. The error from AWS should hopefully be self explanatory but I'll try to make it as clear as possible here.

mjlshen: Yeah, sorry my comment was a little misleading - this is more of a potential feature request.

If SREs have valid AWS creds locally, but can't assume role because of some broken AWS IAM on the customer's part, that SL above is the one to send. I don't think the existing script does this either, so totally fine if it's an improvement for later.

The dangerous part is maybe SREs have invalid creds locally and they send the SL anyway because it's suggested haha

PR #278, opened 2022-10-27
Update pd client to get token from osdctl config

dkeohane:
**Cluster Context Changes:**
- Now utilises the osdctl config to get pd oauth token from `pd_oauth_token` key instead of previously relying on pd cli config.
- Moved some of the aws event skipping into an isolated function as the if statement was getting too lengthy 
- Fixed bug where we weren't handling a case of cluster having no pd events causing stacktrace
- Minor output changes for better ux


fahlmant (approved):
/lgtm
/approve

PR #279, opened 2022-10-27
Minor fixes for the cpd script

fahlmant:
- Uses the -p flag to fill in awsProfile in the options
- Uses the correct filter for finding subnet associations for a route table, see https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-route-tables.html
- Add a closing }


rafael-azevedo (approved):
/lgtm

PR #283, opened 2022-11-08
Resovle issue for finding config path of Mac

weherdh:
Keeps getting the error of finding the config file while already put into `~/.config/osdctl`. After reading the [xdg](https://github.com/OpenPeeDeeP/xdg#locations-per-os) doc, the path for `xdg.ConfigFile` would be different for Linux and macOS.
Provide a fix so that it will support all os types.

### DIFF
--- a/pkg/osdctlConfig/osdctlConfig.go
+++ b/pkg/osdctlConfig/osdctlConfig.go
@@ -2,32 +2,22 @@ package osdctlConfig
 
 import (
 	"errors"
+	"fmt"
 	"os"
 
-	"github.com/adrg/xdg"
 	"github.com/spf13/viper"
 )
 
 const (
 	ConfigFileName = "osdctl"
-	ConfifFilePath = "~/.config"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/osdctlConfig/osdctlConfig.go
```diff
@@ -40,7 +30,8 @@ func EnsureConfigFile() error {
 
 	if err := viper.ReadInConfig(); err != nil {
 		return err
+	} else {
+		fmt.Println("Reading config file from " + configFilePath)
```
fahlmant, line 34:
```suggestion
		fmt.Println("Reading config file from " + configFilePath)
```

fahlmant: The else isn't needed here. Also, imo this should be put behind a verbose flag or something. Many of the osdctl commands have json or other formatted output, and this output may interfere with piping to jq for example

fahlmant (approved):
/approve
/lgtm

PR #284, opened 2022-11-14
Add version protection when running a command

fahlmant:
To ensure that a user isn't accidentally running an outdated version of osdctl, this PR adds a check to compare the current version of the ran binary with the latest tagged version released on GitHub. If the versions don't match, then the user is prompted to check if they want to explicitly run the binary or not. This should help protect the user from accidentally running an outdated or development versions of osdctl commands. 

I added the WIP tag to allow for further discussion

### DIFF
--- a/main.go
+++ b/main.go
@@ -4,9 +4,11 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/openshift/osdctl/cmd"
 	"github.com/openshift/osdctl/pkg/osdctlConfig"
+	"github.com/openshift/osdctl/pkg/utils"
 	"github.com/spf13/pflag"
 
 	"k8s.io/cli-runtime/pkg/genericclioptions"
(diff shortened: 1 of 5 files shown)


rafael-azevedo (approved):
I really like this idea. Saves us from having the run around when we are out of date :)

main.go
```diff
@@ -20,6 +22,31 @@ func main() {
 		return
 	}
 
+	latestVersion, err := utils.GetLatestVersion()
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+
+	if utils.Version != latestVersion {
```
iamkirkbater, line 31: I wonder if there's a way we could only ask this once a day or something like that?

rafael-azevedo: In my opinion that might defeat the purpose. Its easy enough to upgrade. Its probably safe as well to have everyone on the updated Binaries in case procedures or commands have changed.

dkeohane: +1 to Raf's comment - I'm in favour of being annoying to ensure we're not missing functionality

clcollins:
Are we ready to move forward with a decision here?  I'm generally LGTM to this change, and it's been sitting for a while now.

fahlmant:
If we're ok with this method going forward, then let me rebase and fix this up.

pkg/utils/version.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return latest, err
+	}
+
+	res, err := client.Do(req)
+	if err != nil {
+		return latest, err
+	}
+
+	body, err := ioutil.ReadAll(res.Body)
```
mjlshen, line 51: ioutil is deprecated
```suggestion
	body, err := io.ReadAll(res.Body)
```

main.go
```diff
@@ -20,6 +22,31 @@ func main() {
 		return
 	}
 
+	latestVersion, err := utils.GetLatestVersion()
+	if err != nil {
+		fmt.Println(err)
+		return
```
typeid, line 28: Bit late to the party, but wondering if we should run the commands assuming the version didn't change instead of returning here? If I'm not mistaken, this would make osdctl unusable if for whatever reason github isn't reachable.

PR #285, opened 2022-11-15
Fixes segfault when a route isn't using a CIDR block as its target

fahlmant:
Some routes don't use a CIDR block as its target, which causes the current code to segfault as it assumes the field is present for each route. This code fixes the issue by checking that the field is non nil before accessing it.

### DIFF
--- a/cmd/cluster/cpd.go
+++ b/cmd/cluster/cpd.go
@@ -182,9 +182,13 @@ func isSubnetRouteValid(awsClient aws.Client, subnetID string) (bool, error) {
 	}
 
 	for _, route := range describeRouteTablesOutput.RouteTables[0].Routes {
-		if *route.DestinationCidrBlock == "0.0.0.0/0" {
-			return true, nil
+		// Some routes don't use CIDR blocks as targets, so this needs to be checked
+		if route.DestinationCidrBlock != nil {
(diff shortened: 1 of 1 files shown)


mjlshen (approved):
/lgtm

cmd/cluster/cpd.go
```diff
@@ -182,9 +182,13 @@ func isSubnetRouteValid(awsClient aws.Client, subnetID string) (bool, error) {
 	}
 
 	for _, route := range describeRouteTablesOutput.RouteTables[0].Routes {
-		if *route.DestinationCidrBlock == "0.0.0.0/0" {
-			return true, nil
+		// Some routes don't use CIDR blocks as targets, so this needs to be checked
+		if route.DestinationCidrBlock != nil {
+			if *route.DestinationCidrBlock == "0.0.0.0/0" {
```
bdmiller3, line 187: Could we add this to the above if condition for a one-liner?

`if route.DestinationCidrBlock != nil && *route.DestinationCidrBlock == "0.0.0.0/0 {`

mjlshen: I don't think that would work because if `*route.DestinationCidrBlock` is evaluated it'll be a nil pointer dereference and panic. This way the order is enforced and we ensure we don't have a nil pointer to start with

PR #286, opened 2022-11-16
recommend the correct command for the network verifier

yithian:
since the network verifier is part of osdctl, the output of `osdctl cluster cpd` should give the command to run it from here instead of the standalone command


fahlmant (approved):
/approve
/lgtm

PR #287, opened 2022-11-17
Improve internal servicelog template summary

karthikperu7:
Improving internal servicelog summary so that the message is not shared with the customer directly.

Built and tested on a cluster, works ok.


mjlshen (approved):
/lgtm

PR #288, opened 2022-11-17
Added JIRA cards to cluster context cmd

dkeohane:
Now listing OHSS JIRA cards created for a cluster. 

Example output: 
```
============================================================
Cluster JIRAs
============================================================
[OHSS-1234](Incident/Medium): mycluster has gone missing [Status: New]
- Link: https://issues.redhat.com/browse/OHSS-1234

[OHSS-1324](Incident/Medium): [mycluster]: Cluster gone missing [Status: Resolved]
- Link: https://issues.redhat.com/browse/OHSS-1324
```

PR #290, opened 2022-11-21
'cluster context' CLI help improvements

mrbarge:
This PR is a small cosmetic update in relation to the `cluster context` command:

- Added the Jira token (#288) as a parameter to `cluster context`. This was done mainly so that I could..
- ..add additional information on where/how to generate PD and Jira tokens to the CLI helper text (if there is a better way for PD, I'm happy to update that). This was done so that I could..
- re-generated the command documentation to include this information (which also incorporated a handful of other recent changes requiring doc updates [0]) 

I also updated the helper text to make use of the existing config filename constant when referencing the config file path, rather than a hand-written one (I would have used `viper.ConfigFileUsed()` instead, but it contains the full path including home directory of the user generating the docs, which then makes the generated documentation wrong)

[0] Future idea: turning the documentation generation into a github action that auto-generates it, similar to MCC and its deployment files?


dkeohane (approved):
Thanks for the changes Matt, much cleaner :) 

/lgtm

fahlmant (approved):
/lgtm
/approve

PR #291, opened 2022-11-29
Better handling of edge cases around default routes

mjlshen:
The older commit was the result of running `go fmt ./...`

Summary of the changes in the second commit:
- We weren't closing our OCM client connection
- Support of external ID (it's a minor thing since PagerDuty generally provides the internal ID for CPD)
- A bit more context on some of the errors
- Suggest `ocm backplane cloud console` for GCP clusters
- Return `false, nil` instead of `false, err` https://github.com/openshift/osdctl/pull/291/files#diff-8d546d74a081da0204b85af938301169de6c745e6df0e76c9d25397747e96edfL194-R186 when we finish validation of subnet route tables successfully, but determine that a configuration is invalid
- Fill in the cluster id when suggesting to send a service log. TODO it should just try to send on its own and ask for confirmation like normal.

Previously this would just return the contents of an error if no default route was found:
```
error: no default route exists to the internet in subnet subnet-057ae41a68a4924c6
```

Now it returns the intended message
```bash
❯ ./osdctl cluster cpd -C b9b236d4-7505-42b3-b221-f4bba6732013 --profile rhcontrol
Checking if cluster has become ready
Checking if cluster DNS is ready
Checking if OCM error code is already known
Checking if cluster is GCP
Generating AWS credentials for cluster
Checking BYOVPC to ensure subnets have valid routing
error: subnet subnet-057ae41a68a4924c6 does not have a default route to 0.0.0.0/0
 Run the following to send a SerivceLog:
 osdctl servicelog post b9b236d4-7505-42b3-b221-f4bba6732013 -t https://raw.githubusercontent.com/openshift/managed-notifications/master/osd/aws/InstallFailed_NoRouteToInternet.json
```

### DIFF
--- a/cmd/cluster/cpd.go
+++ b/cmd/cluster/cpd.go
@@ -87,8 +87,7 @@ func (o *cpdOptions) run() error {
 	fmt.Println("Checking if cluster is GCP")
 	// If the cluster is GCP, give instructions on how to get console access
 	if cluster.CloudProvider().ID() == "gcp" {
-		fmt.Printf("This command doesn't support GCP yet. Needs manual investigation\n Get the project ID from this command on hive: oc getprojectclaim -n uhc-production-$CLUSTER_INT_ID\nThen use this URL to access the GCP console: https://console.cloud.google.com/home/dashboard?project=${GCP_PROJECT_ID}\n")
-		return nil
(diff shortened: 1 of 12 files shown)


fahlmant (approved):
/approve
/lgtm

PR #292, opened 2022-11-30
Upgrade to Go 1.18

mjlshen:
This also allows us to upgrade to the latest goreleaser again


Dockerfile
```diff
@@ -8,6 +8,6 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 LABEL io.openshift.managed.name="osdctl" \
       io.openshift.managed.description="OSD related command line utilities"
 
-COPY --from=0 /src/dist/osdctl_linux_amd64/osdctl /bin/osdctl
+COPY --from=0 /src/dist/osdctl_linux_amd64_v1/osdctl /bin/osdctl
```
fahlmant, line 11: Is this due to a change in goreleaser?

mjlshen: Yeah, I've been looking on how if it's configurable since I didn't change that behavior (but haven't found anything so far)

mjlshen: It makes `darwin_amd64_v1` as well as `linux_amd64_v1`, but just `darwin_arm64` and `linux_arm64` like normal....so not sure what it's doing with specifically the amd64 builds

fahlmant: ack. No worries, I think its fine in this case. Just wanted to understand where it was coming from.

fahlmant (approved):
/lgtm
/approve

PR #294, opened 2022-12-06
Adds deprecation message to the capability command.

iamkirkbater:
With the capabilities being soon managed within ocm-resources, let's add the deprecation to the command now to be ready for the coming changes.

Part of [SDB-3204](https://issues.redhat.com//browse/SDB-3204)

### DIFF
--- a/cmd/capability/add.go
+++ b/cmd/capability/add.go
@@ -24,7 +24,7 @@ func newAddCmd() *cobra.Command {
 			cmdutil.CheckErr(ops.complete(cmd, args))
 			cmdutil.CheckErr(ops.run(cmd, args[0]))
 		},
-		Deprecated: "This command is being deprecated in lieu of using ocm-resources. Soon, this command will not work, and you will have to follow the SOP v4/howto/capabilities.md.",
+		Deprecated: "This command is being deprecated in lieu of using git-backed capabilities. Soon, this command will not work, and you will have to follow the SOP at https://github.com/openshift/ops-sop/v4/howto/capabilities.md.",
 	}
 
 	addCmd.Flags().StringVarP(&ops.OrganizationID, "organization-id", "g", "", "Specify an OCM Organization to apply a capability to")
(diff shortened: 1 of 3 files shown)


cmd/capability/add.go
```diff
@@ -24,6 +24,7 @@ func newAddCmd() *cobra.Command {
 			cmdutil.CheckErr(ops.complete(cmd, args))
 			cmdutil.CheckErr(ops.run(cmd, args[0]))
 		},
+		Deprecated: "This command is being deprecated in lieu of using ocm-resources. Soon, this command will not work, and you will have to follow the SOP v4/howto/capabilities.md.",
```
clcollins, line 27: Should this just be a URL to that location?  I realize it's a private repo, but I don't think we're exposing anything by linking to it, and it would be one less step to be able to click the link to open.

clcollins (changes requested):
Generally LGTM, but just that one question about directly linking to the SOP

cmd/capability/add.go
jaybeeunix: i think the following is a good guidance for links into private stuff:

- links the just mention an alert name are probably fine, even if the sop only exists in ops-sop
   - the alerts are probably defined in an open place (upstream, our MCC), so anyone can see what's causing the thing to fire
   - we have a goal of open sourcing both our own alerts and sops, so this link would only hint at something private for a "short" period
- links that point into something vpn-private generally shouldn't be in anything public-facing.
   - there's too high of a risk of exposing internal intranet workings and we don't have the authority to assess the risk to those systems.
- links to private repos that probably won't be public, but are generically worded are also probably OK to share
   - for example, "'thiscmd magic' is deprecated: See https://github.com/.....howto/do-magic" doesn't tell the public anything more than "thiscmd magic" did.

for this specific pr, at most i'd maybe change "ocm-resources" to ocm or something (since i don't think we talk about ocm-resources anywhere pub)

fahlmant (approved):
/approve
/lgtm

PR #297, opened 2022-12-11
MVP smart network-verifier

mjlshen:
## Summary of the New Goals

- "Just" specify `--cluster-id`, no seriously, that's it. The tooling should do all the rest and return errors and guidance when it can't. This was recently enabled with https://github.com/openshift/osd-network-verifier/pull/161
- Also documentation and some testing :P

A picture (or maybe just more words?) is worth a thousand words, so here's an example of a cluster with a transparent cluster-wide proxy (e.g. no additional trust bundle). I've redacted IDs, but notice how with only specifying the ID, we get a correctly selected private subnet, the master node's security group, and `HttpProxy` and `HttpsProxy` automatically populated.

```bash
❯ osdctl network verify-egress --cluster-id REDACTED 
2022/12/10 22:17:34 getting AWS credentials from backplane-api
2022/12/10 22:17:37 detected BYOVPC PrivateLink cluster, using first subnet from OCM: subnet-REDACTED
2022/12/10 22:17:37 searching for security group by tags: kubernetes.io/cluster/REDACTED=owned and Name=REDACTED-master-sg
2022/12/10 22:17:38 using security-group-id: sg-REDACTED
2022/12/10 22:17:38 running with config: &{Timeout:2s Ctx:context.TODO SubnetID:subnet-REDACTED CloudImageID:ami-0af4bdc3e6f25374f InstanceType:t3.micro Proxy:{HttpProxy:http://REDACTED:8080 HttpsProxy:http://REDACTED:8080 Cacert: NoTls:false} Tags:map[Name:osd-network-verifier osd-network-verifier:owned red-hat-managed:true] AWS:{KmsKeyID: SecurityGroupId:sg-REDACTED} GCP:{Region: Zone: ProjectID: VpcName:}}
2022/12/10 22:17:40 Created instance with ID: i-0f3d75c4f874eca93
Summary:
printing out failures:
...
 -  egressURL error: Unable to reach REDACTED
...
```

and here's an example of a cluster-wide-proxy with an additional trust bundle, stopping SREs from mistakenly getting incorrect results:

```bash
❯ ~/git/openshift/osdctl/osdctl network verify-egress --cluster-id REDACTED
2022/12/10 22:29:12 getting AWS credentials from backplane-api
2022/12/10 22:29:13 REDACTED has an additional trust bundle configured, but no --cacert supplied
```

## How it Works

- Expects an active `ocm-backplane` tunnel
- Uses the provided OCM internal/external cluster id or cluster name to look up the cluster in OCM, which importantly provides insight into its `InfraId`, region, and whether it has a cluster-wide-proxy
- Retrieves AWS credentials from `backplane-api` and assembles then in-memory with no need for any local AWS profile or environment variables exported.
- Searches AWS for the clusters private subnets and the master node's security group.
- Assembles input to `osd-network-verifier` with all that info, failing if, for example, there's a cluster-wide proxy with a custom trust bundle, but the SRE didn't supply `--cacert`
- ???
- Profit

## Removals

These are options exposed generally exposed by `osd-network-verifier`, but in general shouldn't be used by SREs when investigating and should be reserved for `osd-network-verifier` directly.

- `--image-id`
- `--instance-type`
- `--cloud-tags`
- `--timeout`
- `--kms-key-id` - The support role doesn't have permissions to BYOKMS cluster KMS keys anyway
- `--profile` - No longer does anything, we're passing an AWS Config to osd-network-verifier now instead of a profile
- `--subnet-id` and `--security-group` are no longer required flags

> But does this still work if there's a backplane-api or OCM outage?

Yes and no.

- Yes - the previous workflow of configuring AWS credentials locally on your own first (e.g. exporting `AWS_*` variables or using `aws configure`) still works. SREs would just need to manually specify all the flags again, `--subnet-id`, `--security-group`, `--https-proxy`, etc.
- No - the `--cluster-id` flag wouldn't work.

## Limitations and Future Improvement

- I didn't write logic for intelligently selecting private subnets from non-PrivateLink BYOVPC clusters. It's possible and can be added in the near future (e.g. subnets with routes to an IGW == public subnet, so skip those), but for now I just return a message that says to find and specify a `--subnet-id` on your own. Otherwise, this tool should do the right thing.
- It doesn't test all of the private subnets for multi-AZ clusters, only one.
- It doesn't test egress with the worker node security group, just the master.

In general networking tends to be messed up at the VPC level and not a specific subnet or security group, but if there's enough value seen, this could easily be extended to fire off runs to test all subnets and at least one subnet with both security groups to cover all the cases.


typeid (commented):
Tested on staging. 

Great work! Just some minor typo nitpicks and missing updated documentation generation, rest lgtm. Love that you kept the command usable even when backplane-api is down.

cmd/network/verification.go
```diff
@@ -4,164 +4,352 @@ import (
...
+  # Run against a cluster registered in OCM with a cluster-wide-proxy
+  ocm-backplane tunnel -D
+  touch cacert.txt
+  osdctl network verify-egress --cluster-id my-rosa-cluster --cacert cacert.txt
+
+  # Override automatic selection of a subnet or security group id
+  ocm-backplane tunnel -D
+  osdctl network verify-egress --cluster-id my-rosa-cluster --subnet-id subnet-abcd --security-group sg-abcd
+
+  # (Not recommended) Run against a specific VPC, without specifying cluster-id
```
typeid, line 71: **thought:** The unrecommended way to execute the command is still at the bottom for the README, maybe we should update it there as well? Although it still works that way.

cmd/network/verification.go
```diff
@@ -4,164 +4,352 @@ import (
...
 
-	p := proxy.ProxyConfig{
-		HttpProxy:  config.httpProxy,
-		HttpsProxy: config.httpsProxy,
-		Cacert:     config.CaCert,
-		NoTls:      config.noTls,
+	// Fill in subnetID
+	subnetId, err := e.getSubnetId(context.TODO())
+	if err != nil {
+		log.Fatalf("failed to detect subnet")
```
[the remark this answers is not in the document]
mjlshen: Thanks for catching this, they were artifacts of a previous refactor that I forgot to cutover. All of those should be bubbling up and caught by `log.Fatal`'s in `func (e *egressVerification) run(ctx context.Context)`

cmd/network/verification.go
```diff
@@ -4,164 +4,352 @@ import (
...
-		Short: "Verify essential openshift domains are reachable from given subnet ID.",
-		Long:  `Verify essential openshift domains are reachable from given subnet ID.`,
-		Example: `For AWS, ensure your credential environment vars 
-AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY (also AWS_SESSION_TOKEN for STS credentials) 
-are set correctly before execution.
-
-# Verify that essential openshift domains are reachable from a given SUBNET_ID
-osdctl network verify-egress --subnet-id $(SUBNET_ID) --security-group $(SECURITY_GROUP) --region $(AWS_REGION)`,
+		Short: "Verify required URLs required for a fully supportable AWS OSD/ROSA cluster.",
+		Long: `Verify required URLs required for a fully supportable AWS OSD/ROSA cluster.
```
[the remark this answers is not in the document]
mjlshen: I gave it another shot and updated it to:

`Verify an AWS OSD/ROSA cluster can reach all required external URLs necessary for full support.`

What do you think?

cmd/network/verification.go
mjlshen: > maybe we should update it there as well?

Just want to clarify what you mean here. I wrote that "unrecommended way" example mostly just to show/explain that it could still be done that way without backplane/OCM. I'm ok with removing it too if you think it's confusing to show, but wanted to make sure I understood your suggestion fully.

Thanks for all the feedback in general!

cmd/network/verification.go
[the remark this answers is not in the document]
typeid: That makes it clear, thanks!

cmd/network/verification.go
typeid: Sorry for the confusing comment. The documentation in `verification.go` is perfectly fine, I meant that the project `README.md` still contains the "not recommended" way: https://github.com/openshift/osdctl/blob/master/README.md#osd-network-verifier

Just a thought though, not sure if it should be changed.

typeid:
/lgtm
Last open thread is just a non blocking [thought](https://conventionalcomments.org/).

cmd/network/verification.go
mjlshen: Oh I didn't even notice that! Yes, definitely agreed that should be changed.

clcollins:
/lgtm

This is AMAZING, well done!

fahlmant:
A few things:
- Was there an OSD card tracking this? This is quite a large change but I don't see it attached to a specific prioritized card or shift-improvement card. 
- This pins osdctl to not used a tagged version of the verifier code. The intention of using the tag is to make sure the cli stays in line with the official releases of the verifier. It was also intentional that cluster service and osdctl were pinned to the same version, now there's a divergence between the two and could lead to different results. 
- I'm not a fan of removing all the options from this command. osdctl is intended to be the SRE interface into using the osd-network-verifier in any way. Removing the configuration options for a manual run is a regression of intended feature set for this command and locks SRE into only one way to debug things.
- Using "not recommended" language in the README gives a negative connotation, but running manually on a VPC is not a bad thing. It sometimes may be necessary or can just be a good exercise to understand how the verifier runs, and how a VPC can be tested. 
- Locking this to only the backplane credential method can be tenuous if the backplane api is down or not working. At least enabling the alternative method through public OCM endpoints that `osdctl account cli/console` should be added, but a subsequent PR is fine in this case.

mjlshen:
> Was there an OSD card tracking this? This is quite a large change but I don't see it attached to a specific prioritized feature card or shift-improvement card.

Yeah, sorry didn't link it in this PR. It was the same Epic for the new AWS credential interface for osd-network-verifier: OSD-14120

> This pins osdctl to not used a tagged version of the verifier code. The intention of using the tag is to make sure the cli stays in line with the official releases of the verifier. It was also intentional that cluster service and osdctl were pinned to the same version, now there's a divergence between the two and could lead to different results.

Can we cut a release then? I don't agree that CS and osdctl need to be pinned to the same version. It should be the responsibility of osd-network-verifier to function correctly and for CS and osdctl to upgrade their dependencies independently. If there's a bug in osd-network-verifier then I agree both should upgrade past the bugfix...

mjlshen:
> I'm not a fan of removing all the options from this command. osdctl is intended to be the SRE interface into using the osd-network-verifier in any way. Removing the configuration options for a manual run is a regression of intended feature set for this command and locks SRE into only one way to debug things.

Depends on what "any way" means haha

`osd-network-verifier` CLI exists, my goal is for `osdctl` to do more heavy-lifting for SREs. If an SRE wants to be able to configure every option for `osd-network-verifier` imo that CLI is the perfect tool for the job, whereas `osdctl` should give SREs the correct response for production clusters as easily as possible. For example, if an SRE is running this against a BYOKMS cluster and tries to optionally specify a `kms-key-arn` and it fails - how do we communicate this is expected? Compared to an SRE going out of their way to do this via `osd-network-verifier`. Similar for choosing an incorrect instance type that doesn't support Nitro...or setting timeout...

> Using "not recommended" language in the README/docs gives a negative connotation, but running manually on a VPC is not a bad thing. It sometimes may be necessary or can just be a good exercise to understand how the verifier runs, and how a VPC can be tested.

It's not a bad thing, but it is "not recommended" when running this against real clusters registered in OCM. Related to my previous response, but imo these kinds of exercises/learnings are a better fit for `osd-network-verifier` and not osdctl, which SREs use when responding to production alerts.

> Locking this to only the backplane credential method can be tenuous if the backplane api is down or not working. At least enabling the alternative method through public OCM endpoints that osdctl account cli/console should be added, but a subsequent PR is fine in this case.

It's not locked in - it still works if you setup AWS creds beforehand and manually specify a subnet/security group ID

fahlmant:
- Thanks for the card!
- We can cut a new version, just need to ask the verifier repo maintainers and have a justification. There's a process that needs to be ran for this.
- The `osd-network-verifier` binary is mostly an example for how to integrate with the verifier library, and those cli commands will likely be removed. `osdctl` is meant to be the SRE interface for running verifier tests. Sorry if that wasn't clear, the work should be tracked in our epics and we can cut some new cards if its not. 
- I would just ask that the doc reflect that its not recommended for that specific flow. Its minor, but I think having positive language ("This is the recommended path") vs. negative ("this is not the recommended path") is better overall, but that's just my preference.
- You're right, there are multiple ways, all I'm requesting is that the other flow is supported too.

mjlshen:
Wrote up https://issues.redhat.com/browse/OSD-14257 to add back all the optional inputs for `osd-network-verifier`

I'll make a future PR to update the wording as I don't think I need a card for that and just reference the epic.

PR #298, opened 2022-12-12
Fix README goproxy requirement

typeid:
**Why?**
The proxy url needs to be the first priority in the `GOPROXY` env, otherwise `direct` might be used, and we run into the same error as when we don't configure the proxy.

`github.com/openshift/api@v3.9.1-0.20191111211345-a27ff30ebf09+incompatible: invalid pseudo-version: preceding tag (v3.9.0) not found`


fahlmant (approved):
/lgtm
/approve

PR #299, opened 2022-12-13
Automatically run the network-verifier after running the CPD utility

yithian:
Running the egress network-verifier is the next step after running the
Cluster Provisioning Delay subcommand, so we may as well have it run
automatically as part of the CPD subcommand.

### DIFF
--- a/cmd/cluster/cpd.go
+++ b/cmd/cluster/cpd.go
@@ -156,9 +156,19 @@ func (o *cpdOptions) run() error {
 			}
 
 			output := verifier.ValidateEgress(awsVerifier, vei)
-			_, _, errors := output.Parse()
-			for _, err := range errors {
-				return err
+			failures, exceptions, errors := output.Parse()
+			if len(failures) != 0 || len(exceptions) != 0 || len(errors) != 0 {
+				resultString := "Network verifier results:\n"
+				for _, failure := range failures {
+					resultString = resultString + failure.Error() + "\n"
+				}
+				for _, exception := range exceptions {
+					resultString = resultString + exception.Error() + "\n"
+				}
+				for _, err := range errors {
+					resultString = resultString + err.Error() + "\n"
+				}
+				return fmt.Errorf(resultString)
 			}
 		}
 		return nil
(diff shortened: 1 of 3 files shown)


mjlshen:
/hold
This is somewhat taken care of by #297, which doesn't automatically run the network verifier, but does automatically select the subnet-id/security-group-id for you give a cluster-id.

I think the main problem with this current PR is that it doesn't select a security-group-id.

yithian:
> /hold This is somewhat taken care of by #297, which doesn't automatically run the network verifier, but does automatically select the subnet-id/security-group-id for you give a cluster-id.
> 
> I think the main problem with this current PR is that it doesn't select a security-group-id.

I've included another to select a security group ID for the network verifier (shamelessly ~~stolen from~~ inspired by your changes to `osdctl network verify-egress`)

mjlshen:
/unhold
/lgtm

It would be nice if we could've just called https://github.com/openshift/osdctl/blob/master/cmd/network/verification.go#L100 to not duplicate logic, but we can iterate

cmd/cluster/cpd.go
```diff
@@ -111,8 +116,51 @@ func (o *cpdOptions) run() error {
...
+			}
+
+			awsVerifier, err := onvUtils.GetAwsVerifier(cluster.Region().DisplayName(), o.awsProfile, false)
+			if err != nil {
+				fmt.Printf("could not build awsVerifier %v", err)
+				os.Exit(1)
+			}
+
+			output := verifier.ValidateEgress(awsVerifier, vei)
+			_, _, errors := output.Parse()
```
boranx, line 159: I think we can mention the other outputs as well (like 
```go 
failures,exceptions, errors := output.Parse()
```
_failures_ stores actual egress validation failures
_exceptions_ refers to handled exceptions like vm issues
_errors_ which are most likely cloud provider permissions issues or unhandled exceptions etc. 
Those would be returned by output: https://pkg.go.dev/github.com/openshift/osd-network-verifier@v0.1.0/pkg/output#Output.Parse mb we can clarify the manner if it's confusing

mjlshen:
This has been chilling for a bit so I did do:

> It would be nice if we could've just called https://github.com/openshift/osdctl/blob/master/cmd/network/verification.go#L100 to not duplicate logic, but we can iterate

in #330

PR #300, opened 2022-12-14
Elevate for performing control plane node drain

mrbarge:
# What is being changed

- Elevate to `backplane-cluster-admin` for the first (non-forceful) attempt at a control plane node drain.

# Why

The control plane node drain resize command presently results in an error because it can't list pods in all namespaces, therefore always directing the SRE towards a skip/force/cancel path which is not necessarily needed.


typeid:
It was added to the forced drain call, but was indeed still missing for the normal drain. 

/lgtm

fahlmant (approved):
/lgtm
/approve

PR #301, opened 2022-12-16
Add back advanced flags for `verify egress`

mjlshen:
[OSD-14257](https://issues.redhat.com/browse/OSD-14257) This lets us delete the `osd-network-verifier/cmd` for AWS

Also makes the struct public so it can be used in cases like https://github.com/openshift/osdctl/pull/299 instead of having to duplicate code.

### DIFF
--- a/cmd/network/verification.go
+++ b/cmd/network/verification.go
@@ -22,25 +22,37 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const nonByovpcPrivateSubnetTagKey = "kubernetes.io/role/internal-elb"
+const (
+	nonByovpcPrivateSubnetTagKey = "kubernetes.io/role/internal-elb"
+	defaultInstanceType          = "t3.micro"
+	defaultTimeout               = 2
+)
 
-type egressVerification struct {
+type Egress struct {
 	awsClient egressVerificationAWSClient
 	cluster   *cmv1.Cluster
 	log       logging.Logger
 
-	// clusterId is the cluster identifier that will be used to query backplane for AWS credentials to build an AWS client
-	clusterId       string
-	region          string
-	subnetId        string
-	securityGroupId string
-	debug           bool
-	caCert          string
-	noTls           bool
+	// ClusterId is the cluster identifier that will be used to query backplane for AWS credentials to build an AWS client
+	ClusterId       string
+	Region          string
+	SubnetId        string
+	SecurityGroupId string
+	Debug           bool
+	Cacert          string
+	NoTls           bool
+
+	// Generally only used when developing osd-network-verifier
+	HttpProxy    string
+	HttpsProxy   string
+	Timeout      int
+	KmsKeyId     string
+	InstanceType string
+	AmiId        string
 }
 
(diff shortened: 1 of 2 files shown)


cmd/network/verification.go
```diff
@@ -22,25 +22,37 @@ import (
...
-	clusterId       string
-	region          string
-	subnetId        string
-	securityGroupId string
-	debug           bool
-	caCert          string
-	noTls           bool
+	// ClusterId is the cluster identifier that will be used to query backplane for AWS credentials to build an AWS client
+	ClusterId       string
+	Region          string
```
boranx, line 38: looks to be parameters were previously marked as private, is there a reason for flipping this to public again?

boranx (commented):
thanks @mjlshen , looks good! I've left small questions but all are optional & for getting a clearer context

cmd/network/verification.go
mjlshen: Ah yeah for this I tried to call it out in the description:

> Also makes the struct public so it can be used in cases like https://github.com/openshift/osdctl/pull/299 instead of having to duplicate code.

Sometimes other osdctl commands want to automatically run this afterwards, if it's public they can easily do this. Another approach would be to leave this private and create getter/setter functions

cmd/network/verification.go
```diff
@@ -68,24 +80,30 @@ func NewCmdValidateEgress() *cobra.Command {
...
+	validateEgressCmd.Flags().StringVar(&e.HttpProxy, "http-proxy", "", "(optional) manually specify http proxy")
+	validateEgressCmd.Flags().StringVar(&e.HttpsProxy, "https-proxy", "", "(optional) manually specify https proxy")
+	validateEgressCmd.Flags().StringVar(&e.KmsKeyId, "kms-key-id", "", "(optional) manually specify an AWS KMS Key ID to encrypt the EC2 instance's volume with")
+	validateEgressCmd.Flags().StringVar(&e.AmiId, "ami-id", "", "(optional) manually specify the AMI to run")
+	validateEgressCmd.Flags().StringVar(&e.InstanceType, "instance-type", defaultInstanceType, "(optional) manually the AWS EC2 instance type")
+	validateEgressCmd.Flags().IntVar(&e.Timeout, "timeout", defaultTimeout, "(optional) manually specify timeout in seconds for individual egress health checks")
+	validateEgressCmd.Flags().BoolVarP(&e.Debug, "debug", "v", false, "(optional) if provided, enable additional debug-level logging")
+
+	// If a cluster-id is specified, don't allow the foot-gun of overriding Region
+	validateEgressCmd.MarkFlagsMutuallyExclusive("cluster-id", "Region")
```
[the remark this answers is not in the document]
mjlshen: Yeah, that makes sense. For subnet-id/security-group I like keeping the option of overriding (in case you want to manually try a specific subnet/security-group), but I think making them exclusive for all the proxy configuration is a good idea for example. I'll add this!

cmd/network/verification.go
boranx: ah ok, then let's keep it as is, I'd prefer using public than the getter/setter hell :)

PR #308, opened 2022-12-28
Updated struct data type while unmarshal clusterkubeconfig operation

shitaljante:
Related to - https://issues.redhat.com/browse/OSD-13939
 
Changes include:
1. Added "NewYAMLOrJSONDecoder" to use "json:" annotations while parsing yaml data. 
2. Marshal data to YAML while storing the local kubeconfig file.

Why: 
Currently client-go does not declare yaml go struct tags for the kubeconfig types(clientcmdapiv1.Config). 
In order to marshal the data object used a common method which supports both format yaml and json.

### DIFF
--- a/cmd/cluster/access/access.go
+++ b/cmd/cluster/access/access.go
@@ -70,7 +70,6 @@ func NewCmdAccess(streams genericclioptions.IOStreams, flags *genericclioptions.
 	return accessCmd
 }
 
-
 // clusterCmdComplete verifies the command's invocation, returning an error if the usage is invalid
 func accessCmdComplete(cmd *cobra.Command, args []string) error {
 	if len(args) != 1 {
--- a/cmd/cluster/access/access.go
+++ b/cmd/cluster/access/access.go
@@ -1,6 +1,7 @@
 package access
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


mrbarge:
@shitaljante thanks for making this change. I was able to verify a populated kubeconfig for a private cluster so it functionally seems to be correct.

I think it is worth being clear in the PR description why this change is doing what it's doing, specifically declaring the types. From what I can see, this seems to be because the client-go does not declare yaml go struct tags for the kubeconfig types, is that correct?

I see that there is a way to successfully unmarshal the yaml regardless of the absent go struct tags by using `NewYAMLOrJSONDecoder` (from `k8s.io/apimachinery/pkg/util/yaml`), which would not require declaring a whole new type.

```
	formattedKubeconfig := clientcmdapiv1.Config{}
	d := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(data), len(data))
	if err := d.Decode(&formattedKubeconfig); err != nil {
             // handle error
	}
```

From there, it could potentially be marshalled back to yaml after first marshalling to json (so that the json go struct tags are respected) and then converting the json to yaml. 

I would really like to see that explored if possible, because it might be a lot cleaner. Would you mind having a look? Happy to help with this if needed so let me know.

Otherwise, if that isn't possible, may I suggest that the type declarations be moved outside of the `access.go` file and maybe migrated into the `k8s` package in `pkg/k8s` instead.

shitaljante:
Thank you @mrbarge ! 
Updated changes as per suggestion,PTAL.

mrbarge:
Fantastic, thank you @shitaljante , this makes it a lot nicer.
I gave it a quick test on a private cluster and it seemed to work well.

/lgtm

fahlmant (approved):
/approve
/lgtm

shitaljante:
@fahlmant Can you please let me know if anything you are expecting here to unhold PR?

fahlmant:
/hold cancel
Was just waiting for the change freeze so this didn't get accidentally merged before it was lifted

shitaljante:
Thank You.
I really appreciate if I can get a lgtm label again.

PR #311, opened 2023-01-13
network verify-egress also supports osdtrial

mjlshen:
osd-network-verifier also returns valid results for osdtrial clusters, so adding it to the allowed list.


iamkirkbater:
Weird that this hasn't merged yet :/

/lgtm

PR #313, opened 2023-01-16
Add AWS_REGION env var

geowa4:
When using `osdctl account cli -i ... -p ... -o env` to set local environment variables for `rosa`, I encountered an error where it needed the region set. Per https://docs.aws.amazon.com/sdkref/latest/guide/feature-region.html, AWS_REGION appears to be the more widely accepted variable, however AWS_DEFAULT_REGION appears to be used by AWS CLI v1 and boto3.

### DIFF
--- a/cmd/account/cli.go
+++ b/cmd/account/cli.go
@@ -159,10 +159,11 @@ func (o *cliOptions) run() error {
 	}
 
 	if o.output == "env" {
-		fmt.Printf("AWS_ACCESS_KEY_ID=%s \nAWS_SECRET_ACCESS_KEY=%s \nAWS_SESSION_TOKEN=%s \nAWS_DEFAULT_REGION=%s\n",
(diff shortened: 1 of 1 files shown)


cmd/account/cli.go
```diff
@@ -159,11 +159,12 @@ func (o *cliOptions) run() error {
 	}
 
 	if o.output == "env" {
-		fmt.Printf("AWS_ACCESS_KEY_ID=%s \nAWS_SECRET_ACCESS_KEY=%s \nAWS_SESSION_TOKEN=%s \nAWS_DEFAULT_REGION=%s\n",
+		fmt.Printf("AWS_ACCESS_KEY_ID=%s\nAWS_SECRET_ACCESS_KEY=%s\nAWS_SESSION_TOKEN=%s\nAWS_DEFAULT_REGION=%s\nAWS_REGION=%s\n",
```
geowa4, line 162: I see [line 52](https://github.com/openshift/osdctl/pull/313/files#diff-c20c31ee54301dc0f1a15f46ffc55b79d7dc3e86db5723388b5290caf1163658R152) using `%q`. While I don't think these values will ever have a space in them, should I make that change in this PR as well?

iamkirkbater (commented):
/lgtm

PR #314, opened 2023-01-17
don't prompt to upgrade if running currentVersion-next

yithian:
When compiling from the latest source code, osdctl would prompt the user to upgrade to the latest version because it automatically appends "-next" during compilation. Since the latest commit in the master branch is newer than the latest tagged build, osdctl shouldn't prompt the user to downgrade.


mjlshen (approved):
/lgtm

fahlmant (changes requested):
The intention of adding this check was be based on the current tagged version. Compiling from latest could include new code merged into that hasn't been fully tested or functional, so it's intentional that this check is exclusively based on the tagged released version on GitHub

yithian:
If someone is compiling from the latest commit, they already know that there may be changes that haven't made it into a full release yet; that's probably why they're compiling from the latest commit. So in that case, bothering the SRE every time they run the command seems unnecessary.

It also frustrates development. Iterating on a change and having to go "yes, I know this isn't the latest release, that's why I'm testing these changes" every time feels like a pointless irritant.

Would it be acceptable to add a `--yes-i-know-this-is-a-prerelease-build` flag to bypass the check? Or change it to a non-blocking warning for `<current-release>-next` builds?

fahlmant:
The original catalyst to add this flag was code from an open PR being run against a customer cluster accidentally. The gate here is specifically to ensure developers know they're running a development version of the binary.

A flag (--skip-version-check) still requires an explicit acknowledgement in the action of providing the flag, so I think that would be a good compromise

mjlshen (approved):
/lgtm

yithian:
@fahlmant Does this look better? I've added a `--skip-version-check` flag and removed the auto-bypass for *-next builds

fahlmant (approved):
/lgtm
/approve

PR #316, opened 2023-01-20
add hypershift support to egress validations

boranx:
This PR provides credential support for cases if a given cluster is hypershift (including stage and int)
This allows us to execute some of the osd checks against hypershift  such as egress:
(I haven't checked other functionalities, they may depend on additional resources that need to be fetched, too)

Test:
```shell
./ocm-login stage
ocm describe cluster HCP_CLUSTER_ID  | grep Management
Management Cluster:     hs-mc-8c6cdrxxx

dist/osdctl_darwin_arm64/osdctl network verify-egress --cluster-id HCP_CLUSTER_ID
2023/01/20 19:32:46 getting AWS credentials from backplane-api
2023/01/20 19:32:48 detected BYOVPC Hypershift cluster, using the subnets of the worker nodes(private): subnet-0dabcd1b9cb382df6
2023/01/20 19:32:48 searching for security group by tags: api.openshift.com/id=HCP_CLUSTER_ID
2023/01/20 19:32:49 using security-group-id: sg-0037647435c5c4f8c
2023/01/20 19:32:49 running with config: &{Timeout:2s Ctx:context.TODO SubnetID:subnet-0e3a4046c1c2f1078 CloudImageID:ami-03ab344882b539e44 InstanceType:t3.micro Proxy:{HttpProxy: HttpsProxy: Cacert: NoTls:false} Tags:map[Name:osd-network-verifier osd-network-verifier:owned red-hat-managed:true] AWS:{KmsKeyID: SecurityGroupId:sg-0037647435c5c4f8c} GCP:{Region: Zone: ProjectID: VpcName:}}
2023/01/20 19:32:51 Created instance with ID: i-0176573e578fc0abf
Summary:
All tests pass!
2023/01/20 19:34:53 All tests pass
```

Ref: https://gitlab.cee.redhat.com/service/backplane-cli/-/merge_requests/234

### DIFF
--- a/cmd/network/verification.go
+++ b/cmd/network/verification.go
@@ -316,6 +316,9 @@ func (e *egressVerification) getSubnetId(ctx context.Context) (string, error) {
 		if len(instance.Reservations[0].Instances[0].NetworkInterfaces) == 0 {
 			return "", fmt.Errorf("found 0 network interfaces of the worker node: %s, consider the --subnet-id flag", *instance.Reservations[0].Instances[0].InstanceId)
 		}
+		if len(instance.Reservations[0].Instances[0].NetworkInterfaces) > 1 {
+			return "", fmt.Errorf("more than one interface found on the worker node: %s, consider the --subnet-id flag", *instance.Reservations[0].Instances[0].InstanceId)
+		}
 
 		e.log.Info(ctx, "detected BYOVPC Hypershift cluster, using the subnets of the worker nodes(private): %s", *instance.Reservations[0].Instances[0].NetworkInterfaces[0].SubnetId)
 		return *instance.Reservations[0].Instances[0].NetworkInterfaces[0].SubnetId, nil
--- a/cmd/network/verification.go
+++ b/cmd/network/verification.go
@@ -302,6 +302,10 @@ func (e *egressVerification) getSubnetId(ctx context.Context) (string, error) {
 					Name:   aws.String("tag:red-hat-managed"),
 					Values: []string{"true"},
 				},
+				{
+					Name:   aws.String("instance-state-name"),
+					Values: []string{"running", "pending"},
+				},
 			},
 		})
 		if err != nil {
@@ -311,7 +315,7 @@ func (e *egressVerification) getSubnetId(ctx context.Context) (string, error) {
 			return "", fmt.Errorf("found 0 instances with kubernetes.io/cluster/%s=owned and red-hat-managed, consider the --subnet-id flag", e.clusterId)
 		}
 		if len(instance.Reservations[0].Instances) == 0 {
-			return "", fmt.Errorf("found 0 instances with kubernetes.io/cluster/%s=owned and red-hat-managed, consider the --subnet-id flag", e.clusterId)
+			return "", fmt.Errorf("found 0 instances with kubernetes.io/cluster/%s=owned and red-hat-managed as running/pending, consider the --subnet-id flag", e.clusterId)
 		}
 		if len(instance.Reservations[0].Instances[0].NetworkInterfaces) == 0 {
 			return "", fmt.Errorf("found 0 network interfaces of the worker node: %s, consider the --subnet-id flag", *instance.Reservations[0].Instances[0].InstanceId)
--- a/cmd/network/verification.go
+++ b/cmd/network/verification.go
@@ -91,6 +91,7 @@ func NewCmdValidateEgress() *cobra.Command {
 }
 
 type egressVerificationAWSClient interface {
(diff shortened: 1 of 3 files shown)


pkg/utils/ocm.go
```diff
@@ -171,13 +178,52 @@ func GetHiveShard(clusterID string) (string, error) {
 		shard = shardPath.Body().HiveConfig().Server()
 	}
 
+	if out.Hypershift().Enabled() {
+		return getHiveURLforHCP(GetCurrentOCMEnv(connection)), nil
```
fahlmant, line 182: Is this meant to get the hive cluster for a management cluster? If so, management clusters are just OSD/ROSA and the hive should be available in OCM.

If this is meant to  get the management cluster for a hypershift cluster, that should also be available in OCM. 

How to get Hive Shard from an OSD/ROSA cluster: 
https://github.com/openshift-online/ocm-cli/blob/main/pkg/cluster/describe.go#L109

How to get Management Cluster for a Hypershift HCP:
https://github.com/openshift-online/ocm-cli/blob/main/pkg/cluster/describe.go#L161-L170

boranx: nice suggestion 👍 , I can try that
the reason I was leaning on hard-coding those was: https://gitlab.cee.redhat.com/service/backplane-cli/-/merge_requests/234/diffs#7fb99b6f8725a63d23914a8668dedf12d1872f30_109_135, I'm not sure if this is fixed though

cmd/network/verification.go
```diff
@@ -290,6 +290,11 @@ func (e *egressVerification) getSubnetId(ctx context.Context) (string, error) {
 		return e.cluster.AWS().SubnetIDs()[0], nil
 	}
 
+	if e.cluster.Hypershift().Enabled() {
+		e.log.Info(ctx, "detected BYOVPC Hypershift cluster, using first subnet from OCM: %s", e.cluster.AWS().SubnetIDs()[0])
+		return e.cluster.AWS().SubnetIDs()[0], nil
```
mjlshen, line 295: HyperShift clusters are always BYOVPC, but they don't have to be PrivateLink, so I don't think we can just grab the first subnet from OCM (since the non-PrivateLink BYOVPC HyperShift clusters will also have public subnets stored in OCM)

pkg/utils/ocm.go
boranx: now that part is added as a check to replace the `clusterID` with corresponding management clustersID

cmd/network/verification.go
boranx: good point! I've updated here to a form where the subnetID is queried from worker nodes of the hcp cluster

boranx:
thanks for the reviews folks @fahlmant @mjlshen, I've improved querying the subnetID and hiveshard parts
I think it's now ready to be re-reviewed

mjlshen (approved):
The logic looks good, I haven't tested it on a live HyperShift cluster though, so LGTM with that caveat

cmd/network/verification.go
```diff
@@ -290,6 +291,36 @@ func (e *egressVerification) getSubnetId(ctx context.Context) (string, error) {
...
+		if len(instance.Reservations) == 0 {
+			return "", fmt.Errorf("found 0 instances with kubernetes.io/cluster/%s=owned and red-hat-managed, consider the --subnet-id flag", e.clusterId)
+		}
+		if len(instance.Reservations[0].Instances) == 0 {
+			return "", fmt.Errorf("found 0 instances with kubernetes.io/cluster/%s=owned and red-hat-managed, consider the --subnet-id flag", e.clusterId)
+		}
+		if len(instance.Reservations[0].Instances[0].NetworkInterfaces) == 0 {
+			return "", fmt.Errorf("found 0 network interfaces of the worker node: %s, consider the --subnet-id flag", *instance.Reservations[0].Instances[0].InstanceId)
+		}
+
```
boranx, line 319: should be a very rare case, but we might also check if there is more than one interface of the instance
like

```go

if len(instance.Reservations[0].Instances. NetworkInterfaces) > 1 {
			return "", fmt.Errorf("more than one interface found on the instance", e.clusterId)
		}
```

boranx:
hey @mjlshen 
may I get another review on that? I've just fixed a case where the subnet is queried from a terminated instance so added that to the filtering
but it ended up losing lgtm label 😿 
https://redhat-internal.slack.com/archives/C04EUL1DRHC/p1675177966513449?thread_ts=1675091330.387229&cid=C04EUL1DRHC

mjlshen:
Since [OSD-15143](https://issues.redhat.com/browse/OSD-15143) is close to finishing I'm thinking about closing this PR in favor of a replacement that utilizes that feature later?

mjlshen:
/close

Let's use https://github.com/openshift/osd-network-verifier/pull/168 in this PR's replacement

PR #317, opened 2023-01-24
Force ack if version can't be determined

fahlmant:
If the latest version can't be determined from GitHub, this change will prompt the user to ensure they want to run the command as a safeguard.


rafael-azevedo (approved):
/lgtm

PR #318, opened 2023-01-24
[OSD-12879] Some minor fixes related to the previous changes

wshearn:
This is to fix some minor errors that came up during the last round of testing for the updated rotate secret command and bash script mostly around the syncset and making sure it has synced.

### DIFF
--- a/cmd/account/rotate-secret.go
+++ b/cmd/account/rotate-secret.go
@@ -15,7 +15,6 @@ import (
 	hiveinternalv1alpha1 "github.com/openshift/hive/apis/hiveinternal/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/utils/pointer"
 
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


cmd/account/rotate-secret.go
```diff
@@ -301,15 +310,14 @@ func (o *rotateSecretOptions) run() error {
 		}
 
 		if isSSSynced {
-			fmt.Printf("Sync completed...")
+			fmt.Println("Sync completed...")
 			break
 		}
 
-		fmt.Printf("Sync not completed, sleeping 5 seconds and rechecking...")
 		time.Sleep(time.Second * 5)
```
iamkirkbater, line 317: Would probably be a good idea to at least keep some kind of output here, just so that it doesn't look like the command is hung.

Even just something as simple as `fmt.Print(".")` would probably suffice.

iamkirkbater (commented):
Not hung up on this change, but it would be a nice usability feature to keep.

cmd/account/rotate-secret.go
wshearn: I have opt'd for adding the printing of a . adding on to the previous print statement.

iamkirkbater (approved):
/lgtm

PR #319, opened 2023-01-27
Version check enhancements

geowa4:
Add the version osdctl thinks it is running and the latest version available to the version check output.


fahlmant (approved):
/approve
/lgtm

PR #320, opened 2023-01-30
Trim v for version check

fahlmant:
`utils.Version` is set when a new release is created via the `.goreleaser.yaml` config. However, the `v` in a version (for example v0.1.35) is getting trimmed. This causes a mismatch between what is set in `utils.Version` and what is being pulled from GitHub.

/hold

I'm not sure this is the way we should go. Ideally we should fix the `.goreleaser.yaml` path to set `utils.Version` correctly. In the case that's not possible, this PR can be merged.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -61,7 +61,7 @@ func NewCmdRoot(streams genericclioptions.IOStreams) *cobra.Command {
 					fmt.Println("Please be aware that you are possibly running an outdated or unreleased version.")
 				}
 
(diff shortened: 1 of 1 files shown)


mjlshen:
I found the docs where GoReleaser mentions stripping the leading "v" (footer 1): https://goreleaser.com/customization/templates/#fn:version-prefix

fahlmant:
That's a bit annoying... if that's what goreleaser is doing, we probably shouldn't fight with it. I think we can merge this PR and call it good.

PR #321, opened 2023-01-31
Use cluster ID for SL search if cluster UUID is not available.

mrbarge:
Addresses a minor nit: If a user of `osdctl servicelog list` provides a cluster ID that has no external cluster ID available, `osdctl`'s query results the list of every service log sent for a cluster that had no external ID (which numbers in the hundreds).

Instead, prefer the cluster UUID, but if it's not available, then alter the SL search query to try with the cluster ID instead.

### DIFF
--- a/cmd/servicelog/list.go
+++ b/cmd/servicelog/list.go
@@ -7,6 +7,7 @@ import (
 	"github.com/openshift-online/ocm-cli/pkg/arguments"
 	"github.com/openshift-online/ocm-cli/pkg/dump"
 	sdk "github.com/openshift-online/ocm-sdk-go"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/openshift/osdctl/pkg/utils"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


cmd/servicelog/list.go
```diff
@@ -90,7 +90,7 @@ func init() {
 	listCmd.Flags().BoolVarP(&serviceLogListInternalOnlyFlag, "internal", "i", serviceLogListInternalOnlyFlag, "Toggle if we should see internal messages")
 }
 
-func CreateListSLRequest(ocmClient *sdk.Connection, clusterId string, allMessages bool, internalMessages bool) *sdk.Request {
+func CreateListSLRequest(ocmClient *sdk.Connection, clusterId string, clusterUuid string, allMessages bool, internalMessages bool) *sdk.Request {
```
mjlshen, line 93: The logic is good, but what do you think about this function having a `cluster *cmv1.Cluster` input instead of `clusterId` and `clusterUuid` individually?

To be fair, `CreateListSLRequest` probably won't be reused much, so I'm happy to get this merged without this change, but I think it would be clearer if we need to consume this function later.

mrbarge: Agreed, that makes sense. That change has been made, thanks!

mjlshen (approved):
/lgtm

fahlmant (approved):
/approve

PR #326, opened 2023-02-06
Fix for the case when error is empty.

Makdaam:
The CPD command shows "SL already sent" even when the error field isn't populated.

This fix removes a confusing output.


fahlmant (approved):
/approve

PR #331, opened 2023-02-10
Add commands to check if cluster owner is a banned user, and to verify that their pull secret matches the cluster's

AlexVulaj:
This PR migrates the functionality of our [check_banned_users](https://github.com/openshift/ops-sop/blob/master/v4/utils/check_banned_user.sh) script into osdctl. Because the original script really performs 2 operations, I have made them into two separate commands here.

The first command, `cluster check-banned-user`, will check whether or not the owner of a cluster is a banned user. If so, it will provide next-steps to follow.

The second command, `cluster validate-pull-secret`, will check if the cluster's pull secret email address matches the email address of the owner account. If not, it will provide next-steps to follow.

I'm open to renaming these commands, but wanted to get the functionality and logic up and reviewed. Renaming them would be trivial if desired.

### DIFF
--- a/cmd/cluster/checkbanneduser.go
+++ b/cmd/cluster/checkbanneduser.go
@@ -1,0 +1,72 @@
+package cluster
+
+import (
+	"fmt"
+	"github.com/openshift/osdctl/pkg/utils"
+	"github.com/spf13/cobra"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+const BanCodeExportControlCompliance = "export_control_compliance"
+
+func newCmdCheckBannedUser() *cobra.Command {
+	return &cobra.Command{
+		Use:               "check-banned-user [CLUSTER_ID]",
+		Short:             "Checks if the cluster owner is a banned user.",
+		Args:              cobra.ExactArgs(1),
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(CheckBannedUser(args[0]))
+		},
+	}
+}
+
+func CheckBannedUser(clusterID string) error {
+	ocm := utils.CreateConnection()
+	defer func() {
+		if ocmCloseErr := ocm.Close(); ocmCloseErr != nil {
+			fmt.Printf("Cannot close the ocm (possible memory leak): %q", ocmCloseErr)
+		}
+	}()
+
+	fmt.Print("Finding subscription account: ")
+	subscription, err := utils.GetSubscription(ocm, clusterID)
+	if err != nil {
+		return err
+	}
+
+	if status := subscription.Status(); status != "Active" {
+		return fmt.Errorf("Expecting status 'Active' found %v\n", status)
+	}
+
+	fmt.Printf("Account %v - %v - %v\n", subscription.SupportLevel(), subscription.Creator().HREF(), subscription.Status())
+
+	fmt.Print("Finding account owner: ")
+	creator, err := utils.GetAccount(ocm, subscription.Creator().ID())
+	if err != nil {
+		return err
+	}
+
+	userEmail := creator.Email()
+	userBanned := creator.Banned()
+	userBanCode := creator.BanCode()
+	userBanDescription := creator.BanDescription()
+	lastUpdate := creator.UpdatedAt()
+
+	fmt.Printf("%v\n-------------------\nLast Update : %v\n", userEmail, lastUpdate)
+
+	if userBanned {
+		fmt.Println("User is banned")
+		fmt.Printf("Ban code = %v\n", userBanCode)
+		fmt.Printf("Ban description = %v\n", userBanDescription)
+		if userBanCode == BanCodeExportControlCompliance {
+			fmt.Println("User banned due to export control compliance.\nPlease follow the steps detailed here: https://github.com/openshift/ops-sop/blob/master/v4/alerts/UpgradeConfigSyncFailureOver4HrSRE.md#user-banneddisabled-due-to-export-control-compliance .")
+			return nil
+		}
+		fmt.Println("Recommend sending service log with:")
+		fmt.Printf("osdctl servicelog post -t https://raw.githubusercontent.com/openshift/managed-notifications/master/ocm/cluster_owner_disabled.json %v\n", clusterID)
+		return nil
+	}
+	fmt.Println("User allowed")
+	return nil
+}
--- a/cmd/cluster/validatepullsecret.go
+++ b/cmd/cluster/validatepullsecret.go
@@ -1,0 +1,106 @@
+package cluster
+
+import (
+	"context"
+	"fmt"
+	v1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
+	"github.com/openshift/osdctl/pkg/utils"
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var BackplaneClusterAdmin = "backplane-cluster-admin"
+
+func newCmdValidatePullSecret(kubeCli client.Client, flags *genericclioptions.ConfigFlags) *cobra.Command {
+	return &cobra.Command{
+		Use:   "validate-pull-secret [CLUSTER_ID]",
+		Short: "Checks if the pull secret email matches the owner email",
+		Long: `Checks if the pull secret email matches the owner email.
+
+The owner's email to check will be determined by the cluster identifier passed to the command, while the pull secret checked will be determined by the cluster that the caller is currently logged in to.`,
+		Args:              cobra.ExactArgs(1),
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(ValidatePullSecret(args[0], kubeCli, flags))
+		},
+	}
+}
+
+func ValidatePullSecret(clusterID string, kubeCli client.Client, flags *genericclioptions.ConfigFlags) error {
+	ocm := utils.CreateConnection()
+	defer func() {
+		if ocmCloseErr := ocm.Close(); ocmCloseErr != nil {
+			fmt.Printf("Cannot close the ocm (possible memory leak): %q", ocmCloseErr)
+		}
+	}()
+
+	fmt.Println("Checking if pull secret email matches user email")
+
+	// This is the flagset for the kubeCli object provided from the root command. Set here to retroactively impersonate backplane-cluster-admin
+	flags.Impersonate = &BackplaneClusterAdmin
+	secret := &corev1.Secret{}
+	err := kubeCli.Get(context.TODO(), types.NamespacedName{Namespace: "openshift-config", Name: "pull-secret"}, secret)
+	if err != nil {
+		return err
+	}
+
+	clusterPullSecretEmail, err, done := getPullSecretEmail(clusterID, secret)
+	if done {
+		return err
+	}
+
+	subscription, err := utils.GetSubscription(ocm, clusterID)
+	if err != nil {
+		return err
+	}
+
+	account, err := utils.GetAccount(ocm, subscription.Creator().ID())
+	if err != nil {
+		return err
+	}
+
+	if account.Email() != clusterPullSecretEmail {
+		fmt.Println("Pull secret email doesn't match OCM user email. Recommend sending service log with:")
+		fmt.Printf("osdctl servicelog post -t https://raw.githubusercontent.com/openshift/managed-notifications/master/osd/pull_secret_user_mismatch.json %v\n", clusterID)
+		return nil
+	}
+
+	fmt.Println("Email addresses match.")
+	return nil
+}
+
+func getPullSecretEmail(clusterID string, secret *corev1.Secret) (string, error, bool) {
+	dockerConfigJsonBytes, found := secret.Data[".dockerconfigjson"]
+	if !found {
+		// Indicates issue w/ pull-secret, so we can stop evaluating and specify a more direct course of action
+		fmt.Println("Secret does not contain expected key '.dockerconfigjson'. Recommend sending a service log with the following command:")
+		fmt.Printf("osdctl servicelog post -t https://raw.githubusercontent.com/openshift/managed-notifications/master/osd/pull_secret_change_breaking_upgradesync.json %v\n", clusterID)
+		return "", nil, true
+	}
+
+	dockerConfigJson, err := v1.UnmarshalAccessToken(dockerConfigJsonBytes)
+	if err != nil {
+		return "", err, true
+	}
+
+	cloudOpenshiftAuth, found := dockerConfigJson.Auths()["cloud.openshift.com"]
+	if !found {
+		fmt.Println("Secret does not contain entry for cloud.openshift.com. Recommend sending a service log with the following command:")
+		fmt.Printf("osdctl servicelog post -t https://raw.githubusercontent.com/openshift/managed-notifications/master/osd/pull_secret_change_breaking_upgradesync.json %v\n", clusterID)
+		return "", nil, true
+	}
+
+	clusterPullSecretEmail := cloudOpenshiftAuth.Email()
+	if clusterPullSecretEmail == "" {
+		fmt.Printf("%v\n%v\n%v\n",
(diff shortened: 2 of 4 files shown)


clcollins:
+1 for this effort, but I'd prefer the command names to be `cluster {verb} {noun}` format, so their functions are clearer at a glance.

Will review more in depth this afternoon!

cmd/cluster/pullsecret.go
```diff
@@ -0,0 +1,90 @@
...
+	ocm := utils.CreateConnection()
+	defer func() {
+		if ocmCloseErr := ocm.Close(); ocmCloseErr != nil {
+			fmt.Printf("Cannot close the ocm (possible memory leak): %q", ocmCloseErr)
+		}
+	}()
+
+	fmt.Println("Checking if pull secret email matches user email")
+
+	flags.Impersonate = pointer.StringPtr("backplane-cluster-admin")
```
tnierman, line 39: Correct me if I'm wrong, but I don't believe this ever used? If not, can we remove it?

tnierman: I stand corrected: it looks like this is the same flagset used by the kubeCli client.Client object [provided by the root cmd](https://github.com/openshift/osdctl/blob/master/cmd/cmd.go#L88). So we're retroactively impersonating backplane-cluster-admin after the client has been built, but, in my opinion, it's not immediately evident that's the case.

Could we maybe add a comment indicating as much for future reference?

cmd/cluster/pullsecret.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return err
+	}
+
+	clusterPullSecretEmail := dockerConfigJson.Auths.Cloudopenshiftcom.Email
+	if clusterPullSecretEmail == "" {
+		fmt.Printf("%v\n%v\n%v\n",
+			"Couldn't extract email address from pull secret for cloud.openshift.com",
+			"This can mean the pull secret is misconfigured. Please verify the pull secret manually:",
+			"  oc get secret -n openshift-config pull-secret -o json | jq -r '.data.\".dockerconfigjson\"' | base64 -d")
```
tnierman, line 69: This command doesn't work for me with `jq-1.6` on Fedora.

If the following also works on Mac, can we update this suggested command to:
```
oc get secret -n openshift-config pull-secret -o json | jq -r '.data[".dockerconfigjson"]' | base64 -d
```

cmd/cluster/pullsecret.go
tnierman: It also appears that the `pointer.StringPtr` function is [deprecated](https://pkg.go.dev/k8s.io/utils/pointer#StringPtr).

AlexVulaj: I was typing a response to your first query but I see you've figured it out 😄 Yes, that flag is necessary to impersonate backplane-cluster-admin for our use cases here.

As for `StringPtr` being deprecated - good catch! I had copied this from `/cmd/account/generate-secret.go` and never went back to clean up. I'll do that now.

cmd/cluster/pullsecret.go
AlexVulaj: For what it's worth, these outputs were copied [directly from the old check_banned_user script](https://github.com/openshift/ops-sop/blob/master/v4/utils/check_banned_user.sh#L70).

The command you shared also works on my mac, so I'll update to use that instead. Thanks for checking this!

cmd/cluster/checkbanneduser.go
```diff
@@ -0,0 +1,69 @@
...
+
+const BanCodeExportControlCompliance = "export_control_compliance"
+
+func newCmdBannedUser() *cobra.Command {
+	return &cobra.Command{
+		Use:               "check-banned-user",
+		Short:             "Checks if the cluster owner is a banned user.",
+		Args:              cobra.ExactArgs(1),
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
```
clcollins, line 18: Should this be `RunE:`?  

`func CheckBannedUser` returns an error, but `Run:` just drops it on the floor, so there's no indication if there was an error or not.

cmd/cluster/validatepullsecret.go
```diff
@@ -0,0 +1,92 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var BackplaneClusterAdmin = "backplane-cluster-admin"
+
+func newCmdPullSecret(kubeCli client.Client, flags *genericclioptions.ConfigFlags) *cobra.Command {
```
clcollins, line 18: Dumb little request, but could `newCmdPullSecret` (and `newCmdBannedUser`) be updated to match the changes to the actual command (eg `newCmdValidatePullSecret`)?

Super minor prio.  Just nagging at me.

cmd/cluster/validatepullsecret.go
```diff
@@ -0,0 +1,92 @@
...
+	return &cobra.Command{
+		Use:               "validate-pull-secret",
+		Short:             "Checks if the pull secret email matches the owner email",
+		Args:              cobra.ExactArgs(1),
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
```
clcollins, line 24: Same as the other command: `Run` doesn't do anything with the error returned by ValidatePullSecret.

cmd/cluster/validatepullsecret.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return err
+	}
+
+	clusterPullSecretEmail := dockerConfigJson.Auths.Cloudopenshiftcom.Email
+	if clusterPullSecretEmail == "" {
+		fmt.Printf("%v\n%v\n%v\n",
+			"Couldn't extract email address from pull secret for cloud.openshift.com",
+			"This can mean the pull secret is misconfigured. Please verify the pull secret manually:",
+			"  oc get secret -n openshift-config pull-secret -o json | jq -r '.data[\".dockerconfigjson\"]' | base64 -d")
```
clcollins, line 71: Maybe we could handle this manual validation with `osdctl`?  We have the cluster id, `osdctl` can login to clusters.  Could we just have it do so here?

Not a blocker for me - maybe worth looking at in the future.

cmd/cluster/pullsecret.go
tnierman: Yeah, sorry about the initial confusion. Took my brain a minute to recall how the k8s client works :slightly_smiling_face:

cmd/cluster/checkbanneduser.go
AlexVulaj: I'm not sure honestly. All over the rest of the code base, I see `Run` being used with `cmdutil.CheckErr` wrapping the handler function. Tracing down, `cmdutil.CheckErr` will eventually call the following function, which seems to write to stderr and exit with a non-0 code.

```golang
func fatal(msg string, code int) {
	// nolint:logcheck // Not using the result of klog.V(99) inside the if
	// branch is okay, we just use it to determine how to terminate.
	if klog.V(99).Enabled() {
		klog.FatalDepth(2, msg)
	}
	if len(msg) > 0 {
		// add newline if needed
		if !strings.HasSuffix(msg, "\n") {
			msg += "\n"
		}
		fmt.Fprint(os.Stderr, msg)
	}
	os.Exit(code)
}
```

cmd/cluster/validatepullsecret.go
AlexVulaj: Answered above - I think we're okay here.

cmd/cluster/validatepullsecret.go
AlexVulaj: Could you explain a bit more about what you'd like to see here? What we're doing above in code is the equivalent of what's written here in this output, followed up by trying to find the email address in `.dockerconfigjson` which fails. At that point I'm not sure how else we could programmatically recover.

cmd/cluster/checkbanneduser.go
clcollins: Ah, good catch.  Carry on.  Nothing to see here.

cmd/cluster/validatepullsecret.go
clcollins: Just ignore me today, I guess.  :facepalm:  Apparently my brain is not here at all.

AlexVulaj:
/hold
I want to spend a little bit more time seeing if I can get the `v1.AccessToken` struct to play nicely with my unmarshalling instead of using the custom structs. It's a bit of a pain since that struct has unexported fields.

cmd/cluster/validatepullsecret.go
```diff
@@ -0,0 +1,92 @@
...
+	type auths struct {
+		Cloudopenshiftcom auth `json:"cloud.openshift.com"`
+	}
+
+	type response struct {
+		Auths auths `json:"auths"`
+	}
+
+	dockerConfigJson := response{}
+	err = json.Unmarshal(secret.Data[".dockerconfigjson"], &dockerConfigJson)
```
tnierman, line 61: I think we should lookup the value for this key first, to check that it's populated, before attempting to `Unmarshal`:

```
        dockerconfigjsonBytes, found := secret.Data[".dockerconfigjson"]
        if !found {
            // Indicates issue w/ pull-secret, so we can stop evaluating and specify a more direct course of action
            return fmt.Errorf("secret does not contain expected key '.dockerconfigjson'. Do this thing: <link to SOP>")
        }
	err = json.Unmarshal(secret.Data[".dockerconfigjson"], &dockerConfigJson)
```

AlexVulaj: Today I learned! Thanks for sharing this, I totally agree it's a more robust approach.

AlexVulaj: Does anyone have opinions on what next steps should be here? I can't find an sop  that seems relevant for handling this situation. Perhaps one of the following service log templates?

https://github.com/openshift/managed-notifications/blob/master/osd/update_pull_secret.json

https://github.com/openshift/managed-notifications/blob/master/osd/pull_secret_change_breaking_upgradesync.json

AlexVulaj: I've decided to go with the "update_pull_secret" SL. resolving this thread.

cmd/cluster/validatepullsecret.go
```diff
@@ -0,0 +1,100 @@
...
+
+	type auths struct {
+		Cloudopenshiftcom auth `json:"cloud.openshift.com"`
+	}
+
+	type response struct {
+		Auths auths `json:"auths"`
+	}
+
+	dockerConfigJsonBytes, found := secret.Data[".dockerconfigjson"]
```
fahlmant, line 60: I still there there's likely somewhere we can pull this definition from. I'm going to take a look and see if I can find it. 

If we can't find it, then I propose we should move this out of the function and at least define these at the package level, but potentially at another level. That way, any other script or logic needed to access this same json can use that definition, rather than redefining it each time we need it

AlexVulaj: @fahlmant the struct we want is https://github.com/openshift-online/ocm-sdk-go/blob/main/accountsmgmt/v1/access_token_type.go#L23

However, none of the fields are exported so I wasn't able to unmarshal a response to that struct. I agree that if there's a way we should be using it, and I'm open to input.

cmd/cluster/checkbanneduser.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return err
+	}
+
+	if status := subscription.Status(); status != "Active" {
+		return fmt.Errorf("Expecting status 'Active' found %v\n", status)
+	}
+
+	fmt.Println("Finding subscription account...")
+	creator, err := utils.GetAccount(ocm, subscription.Creator().ID())
```
mjlshen, line 43:
```suggestion
	creator, err := utils.GetAccount(ocm, subscription.Creator().ID())
  if err != nil {
		return err
	}
```

cmd/cluster/checkbanneduser.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/spf13/cobra"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+const BanCodeExportControlCompliance = "export_control_compliance"
+
+func newCmdCheckBannedUser() *cobra.Command {
+	return &cobra.Command{
+		Use:               "check-banned-user",
+		Short:             "Checks if the cluster owner is a banned user.",
```
mjlshen, line 15: We should add documentation like https://github.com/openshift/osdctl/blob/master/cmd/network/verification.go#L57-L83

cmd/cluster/validatepullsecret.go
```diff
@@ -0,0 +1,100 @@
...
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var BackplaneClusterAdmin = "backplane-cluster-admin"
+
+func newCmdValidatePullSecret(kubeCli client.Client, flags *genericclioptions.ConfigFlags) *cobra.Command {
+	return &cobra.Command{
+		Use:               "validate-pull-secret",
+		Short:             "Checks if the pull secret email matches the owner email",
```
mjlshen, line 21: Same docs callout

mjlshen (commented):
I'm a +1 on the intent here, but I think the value of porting this into Go from the original bash script is if we can unit test this. Otherwise if there is a bug in any of these flows an SRE will need to manually piece together the corresponding ocm/oc commands to do this (much more difficult compared to reading the existing bash script) and we would lose a lot of troubleshooting ability.

With unit tests I'd be comfortable deleting the old bash script, otherwise I like this as an alternative option in the meantime.

cmd/cluster/validatepullsecret.go
fahlmant: It looks like we might still be able to use the SDK functions to Unmarshal and access fields
https://github.com/openshift-online/ocm-sdk-go/blob/main/accountsmgmt/v1/access_token_auth_type_json.go#L67
https://github.com/openshift-online/ocm-sdk-go/blob/main/accountsmgmt/v1/access_token_auth_type.go#L55

AlexVulaj: Good find here! I think https://github.com/openshift-online/ocm-sdk-go/blob/main/accountsmgmt/v1/access_token_type_json.go#L79 will get us what we need here.

cmd/cluster/checkbanneduser.go
AlexVulaj: Good catch here - this is a miss by me.

cmd/cluster/validatepullsecret.go
```diff
@@ -0,0 +1,89 @@
...
+
+	dockerConfigJsonBytes, found := secret.Data[".dockerconfigjson"]
+	if !found {
+		// Indicates issue w/ pull-secret, so we can stop evaluating and specify a more direct course of action
+		fmt.Println("Secret does not contain expected key '.dockerconfigjson'. Recommend sending a service log with the following command:")
+		fmt.Printf("osdctl servicelog post -t https://raw.githubusercontent.com/openshift/managed-notifications/master/osd/pull_secret_change_breaking_upgradesync.json %v\n", clusterID)
+		return nil
+	}
+
+	dockerConfigJson, err := v1.UnmarshalAccessToken(dockerConfigJsonBytes)
```
tnierman, line 57: Value of `err` isn't checked here

AlexVulaj: This is actually a result of me forgetting to delete the line below this which is no longer needed. Taking care of that now.

cmd/cluster/checkbanneduser.go
AlexVulaj: To loop in our slack conversation - I've added `[CLUSTER_ID]` to the usage here.

cmd/cluster/validatepullsecret.go
AlexVulaj: per our slack conversation, I've updated the usage and added a Long property here.

AlexVulaj:
@mjlshen I don't mind adding tests, and I agree they'd be valuable here. I'll take care of that as part of this PR.

AlexVulaj:
Added tests that satisfy new logic after chatting with @mjlshen . All should be good now.

mjlshen:
/approve

let's iterate

AlexVulaj:
the `ci/prow/build` check has been stuck for close to 4 hours now and `/retest` doesn't seem to have done anything. Does anyone know how we might be able to force a restart of that action?

fahlmant (approved):
/lgtm

PR #333, opened 2023-02-14
[OSD-15156] Add gcp support to cluster health

bergmannf:
Adds GCP support to the `osdctl cluster health` command.

### DIFF
--- a/cmd/cluster/health.go
+++ b/cmd/cluster/health.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"strconv"
@@ -11,6 +12,7 @@ import (
 	"github.com/openshift/osdctl/pkg/osdCloud"
 	"github.com/openshift/osdctl/pkg/utils"
 	"github.com/spf13/cobra"
+	"google.golang.org/api/iterator"
 	"gopkg.in/yaml.v2"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 )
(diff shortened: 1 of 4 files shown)


cmd/cluster/health.go
```diff
@@ -96,59 +98,120 @@ func (o *healthOptions) run() error {
...
-						if *inst.State.Name == "stopped" {
-							totalStopped += 1
-						}
-
+	if cluster.CloudProvider().ID() == "gcp" {
+		clusterResources, err := ocmClient.ClustersMgmt().V1().Clusters().Cluster(o.clusterID).Resources().Live().Get().Send()
+		if err != nil {
+			return err
+		}
+		projectClaimRaw := clusterResources.Body().Resources()["gcp_project_claim"]
```
AlexVulaj, line 112: is "gcp_project_claim" guaranteed to be present here? I'm wondering if it could be helpful to add a check here, like

```go
projectClaimRaw, ok := clusterResources.Body().Resources()["gcp_project_claim"]
if !ok { ... }
```
```

File: cmd/cluster/health.go [hunk 2]
```diff
@@ -96,59 +98,120 @@ func (o *healthOptions) run() error {
...
-					} else if strings.HasPrefix(*t.Value, cluster.InfraID()) && strings.Contains(*t.Value, "worker") {
-						totalCluster += 1
-						if *inst.State.Name == "running" {
-							runningWorkers += 1
-						}
-						if *inst.State.Name == "stopped" {
-							totalStopped += 1
-						}
-
+	if cluster.CloudProvider().ID() == "gcp" {
```
AlexVulaj, line 107: For the sake of organization, do you think it makes sense to extract the different trees of this `if` (aws, gcp, etc) to different functions?

bergmannf: I have a follow up PR that gets rid of most of the if/else stuff by extracting it into polymorphism.

IMO having a 'healthCheckAWS()' and 'healthCheckGCP()' doesn't make it much better, but I can move it out if you feel it does.

AlexVulaj: I could go either way. I'll wait to see how the follow up PR looks.

PR #335, opened 2023-02-15
Working minimum POC that calls other commands

iamkirkbater:
This is a demo of a working minimum POC that allows us to call other commands (and subsequently mock them out) by adhering to an interface.

Old commands would need to be retrofit into the new paradigm, so I figured I'd start with a quick little PoC.

It's not _terribly_ different from what we're doing, but we'd want to standardize on it and so I figured I'd open the floor for feedback as early as possible on this.

One thing I would do is create some kind of template file that we could initialize new command files with, so that new commands all use the same template.

### DIFF
--- a/cmd/cluster/jumphost/init.go
+++ b/cmd/cluster/jumphost/init.go
@@ -1,0 +1,72 @@
+package jumphost
+
+import (
+	"fmt"
+
+	osdctlCommand "github.com/openshift/osdctl/pkg/osdctlCommand"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+// newCmdInit builds a new Jumphost from start to finish
+func newCmdInit(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
+	ops := &InitOptions{}
+
+	initCmd := &cobra.Command{
+		Use:               "init",
+		Short:             "Initializes a jumphost in a cluster",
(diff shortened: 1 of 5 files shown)


cmd/cluster/jumphost/init.go
```diff
@@ -0,0 +1,72 @@
...
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(ops.Run())
+		},
+	}
+
+	initCmd.Flags().StringVarP(&ops.keyName, "key-name", "k", "$$USER$$-SRE-Jumphost", "Key Name")
+	return initCmd
+}
+
+type InitOptions struct {
```
iamkirkbater, line 30: I'm not terribly sold on keeping the name here `[command]Options`, I'm almost wondering if `[command]Command` might be a better name for it?  So in the case of the key generation command it would be `KeyGenerateCommand`, and would be used within this command as `o.GenerateKeyCommand`, which makes it a bit more clear what it is.  Thoughts?

cmd/cluster/jumphost/init.go
```diff
@@ -0,0 +1,72 @@
...
+	keyName string
+
+	generateKeyOpts osdctlCommand.Command
+}
+
+func (o *InitOptions) Init() error {
```
iamkirkbater, line 36: Also not sold on `Init` here, though I'm not sure what else to call it.

cmd/cluster/jumphost/init.go
iamkirkbater: Potential Idea has been thrown into the `GenerateKey` command above.

iamkirkbater:
Closed because PoC has been completed.

PR #337, opened 2023-02-16
Remove version check for upgrade command

AlexVulaj:
It seems weird to prompt users with a version check when they're trying to upgrade. The likely case is that they know they're out of date and that's why they're trying to upgrade in the first place.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"k8s.io/kubectl/pkg/util/slice"
 	"os"
 	"strings"
 
@@ -53,7 +54,7 @@ func NewCmdRoot(streams genericclioptions.IOStreams) *cobra.Command {
 				os.Exit(1)
 			}
 
-			if !skipVersionCheck {
+			if !skipVersionCheck && !slice.ContainsString(getSkipVersionCommands(), cmd.Use, nil) {
 				latestVersion, err := utils.GetLatestVersion()
 				if err != nil {
 					fmt.Println("Warning: Unable to verify that osdctl is running under the latest released version. Error trying to reach GitHub:")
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -53,7 +53,7 @@ func NewCmdRoot(streams genericclioptions.IOStreams) *cobra.Command {
 				os.Exit(1)
 			}
 
-			if !skipVersionCheck {
+			if !skipVersionCheck || cmd.Use == "upgrade" {
```
iamkirkbater, line 56: Maybe also `|| cmd.Use == "version"` as well?  I think version already tells you if you're not on the latest as part of the output.

fahlmant: Is it possible (and does it make sense) to make this a constant rather than an anonymous string?

fahlmant: SGTM

AlexVulaj: Also just now realizing this should be "&&" but we'll ignore that :)

cmd/cmd.go
```diff
@@ -53,7 +54,7 @@ func NewCmdRoot(streams genericclioptions.IOStreams) *cobra.Command {
 				os.Exit(1)
 			}
 
-			if !skipVersionCheck {
+			if !skipVersionCheck && !slice.ContainsString(getSkipVersionCommands(), cmd.Use, nil) {
```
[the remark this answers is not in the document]
AlexVulaj: I'm fine with either approach. My goal was to keep this change as simple as possible, doing as little refactoring as needed.

clcollins:
/hold

I think I'm more in favor of #339 for readability.  I placed both into `/hold` for further discussion from other interested parties, however.

mjlshen:
/close

Merged #339 instead

PR #338, opened 2023-02-16
Refactor servicelog post such that it can be called from other commands

AlexVulaj:
In some instances, the end result of an investigation performed by an osdctl command is to send a service log to the customer. This process is much more easily facilitated if the original command called automatically calls through to the `servicelog post` command. Two recent commands use this functionality (check-banned-user and validate-pull-secret).

The user will still be prompted to confirm before the service log is actually sent.

NOTE: this PR loosely attempts to follow the interface proposed in [336](https://github.com/openshift/osdctl/pull/336)

### DIFF
--- a/cmd/cluster/checkbanneduser.go
+++ b/cmd/cluster/checkbanneduser.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"fmt"
+	"github.com/openshift/osdctl/cmd/servicelog"
 	"github.com/openshift/osdctl/pkg/utils"
 	"github.com/spf13/cobra"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
@@ -63,8 +64,16 @@ func CheckBannedUser(clusterID string) error {
 			fmt.Println("User banned due to export control compliance.\nPlease follow the steps detailed here: https://github.com/openshift/ops-sop/blob/master/v4/alerts/UpgradeConfigSyncFailureOver4HrSRE.md#user-banneddisabled-due-to-export-control-compliance .")
 			return nil
 		}
-		fmt.Println("Recommend sending service log with:")
-		fmt.Printf("osdctl servicelog post -t https://raw.githubusercontent.com/openshift/managed-notifications/master/ocm/cluster_owner_disabled.json %v\n", clusterID)
+
+		fmt.Println("Sending service log.")
+		postCmd := servicelog.PostCmdOptions{
+			Template:  "https://raw.githubusercontent.com/openshift/managed-notifications/master/ocm/cluster_owner_disabled.json",
+			ClusterId: clusterID,
+		}
+		if err = postCmd.Run(); err != nil {
+			return err
+		}
+
 		return nil
 	}
 	fmt.Println("User allowed")
--- a/cmd/servicelog/post.go
+++ b/cmd/servicelog/post.go
@@ -3,6 +3,8 @@ package servicelog
 import (
 	"encoding/json"
 	"fmt"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/utils/strings/slices"
 	"net/url"
 	"os"
 	"os/signal"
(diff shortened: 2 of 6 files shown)


cmd/cluster/checkbanneduser.go
```diff
@@ -63,8 +64,16 @@ func CheckBannedUser(clusterID string) error {
 			fmt.Println("User banned due to export control compliance.\nPlease follow the steps detailed here: https://github.com/openshift/ops-sop/blob/master/v4/alerts/UpgradeConfigSyncFailureOver4HrSRE.md#user-banneddisabled-due-to-export-control-compliance .")
 			return nil
 		}
-		fmt.Println("Recommend sending service log with:")
-		fmt.Printf("osdctl servicelog post -t https://raw.githubusercontent.com/openshift/managed-notifications/master/ocm/cluster_owner_disabled.json %v\n", clusterID)
+
+		fmt.Println("Sending service log.")
```
typeid, line 68: I'm not completely sure we would want the service logs to be sent fully automatically. We don't do that in any of our automation scripts AFAIK. 
Maybe adding a y/n confirmation to it would be better?

AlexVulaj: As stated in the PR description, there will still be a prompt to send the service log. It will function exactly as if you called `osdctl servicelog post...` yourself.

cmd/servicelog/post.go
```diff
@@ -24,175 +26,208 @@ import (
...
-	postCmd.Flags().BoolVarP(&isDryRun, "dry-run", "d", false, "Dry-run - print the service log about to be sent but don't send it.")
-	postCmd.Flags().StringArrayVarP(&filterParams, "query", "q", filterParams, "Specify a search query (eg. -q \"name like foo\") for a bulk-post to matching clusters.")
-	postCmd.Flags().BoolVarP(&skipPrompts, "yes", "y", false, "Skips all prompts.")
-	postCmd.Flags().StringArrayVarP(&filterFiles, "query-file", "f", filterFiles, "File containing search queries to apply. All lines in the file will be concatenated into a single query. If this flag is called multiple times, every file's search query will be combined with logical AND.")
-	postCmd.Flags().StringVarP(&clustersFile, "clusters-file", "c", clustersFile, `Read a list of clusters to post the servicelog to. the format of the file is: {"clusters":["$CLUSTERID"]}`)
-	postCmd.Flags().BoolVarP(&internalOnly, "internal", "i", false, "Internal only service log. Use MESSAGE for template parameter (eg. -p MESSAGE='My super secret message').")
+func (o *PostCmdOptions) check(response *sdk.Response, clusterMessage servicelog.Message) {
+	body := response.Bytes()
+	if response.Status() < 400 {
```
typeid, line 212: **thought:** A bit unrelated to this PR, but it seems odd to consider anything with code < 400 as "good response".

cmd/cluster/checkbanneduser.go
typeid: Ah, makes sense! Sorry for not reading ;)

cmd/servicelog/post.go
AlexVulaj: Yeah it's definitely a weird way to confirm a "good response". That being said, I don't want to touch anything unnecessary in this command given how important its use is 😅

cmd/servicelog/post.go
```diff
@@ -24,175 +26,208 @@ import (
...
 	ClustersFile    servicelog.ClustersFile
-	template        string
+	Template        string
 	filterFiles     []string // Path to filter file
 	filtersFromFile string   // Contents of filterFiles
 	isDryRun        bool
 	skipPrompts     bool
 	clustersFile    string
 	internalOnly    bool
+	Args            []string
```
mjlshen, line 39: I'm -1 to this. If we're going the `Init()` approach it makes more sense to me for it to be something explicit like this:

```go
func (p *PostCmdOptions) Init(clusterId, template string) error {
  p.template = template
  p.clusterId = clusterId
  // ...etc

  return nil
}

File: cmd/servicelog/post.go [hunk 4]
```diff
@@ -24,175 +26,208 @@ import (
...
+			opts.Args = args
+			cmdutil.CheckErr(opts.Init())
+			cmdutil.CheckErr(opts.Validate())
+			cmdutil.CheckErr(opts.Run())
+		},
+	}
+
+	// define required flags
+	postCmd.Flags().StringVarP(&opts.Template, "template", "t", "", "Message template file or URL")
+	postCmd.Flags().StringArrayVarP(&templateParams, "param", "p", templateParams, "Specify a key-value pair (eg. -p FOO=BAR) to set/override a parameter value in the template.")
```
mjlshen, line 65: How would a user send a servicelog that required parameters to be specified?

cmd/servicelog/post.go
iamkirkbater: The issue with that approach is that we can't hold things to an interface if we accept random parameters for each command.

 Really, the only point of the `Init()` method is to be able to instantiate default sdks that we might want to mock.  
 
 So if a command needs both the ocm and aws clients, the Init would look something like this:
 
 ```
 func (p *PostCmdOptions) Init() error {
   if p.ocmClient == nil {
     p.ocmClient = ocm.newOcmClient()
   }
   
   if p.awsClient == nil {
     p.awsClient = aws.newAwsClient()
   }
}
```

Ideally so that when you want to run this command from a unit test you would do something like
```
postCmd := newPostCmdOptions()
postCmd.awsClient = mockAwsClient.EXPECT()...
postCmd.ocmClient = mockOcmClient.EXPECT()...

postCmd.Run()
```

So the point of `Init` is to be able to ensure that when we run the commands via the CLI invocation that we have the actual OCM/AWS clients vs when we're testing we can push this in.

I'm not terribly convinced this is the greatest name, so if you have any other ideas around the convention to use here then I'm all ears.

File: cmd/servicelog/post.go (hunk 4 above)
Comment (AlexVulaj) [reply cmd/servicelog/post.go:65]: Good catch on this.

There's a "templateParams" global variable in `/cmd/servicelog/common.go` that doesn't really need to be common. I can move that into the "post" command's struct and make it an extracted property so callers can set it.

File: cmd/servicelog/post.go (hunk 3 above)
Comment (AlexVulaj) [reply cmd/servicelog/post.go:39]: A part of me wants to rewrite this entire command to not use an interface at all, but be more in line with what we did for `checkbanneduser.go` and `validatepullsecret.go`. For external context, the functionality executed by those commands is in public functions that are not tied to any struct. Their dependencies are defined as function parameters, and they can be called from anywhere.

Do you think there's value in writing up an alternative PR proposal with that approach to compare how it looks?

Comment (mjlshen) [reply cmd/servicelog/post.go:39]: Ah I brought it up mostly because it seems like a bit of a gap in this specific use-case, I can see Kirk's point where `Init()` may be useful in other situations. My suggestion doesn't need to overwrite the intended use-case for `Init()`, but the core part of my gripe is I think we should have a clearer way of setting up a `PostCmdOptions` struct than making the caller pass a cluster id via the args field.

Comment (AlexVulaj) [reply cmd/servicelog/post.go:39]: The "args" can only ever be of length 0 or 1. How about instead of having the "Args" field at all we just make it `clusterId string` and it can either be empty or have a value? I think I'm going to go with that approach unless there are objections.

Comment (mjlshen) [reply cmd/servicelog/post.go]: Agreed, because we're essentially not caring about the additional functionality of the servicelog post command with regards to `filterParams` and `clustersFile`, so we can make a clear interface for sending servicelogs to one cluster id.

(i.e. this check in the code)
```go
if o.ClusterId == "" && len(filterParams) == 0 && o.clustersFile == ""
```

AlexVulaj: I'm going to throw "Validate" and "init" calls at the start of "Run" for now to make sure they _can't_ get skipped over. They're idempotent functions and very lightweight so calling them multiple times shouldn't cause any harm.

mjlshen (approved):
/lgtm

PR #339, opened 2023-02-17
Refactor version check to allow some commands to skip it

fahlmant:
Alternative proposal to #337 

This PR:
- Moves the logic for the actual version check to a separate function. This allows for cleaner reading of the control logic
- Creates a function that determines if the version check should be run
- Creates a function that generates an allowlist of commands that can skip the version check
- Creates a function that checks if the provided command is in the allowlist
- Abstracts the implementation of each step behind a function that makes for easier reading and separates the logic into parts that are easier to reason about

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -13,6 +13,7 @@ import (
 
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/kubectl/pkg/util/slice"
 
 	"github.com/openshift/osdctl/cmd/aao"
 	"github.com/openshift/osdctl/cmd/account"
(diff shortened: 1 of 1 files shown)


cmd/cmd.go
```diff
@@ -129,3 +108,47 @@ func help(cmd *cobra.Command, _ []string) {
 		fmt.Println("Error while printing help: ", err.Error())
 	}
 }
+
+// Checks if the version check should be run
+func shouldRunVersionCheck(skipVersionCheckFlag bool, commandName string) bool {
+
+	// If either are true, then the version check should NOT run, hence negation
+	return !(skipVersionCheckFlag || canCommandSkipVersionCheck(commandName))
```
fahlmant, line 116: This benefits from the fact that `!A AND !B` is equivalent to `!(A or B)` and the latter benefits from Short circuit evaluation, as shown by this small test https://go.dev/play/p/M0gfXelQf2y

PR #340, opened 2023-02-17
Initial jumphost create/delete

mjlshen:
This is groundwork for OSD-7236 that I would like to get reviewed before continuing.

### New
This PR creates two new commands: `osdctl jumphost create` and `osdctl jumphost delete` which is only supported for AWS, non-STS clusters (to limit initial impact, it can be un-gated as we gain confidence it will work when we ask customers to grant permissions X, Y, Z)

### What it does
Currently it just automates the EC2 Key Pair creation/deletion as a first step. If this PR gets merged the work in the story will continue to create the rest of the jumphost infra.

### Food for thought
The command only works currently if you give it a cluster-id, not via arbitrarily provided AWS credentials for now.


rhinofeeder:
Not related to this PR specifically, but I've noticed that some commands in osdctl take the cluster ID as an argument (i.e. `osdctl foo bar abcd1234`) while others take it as a flag (i.e. `osdctl foo bar -c abcd1234`). Do we have a preferred standard?

AlexVulaj:
Not related to this PR specifically, but I've noticed that some commands in osdctl take the cluster ID as an argument (i.e. `osdctl foo bar abcd1234`) while others take it as a flag (i.e. `osdctl foo bar -c abcd1234`). Do we have a preferred standard?

mjlshen:
> Do we have a preferred standard?

I don't think so 😅

mjlshen:
Future plans:

* Allow using a `--cluster-id` flag so you don't have to go and pick out a public subnet
* Allow specifying `--instance-type` in case `t3.micro` doesn't work
* Having the jumphost automatically shutdown after 8 hours to minimize the impact of forgetting to cleanup

cmd/jumphost/cmd.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/osdctl/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+const (
+	awsResourceName     = "red-hat-sre-jumphost"
+	publicSubnetTagKey  = "kubernetes.io/role/elb"
+	privateSubnetTagKey = "kubernetes.io/role/internal-elb"
```
AlexVulaj, line 20: are publicSubnetTagKey or privateSubnetTagKey used anywhere?

cmd/jumphost/create.go
```diff
@@ -0,0 +1,392 @@
...
+		return nil
+	}
+
+	log.Printf("created file %s", f.Name())
+	j.keyFilepath = f.Name()
+	return nil
+}
+
+// createSecurityGroup creates a security group and creates a single inbound rule to allow the user's public IP to SSH.
+func (j jumphostConfig) createSecurityGroup(ctx context.Context) (string, error) {
```
AlexVulaj, line 126: Should this method be on the pointer receiver?

cmd/jumphost/cmd.go
mjlshen: It's not used right now, but when I add back the `--cluster-id` flag these tags are commonly used to identify public and private subnets.

cmd/jumphost/create.go
mjlshen: I have mixed them where it's necessary to modify `jumphostConfig` - in this case `createSecurityGroup` does not need to modify `jumphostConfig` so I made it a value receiver.

I'm ok with making them all pointer receivers so that in the future people don't try to mutate state with a value receiver and get unexpected results.

mjlshen: Latest push converts all methods to act on to pointer receivers

PR #341, opened 2023-03-02
[OSD-15156] Create new client for AWS/GCP abstraction.

bergmannf:
Currently the client is only used for the health command and abstracts GCP & AWS into a common client.

### DIFF
--- a/cmd/cluster/health.go
+++ b/cmd/cluster/health.go
@@ -7,12 +7,10 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/aws/aws-sdk-go/service/ec2"
 	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/openshift/osdctl/pkg/osdCloud"
 	"github.com/openshift/osdctl/pkg/utils"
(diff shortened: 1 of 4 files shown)


dkeohane (approved):
Lgtm

pkg/osdCloud/common.go
```diff
@@ -0,0 +1,215 @@
+package osdCloud
```
iamkirkbater, line 1: files with names like `common.go` are usually an antipattern.

Could we abstract this further into `pkg/osdCloud/interfaces.go` `pkg/osdCloud/gcp.go` `pkg/osdCloud/aws.go`?

Also, instead of `osdCloud` maybe something like `cloudprovider`?  I think the camelCase will come back to bite us with package naming in the future.

iamkirkbater (changes requested):

pkg/osdCloud/common.go
bergmannf: There are other camelCase package names in osdctl already, so I'd leave it as it was for now.

But I moved the implementations of the clients to the respective files and only left the base types in interfaces.go

PR #342, opened 2023-03-08
Use aws_proxy_url for awsClient if it's set in the osdctl config

AlexVulaj:
This work comes as a result of [SDE-2305](https://issues.redhat.com/browse/SDE-2305) in preparation of future work.

When creating the the AWS Client - and by extension the Session - used for communication with AWS services, osdctl will check to see if "aws_proxy_url" is defined in `~/.config/osdctl` and use the defined value as a proxy for requests.

### DIFF
--- a/cmd/account/console.go
+++ b/cmd/account/console.go
@@ -102,7 +102,7 @@ func (o *consoleOptions) run() error {
 		}
 	}
 
-	// Build the base AWS client using the provide credentials (profile or env vars)
+	// Build the base AWS client using the provided credentials (profile or env vars)
 	awsClient, err := aws.NewAwsClient(o.awsProfile, o.region, "")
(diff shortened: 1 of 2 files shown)


iamkirkbater:
Nice work so far!  I'm going to hold this though, until after the design doc is accepted and we begin work on M2 of this epic (which will encapsulate the work done here)

/hold

AlexVulaj:
@iamkirkbater I agree about waiting until there is wider approval on the design doc. I'd still like to get this in early to see if we can't get some on-call SREs willing to volunteer to start using this out in "the field."

geowa4:
Superseded by #389

AlexVulaj:
Closed per @geowa4 's comment.

PR #347, opened 2023-03-22
Drop hive as a dependency to remove all the replace directives

mjlshen:
This lets us keep the K8s dependencies better up-to-date, in this case OCP 4.12/K8s 1.26.

This was enabled by https://github.com/openshift/hive/pull/1976 which removed the last replace statement from `github.com/openshift/hive/api`


pkg/k8s/client.go
```diff
@@ -27,7 +27,7 @@ func (s *LazyClient) err() error {
 	return fmt.Errorf("not connected to real cluster, please verify KUBECONFIG is correct")
 }
 
-func (s *LazyClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object) error {
+func (s *LazyClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
```
mjlshen, line 30: This reflects a change to this client interface in controller-runtime `v0.13.0`, and resulted in generating an updated mock as well.

cmd/clusterdeployment/list.go
```diff
@@ -14,6 +13,8 @@ import (
 	"github.com/openshift/osdctl/pkg/printer"
 )
 
+const hiveVersionMajorMinorPatchLabel string = "hive.openshift.io/version-major-minor-patch"
```
mjlshen, line 16: This constant was the only reason we have hive as a dependency. I acknowledge that this does represent some risk as Hive could choose to change this tag and we would be none the wiser, but this command has rarely been used since the initial v3 --> v4 migration.

PR #349, opened 2023-03-25
[OSD-15150] osdctl network verify-egress should have a way to specify all subnets

aliceh:
In this commit, I added an --all-networks option to osdctl network verify-egress commands. When this option is specified, the cluster will run the network verifier against all subnets listed by ocm if the cluster is Privatelink. If the cluster is not private link, and non-BYOVPC the --all-networks option will run the network verifier against all private subnets found by tags, and the command will run as it would without the --all-networks flag for all other cases.

https://issues.redhat.com/browse/OSD-15150

### DIFF
--- a/cmd/network/verification.go
+++ b/cmd/network/verification.go
@@ -36,7 +36,7 @@ type EgressVerification struct {
 	Region string
 	// SubnetId is an optional override for specifying an AWS subnet ID.
 	// Must be a private subnet to provide accurate results.
-	SubnetId string
+	SubnetId []string
 	// SecurityGroupId is an optional override for specifying an AWS security group ID.
 	SecurityGroupId string
 	// Debug optionally enables debug-level logging for underlying calls to osd-network-verifier.
@@ -47,6 +47,8 @@ type EgressVerification struct {
 	CaCert string
 	// NoTls optionally ignores SSL validation on the client-side.
 	NoTls bool
+	// AllSubnets is an option for multi-AZ clusters that will run the network verification against all subnets listed by ocm
+	AllSubnets bool
 }
 
 func NewCmdValidateEgress() *cobra.Command {
@@ -86,13 +88,14 @@ func NewCmdValidateEgress() *cobra.Command {
 		},
 	}
 
-	validateEgressCmd.Flags().StringVar(&e.ClusterId, "cluster-id", "", "(optional) OCM internal/external cluster id to run osd-network-verifier against.")
-	validateEgressCmd.Flags().StringVar(&e.SubnetId, "subnet-id", "", "(optional) private subnet ID override, required if not specifying --cluster-id")
+	validateEgressCmd.Flags().StringVarP(&e.ClusterId, "cluster-id", "C", "", "(optional) OCM internal/external cluster id to run osd-network-verifier against.")
+	validateEgressCmd.Flags().StringArrayVar(&e.SubnetId, "subnet-id", nil, "(optional) private subnet ID override, required if not specifying --cluster-id and can be specified multiple times to run against multiple subnets")
 	validateEgressCmd.Flags().StringVar(&e.SecurityGroupId, "security-group", "", "(optional) security group ID override for osd-network-verifier, required if not specifying --cluster-id")
 	validateEgressCmd.Flags().StringVar(&e.CaCert, "cacert", "", "(optional) path to a file containing the additional CA trust bundle. Typically set so that the verifier can use a configured cluster-wide proxy.")
 	validateEgressCmd.Flags().BoolVar(&e.NoTls, "no-tls", false, "(optional) if provided, ignore all ssl certificate validations on client-side.")
 	validateEgressCmd.Flags().StringVar(&e.Region, "region", "", "(optional) AWS region")
 	validateEgressCmd.Flags().BoolVar(&e.Debug, "debug", false, "(optional) if provided, enable additional debug-level logging")
+	validateEgressCmd.Flags().BoolVarP(&e.AllSubnets, "all-subnets", "A", false, "(optional) an option to run osd-network-verifier against all subnets listed by ocm.")
 
 	// If a cluster-id is specified, don't allow the foot-gun of overriding region
 	validateEgressCmd.MarkFlagsMutuallyExclusive("cluster-id", "region")
@@ -119,16 +122,19 @@ func (e *EgressVerification) Run(ctx context.Context) {
 		log.Fatal(fmt.Errorf("failed to assemble osd-network-verifier client: %s", err))
 	}
 
-	input, err := e.generateAWSValidateEgressInput(ctx, cfg.Region)
+	inputs, err := e.generateAWSValidateEgressInput(ctx, cfg.Region)
 	if err != nil {
 		log.Fatal(err)
 	}
-	e.log.Info(ctx, "running with config: %+v", input)
 
-	out := onv.ValidateEgress(c, *input)
-	out.Summary(e.Debug)
-	if out.IsSuccessful() {
-		log.Println("All tests pass")
+	for i, _ := range inputs {
+		e.log.Info(ctx, "running with config: %+v", inputs[i])
+		e.log.Info(ctx, "running network verifier for subnet  %+v, security group %+v", inputs[i].SubnetID, inputs[i].AWS.SecurityGroupId)
+		out := onv.ValidateEgress(c, *inputs[i])
+		out.Summary(e.Debug)
+		if out.IsSuccessful() {
+			log.Println("All tests pass")
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cmd/network/verification.go
```diff
@@ -195,32 +201,32 @@ func (e *EgressVerification) setup(ctx context.Context) (*aws.Config, error) {
 // Its input is an OCM internal/external ClusterId and it returns the corresponding input to osd-network-verifier with
 // default AWS tags, one of the cluster's private subnet IDs, and the cluster's master security group.
 // Can override SecurityGroupId and SubnetId.
-func (e *EgressVerification) generateAWSValidateEgressInput(ctx context.Context, region string) (*onv.ValidateEgressInput, error) {
+func (e *EgressVerification) generateAWSValidateEgressInput(ctx context.Context, region string) (*onv.ValidateEgressInput, []*onv.ValidateEgressInput, error) {
```
mjlshen, line 204:
```suggestion
func (e *EgressVerification) generateAWSValidateEgressInput(ctx context.Context, region string) ([]*onv.ValidateEgressInput, error) {
```

It changes code in this function when returning, but lets only return the slice of `ValidateEgressInput` since we won't need the first return value anymore

cmd/network/verification.go
```diff
@@ -87,12 +89,13 @@ func NewCmdValidateEgress() *cobra.Command {
 	}
 
 	validateEgressCmd.Flags().StringVar(&e.ClusterId, "cluster-id", "", "(optional) OCM internal/external cluster id to run osd-network-verifier against.")
-	validateEgressCmd.Flags().StringVar(&e.SubnetId, "subnet-id", "", "(optional) private subnet ID override, required if not specifying --cluster-id")
+	validateEgressCmd.Flags().StringArrayVar(&e.SubnetId, "subnet-id", nil, "(optional) private subnet ID override, required if not specifying --cluster-id")
```
mjlshen, line 92:
```suggestion
	validateEgressCmd.Flags().StringArrayVar(&e.SubnetId, "subnet-id", nil, "(optional) private subnet ID override, required if not specifying --cluster-id" and can be specified multiple times to run against multiple subnets)
```
We should add some documentation around the new functionality. I think an example would be nice too above, around lines 69-85

cmd/network/verification.go
```diff
@@ -280,30 +309,30 @@ func (e *EgressVerification) getSubnetId(ctx context.Context) (string, error) {
...
 		})
 		if err != nil {
-			return "", fmt.Errorf("failed to find private subnets for %s: %w", e.cluster.InfraID(), err)
+			return nil, fmt.Errorf("failed to find private subnets for %s: %w", e.cluster.InfraID(), err)
 		}
 
 		if len(resp.Subnets) == 0 {
-			return "", fmt.Errorf("found 0 subnets with kubernetes.io/cluster/%s=owned and %s, consider the --subnet-id flag", e.cluster.InfraID(), e.cluster.InfraID())
+			return nil, fmt.Errorf("found 0 subnets with kubernetes.io/cluster/%s=owned and %s, consider the --subnet-id flag", e.cluster.InfraID(), e.cluster.InfraID())
 		}
```
mjlshen, line 318:
```suggestion
	if e.AllSubnets {
	  subnetIds := make([]string, *resp.Subnets)
	  for i, subnet := range *resp.Subnets {
	    subnetIds[i] = subnet.SubnetId
	  }
	  
	  e.log.Info(ctx, "using subnet-ids: %v", subnetIds)
    return subnetIds, nil
	}
```
We can find all the private subnets for non-BYOVPC clusters too

cmd/network/verification.go
```diff
@@ -280,30 +309,30 @@ func (e *EgressVerification) getSubnetId(ctx context.Context) (string, error) {
...
-		return *resp.Subnets[0].SubnetId, nil
+		return []string{*resp.Subnets[0].SubnetId}, nil
 	}
 
 	// For PrivateLink clusters, any provided subnet is considered a private subnet
 	if e.cluster.AWS().PrivateLink() {
 		if len(e.cluster.AWS().SubnetIDs()) == 0 {
-			return "", fmt.Errorf("unexpected error: %s is a PrivateLink cluster, but no subnets in OCM", e.cluster.InfraID())
+			return nil, fmt.Errorf("unexpected error: %s is a PrivateLink cluster, but no subnets in OCM", e.cluster.InfraID())
 		}
```
mjlshen, line 328: The earlier logic ~line 281 should shift here since we only want to grab all the subnets from OCM when it is a PrivateLink cluster (checked on line 324). Otherwise, for non-PrivateLink BYOVPC clusters all public and private subnets are listed in OCM, which will cause network verifier to confusingly fail when it runs on the public subnets.

cmd/network/verification.go
```diff
@@ -87,12 +89,13 @@ func NewCmdValidateEgress() *cobra.Command {
...
+	validateEgressCmd.Flags().StringArrayVar(&e.SubnetId, "subnet-id", nil, "(optional) private subnet ID override, required if not specifying --cluster-id")
 	validateEgressCmd.Flags().StringVar(&e.SecurityGroupId, "security-group", "", "(optional) security group ID override for osd-network-verifier, required if not specifying --cluster-id")
 	validateEgressCmd.Flags().StringVar(&e.CaCert, "cacert", "", "(optional) path to a file containing the additional CA trust bundle. Typically set so that the verifier can use a configured cluster-wide proxy.")
 	validateEgressCmd.Flags().BoolVar(&e.NoTls, "no-tls", false, "(optional) if provided, ignore all ssl certificate validations on client-side.")
 	validateEgressCmd.Flags().StringVar(&e.Region, "region", "", "(optional) AWS region")
 	validateEgressCmd.Flags().BoolVar(&e.Debug, "debug", false, "(optional) if provided, enable additional debug-level logging")
+	validateEgressCmd.Flags().BoolVar(&e.AllSubnets, "all-subnets", false, "(optional) an option to run osd-network-verifier against all subnets listed by ocm.")
```
mjlshen, line 98: What do you think of changing this to `BoolVarP` and allowing the `-A` flag to be shorthand for `--all-subnets`?

mjlshen (changes requested):
This looks great and I'm sure it will find a lot of use! Just a few tweaks to get it over the finish line - I'm happy to give some pointers on the unit tests as well if needed.

cmd/network/verification_test.go
```diff
@@ -64,7 +64,7 @@ func Test_egressVerificationSetup(t *testing.T) {
 			name: "ClusterId optional",
 			e: &EgressVerification{
 				ClusterId:       "",
-				SubnetId:        "subnet-a",
+				SubnetId:        []string{"subnet-a"},
```
fahlmant, line 67: Can you also write a test with multiple subnets?

aliceh: Yes, we definitely need more test coverage.

aliceh:
I am not sure what UX would be better, to let the user specify the -A flag whenever, and only have it iterate through all networks if the cluster is privatelink, or when the -A flag is not appropriate have the command throw an error explaining that the -A option is only valid for privatelink clusters?

aliceh:
Alternatively, we can add an Info level log output noting that the -A flag does not do anything for those cases when it does not apply.

mjlshen:
> I am not sure what UX would be better, to let the user specify the -A flag whenever, and only have it iterate through all networks if the cluster is privatelink, or when the -A flag is not appropriate have the command throw an error explaining that the -A option is only valid for privatelink clusters?

+1 to letting the user specify `-A` whenever - though the difference is more when the cluster is multi-AZ or not right (I could be missing something)?

cmd/network/verification.go
```diff
@@ -280,30 +309,31 @@ func (e *EgressVerification) getSubnetId(ctx context.Context) (string, error) {
...
 		}
 
 		if len(resp.Subnets) == 0 {
-			return "", fmt.Errorf("found 0 subnets with kubernetes.io/cluster/%s=owned and %s, consider the --subnet-id flag", e.cluster.InfraID(), e.cluster.InfraID())
+			return nil, fmt.Errorf("found 0 subnets with kubernetes.io/cluster/%s=owned and %s, consider the --subnet-id flag", e.cluster.InfraID(), e.cluster.InfraID())
 		}
 
 		e.log.Info(ctx, "using subnet-id: %s", *resp.Subnets[0].SubnetId)
-		return *resp.Subnets[0].SubnetId, nil
+		return []string{*resp.Subnets[0].SubnetId}, nil
```
mjlshen, line 320: Something like:
```
	if e.AllSubnets {
	    return all the subnets
	 }
```

cmd/network/verification_test.go
```diff
@@ -390,3 +395,84 @@ func compareValidateEgressInput(expected, actual *onv.ValidateEgressInput) bool
...
+		},
+		{
+			name: "non-BYOVPC clusters get subnets from AWS, with --all-subnets flag",
+			e: &EgressVerification{
+				awsClient: mockEgressVerificationAWSClient{
+					describeSubnetsResp: &ec2.DescribeSubnetsOutput{
+						Subnets: []types.Subnet{
+							{
+								SubnetId: aws.String("subnet-abcd"),
+							},
```
mjlshen, line 445: We can add a test case for multiple subnets here

mjlshen (approved):
/approve

fahlmant:
Please remember to squash commits or apply the squash label before approving

PR #350, opened 2023-03-28
Bumping AAO Api version to include service quota changes in Account CR

dkeohane:
Need to bump the AAO Api version to now include the Service Quota changes in the Account CR. These changes are needed specifically for the account reset functionality - without them, if we attempt to reset an Account CR, we'll wipe the Service Quota from the Account Spec.


go.sum
```diff
@@ -704,6 +704,8 @@ github.com/openshift/api v0.0.0-20230320211411-560b2fb170af h1:js8+NVUCTK7cz+IO6
 github.com/openshift/api v0.0.0-20230320211411-560b2fb170af/go.mod h1:ctXNyWanKEjGj8sss1KjjHQ3ENKFm33FFnS5BKaIPh4=
 github.com/openshift/aws-account-operator/api v0.0.0-20220921130557-58a04a5f1ba2 h1:qK7JrY2cBOKN/cDnHI6kRVHOIqMWHhkV7b3BQgldQ7A=
 github.com/openshift/aws-account-operator/api v0.0.0-20220921130557-58a04a5f1ba2/go.mod h1:1PdbQqTDrejSl9zsScM1x59f0oHNTsAgoJqTZqTkH/U=
```
mjlshen, line 706: Changes make sense, but I think this needs a `go mod tidy` which will delete these two lines

mjlshen (approved):
/lgtm

PR #356, opened 2023-03-30
use single-quoted character literal for region value

T0MASD:
currently osdctl account cli -o json returns invalid json object because region value is not quoted
```
{
"AccessKeyId": "***", 
"Expiration": "***", 
"SecretAccessKey": "***", 
"SessionToken": "***", 
"Region": us-west-2
}
```


fahlmant (approved):
/lgtm
/approve

PR #357, opened 2023-04-04
refactor getting AWS client from backplane-api

mjlshen:
[OSD-15862](https://issues.redhat.com//browse/OSD-15862)

There's a new backplane-cli which functions differently:

github.com/openshift/backplane-api and
github.com/openshift/backplane-cli

especially in terms of the URL of backplane-api.

This PR is quite easy to test via `osdctl network verify-egress --cluster-id ${CLUSTER_ID}`

e.g. before:
```
❯ osdctl network verify-egress --cluster-id ${CLUSTER_ID}                         
2023/04/04 11:02:25 getting AWS credentials from backplane-api
2023/04/04 11:02:55 failed to retrieve AWS credentials from backplane, ensure there is an active backplane tunnel: Get "https://api-backplane.apps.hivep06uw2.bjee.p1.openshiftapps.com/backplane/cloud/credentials/${CLUSTER_ID}": dial tcp 10.135.143.238:443: i/o timeout 
```

(works after)

`osdctl cluster resize-control-plane-node` uses similar logic and so I've chosen to refactor it to use aws-sdk-go-v2 rather than implement an aws-sdk-go client in backplane-cli


mjlshen:
/hold

This PR isn't backwards compatible with the old, hive-hosted backplane-api

iamkirkbater:
Why are we even getting the login from backplane here in the first place?  osdctl has it's own AWS credential generation logic, we should be utilizing that instead here, and entirely removing the dependency on backplane from osdctl. 

I don't see a long-term value in coupling backplane to osdctl, but I might be missing something here.

mjlshen:
It's mostly a UX reason that allows `osdctl network verify-egress --cluster-id ${CLUSTER_ID}` and other commands that use this to work without having depending on a profile.

This is just personal experience but it's the same reason why I always prefer to `ocm backplane cloud credentials` vs `osdctl account cli` unless I have to use osdctl (e.g. AWS account not associated with a cluster)

pkg/osdCloud/aws.go
```diff
@@ -237,48 +238,50 @@ func CreateAWSV2Config(ctx context.Context, clusterID string) (awsv2.Config, err
...
 func GetAWSClientInputFromBackplane(clusterID string) (*aws.AwsClientInput, error) {
-	token, err := utils.GetOCMApiServerToken()
-
-	getUrl, err := utils.GetBackplaneURL(clusterID)
-	if err != nil {
-		return nil, fmt.Errorf("Unable to retrieve backplane URL for cluster %s: %s", clusterID, err)
+	ocmToBackplane := map[string]string{
+		"https://api.openshift.com":             "https://api.backplane.openshift.com",
+		"https://api.stage.openshift.com":       "https://api.stage.backplane.openshift.com",
+		"https://api.integration.openshift.com": "https://api.integration.backplane.openshift.com",
```
wanghaoran1988, line 244: Could we get those URL from the backplane config? we are intent to not expose those URLs in public.

mjlshen: Definitely, I would like to - I think what we are missing currently is helper functions to read the backplane config, but will look into writing some in backplane-cli

mjlshen: I tested the new functions I added in https://github.com/mjlshen/mirrosa/pull/138 (I'll port that over here as I think it satisfies this request!)

go.mod
```diff
@@ -6,20 +6,20 @@ require (
...
 	github.com/fatih/color v1.13.0
 	github.com/golang/mock v1.6.0
 	github.com/onsi/gomega v1.23.0
-	github.com/openshift-online/ocm-cli v0.1.65-0.20220913083421-16b44c698a1f
-	github.com/openshift-online/ocm-sdk-go v0.1.300
+	github.com/openshift-online/ocm-cli v0.1.65
+	github.com/openshift-online/ocm-sdk-go v0.1.316
 	github.com/openshift/api v0.0.0-20230320211411-560b2fb170af
 	github.com/openshift/aws-account-operator/api v0.0.0-20230322125717-5b5a00a3e99f
+	github.com/openshift/backplane-cli v0.1.3-0.20230413011330-a35ce4597ee5
```
mjlshen, line 22: We should wait for this code to be in a released version of backplane-cli

pkg/osdCloud/aws.go
mjlshen, line 1: More reviews always welcome :D but would especially appreciate a check from @wanghaoran1988 on this refactor

cmd/cluster/resizecontrolplanenode.go
mjlshen, line 1: More reviews always welcome, but would especially appreciate a review from @typeid

pkg/osdCloud/aws.go
wanghaoran1988: this looks good to me, thanks :)

mjlshen:
backplane-cli v0.1.3 was released, which unblocks this PR after a rebase 🎉 

/unhold

typeid:
> Why are we even getting the login from backplane here in the first place? osdctl has it's own AWS credential generation logic, we should be utilizing that instead here, and entirely removing the dependency on backplane from osdctl.
> 
> I don't see a long-term value in coupling backplane to osdctl, but I might be missing something here.

This was initially for a UX reason as Michael said. Now that the profiles are part of the osdctl config file, I think we can rethink moving away from backplane api. 
Note that some commands require access to a prod and non-prod cluster, which might complicate the jumpRole chain in some cases (hive = prod, cluster = staging example command would be the resize-control-plane), as they will require different starting profiles.

PR #358, opened 2023-04-05
fix bug where refreshToken was not being parsed when needed

mjlshen:
The existing code would trip up if a user had set the `OCM_TOKEN` environment variable but also uses `ocm login` as the refresh_token would never be read from the ocm config file.

```go
	if token == "" {
		token = config.AccessToken
		refresh_token = config.RefreshToken

		// Can't both be nil
		if token == "" && refresh_token == "" {
			log.Fatal(ocmConfigError)
			return nil
		}
	}
```

This PR addresses this by attempting to read the OCM config file first - then filling in any missing values from environment variables.

### DIFF
--- a/pkg/utils/ocm.go
+++ b/pkg/utils/ocm.go
@@ -214,63 +214,51 @@ func loadOCMConfig() (*Config, error) {
 }
 
 func CreateConnection() *sdk.Connection {
-	token := os.Getenv("OCM_TOKEN")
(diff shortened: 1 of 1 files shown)


mjlshen:
/hold

This works for non-ocm-container, but someone should test this on ocm-container

mjlshen:
/close

In favor of #359

PR #359, opened 2023-04-05
fix bug where refreshToken was not being parsed when OCM_TOKEN was set

typeid:
Fixes:
- refreshToken is now being parsed as well when OCM_TOKEN is set

Refactor:
- getOcmConfiguration creates the minimal OCM configuration needed to create a connection: it prioritises environment variables but uses the ocm configuration file as base in case not all ENV values are set.

### DIFF
--- a/pkg/utils/ocm.go
+++ b/pkg/utils/ocm.go
@@ -221,10 +221,10 @@ func getOcmConfiguration() (*Config, error) {
 	config := &Config{}
 	var err error
 
-	// If any of the environment variables 'OCM_TOKEN', 'OCM_REFRESH_TOKEN', 'OCM_URL' are not set,
+	// If neither of the token ENVS 'OCM_TOKEN', 'OCM_REFRESH_TOKEN' are set, or 'OCM_URL' is not set,
 	// use the configuration file as base.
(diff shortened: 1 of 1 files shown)


pkg/utils/ocm.go
```diff
@@ -213,61 +213,62 @@ func loadOCMConfig() (*Config, error) {
...
 	config := &Config{}
-	err := error(nil)
+	var err error
 
-	if token == "" || url == "" {
-		// If either token or url are not set, try to load them from the config file
+	// If any of the environment variables 'OCM_TOKEN', 'OCM_REFRESH_TOKEN', 'OCM_URL' are not set,
+	// use the configuration file as base.
+	// We don't want to always load this, because the user might only use environment variables.
+	if tokenEnv == "" || urlEnv == "" || refreshTokenEnv == "" {
```
mjlshen, line 227: I think ocm-container will fall into this block and result in an unrecoverable error when it fails to load the ocm config because they may not set OCM_REFRESH_TOKEN

typeid: Changed to check if:
- both tokens are empty
**or**
- url is empty

PR #363, opened 2023-04-06
Convert from infra id to internal and external id

geowa4:
When looking at security alerts, the ID commonly shown is not the ID by which we describe or to which we send service logs. With this change, we can more easily and efficiently discover the cluster's ID and the replace the Bash script in our SOPs.

### DIFF
--- a/cmd/cluster/from_infra_id.go
+++ b/cmd/cluster/from_infra_id.go
@@ -1,0 +1,107 @@
+package cluster
+
+import (
+	"encoding/json"
+	"fmt"
+	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/osdctl/internal/utils/globalflags"
+	"github.com/openshift/osdctl/pkg/utils"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"os"
+	"strings"
+)
+
+type fromInfraIdOptions struct {
+	globalOpts *globalflags.GlobalOptions
+}
+
+type clusterIdentification struct {
+	ID         string `json:"id" yaml:"ID"`
+	ExternalID string `json:"external_id" yaml:"ExternalID"`
+	Name       string `json:"name" yaml:"Name"`
+}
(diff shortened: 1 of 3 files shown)


cmd/cluster/from_infra_id.go
```diff
@@ -0,0 +1,107 @@
...
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(opts.complete(cmd, args))
+			cmdutil.CheckErr(opts.run(cmd, args))
+		},
+	}
+}
+
+func (ops *fromInfraIdOptions) complete(cmd *cobra.Command, args []string) error {
+	if len(args) != 1 {
```
AlexVulaj, line 47: It's not hurting anything, but this if statement shouldn't be needed as you're ensuring `cobra.ExactArgs(1)` on the command definition.

geowa4: Good call.

geowa4: Fixed.

cmd/cluster/from_infra_id.go
```diff
@@ -0,0 +1,99 @@
...
+	ocmClient := utils.CreateConnection()
+	defer func() {
+		if err := ocmClient.Close(); err != nil {
+			fmt.Printf("cannot close the ocmClient (possible memory leak): %q", err)
+		}
+	}()
+
+	infraId := args[0]
+	clusterNameAndHash := strings.Split(infraId, "-")
+	if len(clusterNameAndHash) != 2 {
```
mjlshen, line 55: A cluster name (not just the infra id) can have a "-" by itself

AlexVulaj: If you're still looking for ideas, `strings.LastIndex` might be helpful here. This is completely untested but I think it would look something like:
```
lastIndex := strings.LastIndex(infraId, "-")
clusterName := infraId[:lastIndex]
clusterHash := infraId[lastIndex+1:]
```

geowa4: Fixed and tested

mjlshen (approved):
/lgtm

PR #364, opened 2023-04-07
Cleanup

geowa4:
Recently when making changes to this project, I noticed a couple things to tidy up.

1. the Makefile did not run if you had a space in a PATH element
2. using gmake on macOS requires fewer changes and is what our Linux colleagues are using so doc that in the README
3. doc generation for new and updated commands
4. fix copy-paste issue in a test I recently added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ unexport GOFLAGS
 # but otherwise can be installed elsewhere on developers machines
 BASE_DIR=$(shell pwd)
 export PATH:=$(BASE_DIR)/bin:$(PATH)
-SHELL := env PATH=$(PATH) /bin/bash
+SHELL := /bin/bash
 
 
 all: format mod build test lint
@@ -39,7 +39,7 @@ build:
 	goreleaser build --clean --snapshot --single-target=${SINGLE_TARGET}
 
 release:
-	./bin/goreleaser release --clean
+	goreleaser release --clean
 
 vet:
 	go vet ${BUILDFLAGS} ./...
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Currently, it mainly supports related work for AWS, especially [aws-account-oper
 #### Requirements
 
 - Go >= 1.19
-- make
+- make (gmake if on macOS)
 - [goreleaser](https://github.com/goreleaser)
 - `GOPROXY` contains `proxy.golang.org` as highest priority
   ```shell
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -6,7 +6,7 @@ unexport GOFLAGS
 # but otherwise can be installed elsewhere on developers machines
 BASE_DIR=$(shell pwd)
 export PATH:=$(BASE_DIR)/bin:$(PATH)
-SHELL := env PATH=$(PATH) /bin/bash
+SHELL := /bin/bash
```
geowa4, line 9: Please take note of this. I don't think this will impact anything, but it's probably the the most significant change in this PR.

README.md
```diff
@@ -15,7 +15,7 @@ Currently, it mainly supports related work for AWS, especially [aws-account-oper
 #### Requirements
 
 - Go >= 1.19
-- make
+- make (gmake if on macOS)
```
geowa4, line 18: I'm recommending gmake on macOS, because it behaves differently from the built-in make and aligns with our Linux colleagues. For example, the built-in make doesn't seem to respect the PATH manipulation at the top of README.md.

Makefile
clcollins: If I understand this correctly, the original line is trying to reset the env to use the actual user $PATH to then pick the bash interpreter.  I *think* this is fine.

Makefile
```diff
@@ -39,7 +39,7 @@ build:
 	goreleaser build --clean --snapshot --single-target=${SINGLE_TARGET}
 
 release:
-	./bin/goreleaser release --clean
+	goreleaser release --clean
```
clcollins, line 42: I also think this won't be a big deal, but what problem were you solving here?  I think either way it would be picking goreleaser from the local `./bin` dir, because of the exported `$PATH` (line 8).  Was there another issue?

clcollins (commented):
I think this is all generally LGTM.  The Makefile changes still work fine with my local testing, and the other changes are good.

Only a couple comments/questions, but nothing that should be considered blocking for anyone else if they want to add an LGTM.

Makefile
geowa4: Really, my goal here was to normalize how we called it and I picked the way `make build` did it, but either make the same sense to me.

Makefile
geowa4: Yeah, I would like to keep this how it was but since installing JetBrains Toolbox, I now have a space in my PATH that breaks this line as is. I'm no Makefile expert, but this worked for me.

Makefile
clcollins: Works for me.  I was just curious about the change.

PR #367, opened 2023-04-07
Make ocm.go testable

geowa4:
Refactor and add tests to ocm.go. Functionality should be little changed, but I believe I found the cause of a bug and made the correction in ocm.go.

### DIFF
--- a/pkg/utils/ocm.go
+++ b/pkg/utils/ocm.go
@@ -213,21 +213,20 @@ func loadOCMConfig() (*Config, error) {
 	return cfg, nil
 }
 
-func getOcmConfiguration() (*Config, error) {
+func getOcmConfiguration(ocmConfigLoader func() (*Config, error)) (*Config, error) {
 	tokenEnv := os.Getenv("OCM_TOKEN")
 	urlEnv := os.Getenv("OCM_URL")
 	refreshTokenEnv := os.Getenv("OCM_REFRESH_TOKEN") // Unlikely to be set, but check anyway
 
 	config := &Config{}
-	var err error
 
-	// If neither of the token ENVS 'OCM_TOKEN', 'OCM_REFRESH_TOKEN' are set, or 'OCM_URL' is not set,
-	// use the configuration file as base.
+	// If missing required data, load from the config file.
 	// We don't want to always load this, because the user might only use environment variables.
-	if (tokenEnv == "" && refreshTokenEnv == "") || urlEnv == "" {
-		config, err = loadOCMConfig()
-		if err != nil {
-			return &Config{}, fmt.Errorf("Could not load OCM configuration file")
+	if tokenEnv == "" || refreshTokenEnv == "" || urlEnv == "" {
+		var fileConfigLoadError error
+		config, fileConfigLoadError = ocmConfigLoader()
+		if fileConfigLoadError != nil {
+			return config, fmt.Errorf("could not load OCM configuration file")
 		}
 	}
 
@@ -251,7 +250,7 @@ func CreateConnection() *sdk.Connection {
 
 	connectionBuilder := sdk.NewConnectionBuilder()
 
-	config, err := getOcmConfiguration()
+	config, err := getOcmConfiguration(loadOCMConfig)
 	if err != nil {
 		log.Fatal(ocmConfigError)
 	}
--- a/pkg/utils/ocm_test.go
+++ b/pkg/utils/ocm_test.go
@@ -1,0 +1,135 @@
+package utils
+
+import (
+	"os"
+	"testing"
+)
+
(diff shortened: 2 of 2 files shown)


pkg/utils/ocm.go
```diff
@@ -213,34 +213,43 @@ func loadOCMConfig() (*Config, error) {
...
+	)
 
 	// If neither of the token ENVS 'OCM_TOKEN', 'OCM_REFRESH_TOKEN' are set, or 'OCM_URL' is not set,
 	// use the configuration file as base.
 	// We don't want to always load this, because the user might only use environment variables.
-	if (tokenEnv == "" && refreshTokenEnv == "") || urlEnv == "" {
-		config, err = loadOCMConfig()
-		if err != nil {
-			return &Config{}, fmt.Errorf("Could not load OCM configuration file")
+	if tokenEnv == "" || refreshTokenEnv == "" || urlEnv == "" {
```
AlexVulaj, line 230: Does this warrant changing the language of the above comment?

typeid: +1 on changing the above documentation to explain this change. 
From my understanding we only need either the refresh token or the token alongside an URL for an ocm connection, which is why we had this logic. Is this understanding wrong?

pkg/utils/ocm.go
```diff
@@ -213,34 +213,43 @@ func loadOCMConfig() (*Config, error) {
...
 	}
 	if urlEnv != "" {
 		config.URL = urlEnv
+	} else {
+		config.URL = defaultConfig.URL
 	}
 	if refreshTokenEnv != "" {
 		config.RefreshToken = refreshTokenEnv
+	} else {
+		config.RefreshToken = defaultConfig.RefreshToken
```
typeid, line 252: What's the benefit here of writing to a new config object instead of overwriting the same potentially already loaded one with the environment variables (if they exist)? It seems less readable to me.

pkg/utils/ocm.go
geowa4: When I had OCM_TOKEN set and not OCM_REFRESH_TOKEN, I encountered the error below. I'll comment more privately. 

```
FATA[0002] Cannot retrieve clusters: "Get \"https://api.stage.openshift.com/api/clusters_mgmt/v1/clusters?search=omitted&size=50\": can't get access token: invalid_grant: Invalid refresh token" 
```

pkg/utils/ocm.go
geowa4: I agree with you that it is a bit annoying. I couldn't immediately think of a way to handle optionally loading the config from the file without changing the contract of `loadOcmConfig` to only fill in unset values. But I see now that `loadOcmConfig` is only used once so I could pass in a reference to the config object. 

I also now see that `loadOcmConfig` does not throw an error if the config file is not present. I'm not sure how I feel about that, but we could use that to our advantage and invert the logic here. Is that what you were thinking? 

I think I'm about even on all three options--leaving it as is in this PR, changing to pass a `*Config` reference to `loadOcmConfig`, and overwriting the result of `loadOcmConfig` if the env vars are present. Let me know which direction you prefer, and I'll make the change. Thanks 🙇🏻

typeid: >I agree with you that it is a bit annoying. I couldn't immediately think of a way to handle optionally loading the config from the file without changing the contract of loadOcmConfig to only fill in unset values.
Do we need loadOcmConfig to only fill unset values? The way `getOcmConfiguration` currently works, I don't think we do?

>I also now see that loadOcmConfig does not throw an error if the config file is not present.
That's a bit nasty, but also seems intended. It's no big deal here. If the configuration couldn't be loaded, it's an empty configuration, and we will run into an error when trying to establish the connection. I wouldn't worry about it in this PR.

I would prefer leaving it as is in this specific PR, and revert the above block of code to only overwrite the already initiated `Config` object. 

Basically, what we're currently doing is (correct me if I'm wrong here):

```
var config
var tmpConfig

if should_load_ocm_config_file
   tmpConfig = loadOcmConfigFile()

// Check which one has a value with priority and load it into our final config object
// If other values are set by `loadOcmConfigFile`, we lose them.
if OCM_ENV_1 != "" {
    config.ocm_1 = OCM_ENV_1
else {
   config.ocm_1 = tmpConfig.ocm_1
```
vs before
```
var config

if should_load_ocm_config_file
   // Pre-load into the config object
   config = loadOcmConfigFile()

// Overwrite the same config object, in case the env has a value
if OCM_ENV_1 != "" {
    config.ocm_1 = OCM_ENV_1
```

geowa4: Ok, I see you. One minute.

geowa4: Updated; pushed; tests still pass.

PR #370, opened 2023-04-14
Resolves OSD-15315

givaldolins:
This MR changes the behaviour for the context command to: 
1- Show all PD alerts for all PD services by default 
2- Allow users to filter the PD Alerts by Team when specifying the Team ID associated with the Service ID. 
3- The Team ID can be passed either via CLI or osdctl config file. (CLI has priority)


cmd/cluster/context.go
```diff
@@ -601,28 +619,38 @@ func GetPagerdutyClient(usertoken string, oauthtoken string) (*pd.Client, error)
...
 	client, err = getPDOauthClient(oauthtoken)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create both user and oauth clients for pd")
+		return nil, fmt.Errorf("failed to create both user and oauth clients for pd, neither key pd_oauth_token or pd_user_token are set in config file")
 	}
 	return client, err
 }
 
-func GetPDSeviceID(baseDomain string, usertoken string, oauthtoken string) (string, error) {
+func GetPDSeviceID(baseDomain string, usertoken string, oauthtoken string, team_ids []string) ([]string, error) {
```
mjlshen, line 630:
```suggestion
func GetPDServiceID(baseDomain string, usertoken string, oauthtoken string, team_ids []string) ([]string, error) {
```

Small nit :D Also where this function gets called in this file

cmd/cluster/context.go
```diff
@@ -601,28 +619,38 @@ func GetPagerdutyClient(usertoken string, oauthtoken string) (*pd.Client, error)
...
 	ctx := context.TODO()
 	pdClient, err := GetPagerdutyClient(usertoken, oauthtoken)
-	lsResponse, err := pdClient.ListServicesWithContext(ctx, pd.ListServiceOptions{Query: baseDomain})
+	if err != nil {
+		fmt.Printf("Failed to GetPagerdutyClient %q\n", err)
+		return []string{}, err
```
mjlshen, line 635:
```suggestion
		return nil, fmt.Errorf("failed to GetPagerdutyClient: %w", err)
```

Generally we shouldn't be printing errors when they happen and return them up the stack instead.

cmd/cluster/context.go
```diff
@@ -601,28 +619,38 @@ func GetPagerdutyClient(usertoken string, oauthtoken string) (*pd.Client, error)
...
+	}
+
+	// Gets the PD Team IDS
+	teams := getPDTeamIDs(team_ids)
+
+	lsResponse, err := pdClient.ListServicesWithContext(ctx, pd.ListServiceOptions{Query: baseDomain, TeamIDs: teams})
```
mjlshen, line 641:
```suggestion
	lsResponse, err := pdClient.ListServicesWithContext(context.TODO(), pd.ListServiceOptions{Query: baseDomain, TeamIDs: teams})
```

No need to store this in a variable named `ctx` if we're just using it once. Let's delete line 631 as well where this variable is defined.

cmd/cluster/context.go
mjlshen, line 630: With regards to your comment in the story:
```
After some troubleshooting, I found that the latest version of osdctl changed how cluster context is handled and now requires that  pd_oauth_token, pd_user_token, and jira_token are either passed as parameters or informed in the osdctl config file.
```

Can we add an initial check to this function with something like this (exact error message up for debate :D)? Would that let `osdctl cluster context` work without specifying either of these and just print an informational message about how the user could get PagerDuty information if they wanted?

```go
if usertoken == "" || oauthtoken == "" {
  return nil, errors.New("--usertoken and --oauthtoken must be specified to view PagerDuty information")
}
```

mjlshen (changes requested):
The PR seems reasonable. I'm not sure there is an existing use-case for wanting to get the alerts from multiple `team_ids`, but I suppose there is also no harm.

mjlshen (approved):
/label tide/merge-method squash
/lgtm

PR #373, opened 2023-04-20
Use existing struct from the SDK instead of our own custom code

AlexVulaj:
Removes a custom struct (and the code needed to translate to it) in favor of using the structs defined by the OCM sdk.

OSD-16105


mjlshen (approved):
/lgtm

PR #374, opened 2023-04-21
When rotating secrets, check for IAM username without account ID suffix label at the end

AlexVulaj:
As part of investigating https://issues.redhat.com/browse/OHSS-21052, we determined that osdctl was failing to rotate secrets because older clusters do not have the account ID suffix labels appended at the end of IAM usernames. This fix will have osdctl check to see if the IAM username is present without the label, and then check again with the label upon failure.

### DIFF
--- a/cmd/account/rotate-secret.go
+++ b/cmd/account/rotate-secret.go
@@ -3,6 +3,7 @@ package account
 import (
 	"context"
 	"fmt"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


mjlshen (approved):
/approve

cmd/account/rotate-secret.go
```diff
@@ -205,14 +205,17 @@ func (o *rotateSecretOptions) run() error {
...
 	// Username is osdManagedAdmin-aaabbb
-	osdManagedAdminUsername := common.OSDManagedAdminIAM + "-" + accountIDSuffixLabel
+	osdManagedAdminUsername := common.OSDManagedAdminIAM
 
 	// Create new access key
-	createAccessKeyOutput, err := awsClient.CreateAccessKey(&iam.CreateAccessKeyInput{
-		UserName: aws.String(osdManagedAdminUsername),
-	})
+	createAccessKeyOutput, err := awsClient.CreateAccessKey(&iam.CreateAccessKeyInput{UserName: aws.String(osdManagedAdminUsername)})
 	if err != nil {
```
wshearn, line 212: instead of checking if not nil here, can we check if it is the user not found error?

AlexVulaj: Added a check below.

iamkirkbater:
Was this able to have been tested on a cluster in staging?

AlexVulaj:
@iamkirkbater Yes, this was tested on a staging cluster for both the new naming convention (`osdManagedAdmin-abcdef`) and the older convention (`osdManagedAdmin`).

PR #376, opened 2023-04-21
Fix bug where osdctl network verify-egress was not requiring --cacert

mjlshen:
… when no http_proxy/https_proxy is specified

As described in [OSD-16103](https://issues.redhat.com//browse/OSD-16103)

Basically, some clusters have a trust bundle specified, but no HTTP_PROXY/HTTPS_PROXY, so osdctl wasn't prompting SREs to supply an additional trust bundle when it was actually required.

Also renaming `SubnetId` --> `SubnetIds` for clarity because it's an array now


mjlshen:
Had to rebase in the changes from #375

PR #377, opened 2023-04-22
Initialize empty maps at start of Run for servicelog post

AlexVulaj:
When called from external functions, servicelogs would successfully send but reporting would fail. This is because the maps that we use to hold results for failed/successful posts would be nil at the time where we tried to populate them, causing a panic. This change will make sure that the maps are always initialized.


cmd/servicelog/post.go
```diff
@@ -75,6 +75,12 @@ func newPostCmd() *cobra.Command {
 }
 
 func (o *PostCmdOptions) Init() error {
+	if o.successfulClusters == nil {
+		o.successfulClusters = make(map[string]string)
+	}
+	if o.failedClusters == nil {
+		o.failedClusters = make(map[string]string)
+	}
```
mjlshen, line 83: Since the function initializes the struct should it just always initialize?

```suggestion
  o.successfulClusters = make(map[string]string)
	o.failedClusters = make(map[string]string)
```

AlexVulaj:
@typeid @mjlshen I can answer both of your questions at the same time actually 😄 

Yes, we can remove the initialization of `successfulClusters` and `failedClusters` on lines 49/50. Then, `Init` no longer needs to worry about overwriting existing maps, and thus the `nil` checks are no longer needed. I'll push an update now.

mjlshen (approved):
/lgtm

PR #378, opened 2023-04-24
Deprecate federatedrole apply

AlexVulaj:
After the work done in https://issues.redhat.com/browse/OSD-15564 to move this functionality into AAO, this command is no longer needed.


iamkirkbater (changes requested):
Can we add a `return nil` to the top of the `run()` function to make this a noop?  I think the `Deprecated` flag here will just show a message, but still try to run through the script.

iamkirkbater:
/lgtm
/approve

/hold

While I approve and think it looks good, I do want to allow some time for further discussion on the points of whether we should keep this command in a deprecated state or remove it entirely.

If we remove it entirely we automatically break anything that may have this command as a dependency. While _this specific_ command might not have an impact, I personally think that we should still practice good semver hygiene and attempt to be good maintainers and provide stability.

With that said, if the majority thinks we should just remove the command entirely now I won't be the one to hold it up.

mjlshen:
I'm a +1 to deleting and accounting for it in the semver

PR #379, opened 2023-04-25
Update env command to use new backplane API

NautiluX:
* Use new backplane API and remove tunnels
* Include the requested PS1 command
* Add basic explanation about how to set up the PS1 and use `osdctl env`

jira: [OSD-16138](https://issues.redhat.com//browse/OSD-16138)


mjlshen (commented):
/approve

PR #386, opened 2023-05-11
Print correct splunk index for stg clusters

abyrne55:
Currently, the Splunk URL generated by `osdctl cluster context` sets the value of the `index` parameter to "openshift_managed_audit"; this value, however, is only valid for clusters in OCM's production environment. This PR adds logic for generating the correct value of `index` for staging clusters ("openshift_managed_audit_stage"). It also skips Splunk URL generation for clusters in integration, as these clusters don't forward their logs to Splunk AFAICT.

This PR implements [OSD-15169](https://issues.redhat.com/browse/OSD-15169), which calls for this fix to the above-mentioned bug.

/assign @AlexVulaj

### DIFF
--- a/cmd/cluster/context.go
+++ b/cmd/cluster/context.go
@@ -63,6 +63,10 @@ type contextData struct {
 	ClusterName    string
 	ClusterVersion string
 	ClusterID      string
+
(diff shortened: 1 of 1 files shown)


mjlshen (approved):
/lgtm

PR #388, opened 2023-05-16
Enhance internal service log output

geowa4:
In a recent meeting we discussed enhancing internal service log output. Currently, we only see _that_ an internal service log has been set via the hard coded summary. This change uses the first line of the description to provide more information. Abbreviated output is shown below to demonstrate the effect of this change. 

```
$ ./osdctl cluster context -S a-test-cluster --output short

Version             Supported?          SLs (last 30 d)     Jira Tickets        Current Alerts      Historical Alerts (last 30 d)
4.12.15             true                1 (1 internal)      0                   H: 0 | L: 0         N/A
```

```
$ ./osdctl cluster context -S a-test-cluster

============================================================
Service Logs sent in the past 30 Days
============================================================
0. INT # This is a test description (2023-05-16T14:06:55Z)

```

### DIFF
--- a/cmd/cluster/context.go
+++ b/cmd/cluster/context.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"math"
 	"os"
 	"os/exec"
 	"sort"
@@ -261,6 +262,13 @@ func (o *contextOptions) printShortOutput(data *contextData) {
 		historicalAlertsString = fmt.Sprintf("%d", historicalAlertsCount)
 	}
 
+	var numInternalServiceLogs int
+	for _, serviceLog := range data.ServiceLogs {
+		if serviceLog.InternalOnly {
+			numInternalServiceLogs++
+		}
+	}
+
 	table := printer.NewTablePrinter(os.Stdout, 20, 1, 2, ' ')
 	table.AddRow([]string{
 		"Version",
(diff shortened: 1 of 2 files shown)


cmd/cluster/context.go
```diff
@@ -726,13 +734,26 @@ func printServiceLogs(serviceLogs []sl.ServiceLogShort, verbose bool, sinceDays
...
 		for i, errorServiceLog := range serviceLogs {
-			fmt.Printf("%d. %s (%s)\n", i, errorServiceLog.Summary, errorServiceLog.CreatedAt.Format(time.RFC3339))
+			var serviceLogSummary string
+			if errorServiceLog.InternalOnly {
+				internalServiceLogLines := strings.Split(errorServiceLog.Description, "\n")
+				if len(internalServiceLogLines) > 0 {
+					// if the description is "", Split returns []
+					serviceLogSummary = fmt.Sprintf("INT %s", internalServiceLogLines[0])
+				} else {
+					serviceLogSummary = errorServiceLog.Summary
```
typeid, line 750: **Question:** Would it make sense to also format internal service logs without description in a way that we know it's an internal SL? Also, do you know if service logs without a description exist?

geowa4: I suppose you could have an internal service log without any meaningful message but osdctl will make you work for it. We could change the internal template to have the "INT" instead of printing it out here, but then the older ones wouldn't get formatted. Also, for the short form presented by cluster context, I like the indicator that it's an internal service log to be short. Other formats might prefer longer formats if they have more space. 

```
❯ osdctl servicelog post $CLUSTER_ID --internal
ERRO[0000] The one of the template files is using '${MESSAGE}' parameter, but '--param' flag is not set for this one. Use '-p MESSAGE="FOOBAR"' to fix this.
FATA[0000] Please define this missing parameter properly.
```

cmd/cluster/context.go
```diff
@@ -726,13 +734,26 @@ func printServiceLogs(serviceLogs []sl.ServiceLogShort, verbose bool, sinceDays
...
-		dump.Pretty(os.Stdout, marshalledSLs)
+		_ = dump.Pretty(os.Stdout, marshalledSLs)
 	} else {
 		// Non verbose only prints the summaries
 		for i, errorServiceLog := range serviceLogs {
-			fmt.Printf("%d. %s (%s)\n", i, errorServiceLog.Summary, errorServiceLog.CreatedAt.Format(time.RFC3339))
+			var serviceLogSummary string
+			if errorServiceLog.InternalOnly {
+				internalServiceLogLines := strings.Split(errorServiceLog.Description, "\n")
+				if len(internalServiceLogLines) > 0 {
```
AlexVulaj, line 746: Do we need to split here instead of just checking for `""` ?

PR #392, opened 2023-06-07
Forego grace period and force proxy to be set

geowa4:
Due to timelines for requiring the proxy, we're going to require the proxy configuration for AWS immediately.

[OSD-15754](https://issues.redhat.com//browse/OSD-15754)

### DIFF
--- a/cmd/account/mgmt/account-unassign_test.go
+++ b/cmd/account/mgmt/account-unassign_test.go
@@ -21,8 +21,7 @@ import (
 )
 
 func TestAssumeRoleForAccount(t *testing.T) {
-	viper.Set(awsInternal.ProxyConfigKey, "")
-	viper.Set(awsInternal.SkipProxyCheckKey, true)
+	viper.Set(awsInternal.ProxyConfigKey, "a-random-proxy")
 	mocks := setupDefaultMocks(t, []runtime.Object{})
 
 	mockAWSClient := mock.NewMockClient(mocks.mockCtrl)
--- a/pkg/provider/aws/client.go
+++ b/pkg/provider/aws/client.go
@@ -45,7 +45,6 @@ type AwsClientInput struct {
 }
 
 const ProxyConfigKey = "aws_proxy"
-const SkipProxyCheckKey = "skip_aws_proxy_check"
 
 // TODO: Add more methods when needed
 type Client interface {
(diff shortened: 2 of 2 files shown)


pkg/provider/aws/client.go
```diff
@@ -140,15 +139,13 @@ type AwsClient struct {
...
-		_, _ = fmt.Fprintf(os.Stderr, "[WARNING] `%s` not configured. Please add this to your osdctl configuration to ensure traffic is routed though a proxy.\n", ProxyConfigKey)
-		if skipAwsProxyCheck := viper.GetBool(SkipProxyCheckKey); !skipAwsProxyCheck {
-			_, _ = fmt.Fprint(os.Stderr, "Please confirm that you would like to continue with [y|N] ")
-			var input string
-			_, _ = fmt.Scanln(&input)
-			if strings.ToLower(input) != "y" {
-				_, _ = fmt.Fprintln(os.Stderr, "Must enter 'y' to continue; exiting...")
-				os.Exit(0)
-			}
+		_, _ = fmt.Fprintf(os.Stderr, "[ERROR] `%s` not configured. Please add this to your osdctl configuration to ensure traffic is routed though a proxy.\n", ProxyConfigKey)
```
AlexVulaj, line 142: This still feels more like a warning to me, as we aren't crashing or hard-enforcing anything in the absence of an `aws_proxy`. I'm not fully opposed, but what made you up the severity here?

cmd/account/mgmt/account-unassign_test.go
```diff
@@ -21,8 +21,7 @@ import (
 )
 
 func TestAssumeRoleForAccount(t *testing.T) {
-	viper.Set(awsInternal.ProxyConfigKey, "")
-	viper.Set(awsInternal.SkipProxyCheckKey, true)
+	viper.Set(awsInternal.ProxyConfigKey, "a-random-proxy")
```
AlexVulaj, line 24: Is this left over from testing? Either way I think we should default to empty-string still.

pkg/provider/aws/client.go
geowa4: In my head, warnings are often allowed to accumulate. Errors must be dealt with, which I'm demanding the operator does in the next couple lines of code. But I also don't really mind either way.

cmd/account/mgmt/account-unassign_test.go
AlexVulaj: _sigh_ That's my bad. I totally missed that this was the test file. Carry on.

PR #398, opened 2023-06-23
Make 'ConfirmSend' return a bool instead of an error (and rename it)

AlexVulaj:
Currently:
```
$ ./osdctl servicelog post -t <template_id> <cluster_id>
...
INFO[0009] The following template will be sent:         
...
Continue? (y/N): N
FATA[0003] Exiting...
```

Outputting a fatal error message on a negative confirmation seems like something crashed. With this new change:
```
$ ./osdctl servicelog post -t <template_id> <cluster_id>
...
INFO[0009] The following template will be sent:         
...
Continue? (y/N): N
```

Also on the side - I came across a situation on accident where the version check logic would seemingly stall out if the user input anything other than "y" or "n". For consistency in UX, I updated it to use the same logic as everywhere else.


mjlshen (approved):
/lgtm

PR #399, opened 2023-06-23
Let erroneous input fall through to the default switch case

AlexVulaj:
Simplify the logic of ConfirmPrompt a bit - since an error in `Scanln` is handled the same as the `default` case in the switch statement, just let it fall through to there.


fahlmant (approved):
/approve
/lgtm

PR #401, opened 2023-06-23
Adds concurrency to cluster context command

clcollins:
Adds concurrent go connections to retrieve data for the cluster context command.  In testing, this sped up retrieval and printing, shaving 6-12s from the overall time required to gather cluster context. It's even more impactful with the --full flag, queuing up the time-consuming AWS stuff right away.

### DIFF
--- a/cmd/cluster/context.go
+++ b/cmd/cluster/context.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	pd "github.com/PagerDuty/go-pagerduty"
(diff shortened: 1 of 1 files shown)


typeid:
**Question**: what prevents the function from returning before all the data is queried? In the current setup, I would assume we just start all the coroutines and return immediately? 

See my go playground example:
https://go.dev/play/p/QBqHvB_0yFe

AlexVulaj:
My usual stance is to avoid multithreading unless there's a noticeable performance issue. I'm not opposed to this change, but there are always more risks and maintenance woes with concurrency - were we observing any issues with this command's response times?

AlexVulaj:
Another question I have: will the output be in a different order every time now potentially? I don't think this is a problem to be clear, but I'm curious regardless.

mjlshen (approved):
/lgtm

PR #403, opened 2023-06-29
Fixes PD retrieval bug; clean up long format output; other fixes

clcollins:
This PR fixes a bug in the pager duty alert retrieval for the cluster
context command.

It also fixes a duplicate printing of the limited support reason where
support exceptions should be listed.

It also cleans up the long format output:

* Changes delimiters for more compact output (fit more on-screen)
* standardizes line breaks
* removes "not polling for cloudtrail" message
* cleans up "external data" links
* replaces "Cluster Info" generic head with cluster name header, as
  short-form output uses

Finally, this PR addresses a handful of spelling mistakes and renames
some variables to Golang convention (and helps with cSpell).

Cleans up the formatting for external links, standardizes printing


cmd/cluster/context.go
```diff
@@ -201,46 +201,40 @@ func (o *contextOptions) run() error {
...
+		printHistoricalPDAlertSummary(data.HistoricalAlerts, data.pdServiceID, o.days)
+		fmt.Println()
 
-		err := printCloudTrailLogs(data.CloudtrailEvents)
-		if err != nil {
-			_, _ = fmt.Fprintf(os.Stderr, "Can't print cloudtrail: %v\n", err)
-			os.Exit(1)
-		}
-	} else {
+		printCloudTrailLogs(data.CloudtrailEvents)
```
tnierman, line 225: Is there a reason we're no longer handling the error returned by `printCloudTrailLogs`?

cmd/cluster/context.go
```diff
@@ -201,46 +201,40 @@ func (o *contextOptions) run() error {
...
-
+	fmt.Println()
+	printJiraIssues(data.JiraIssues)
+	fmt.Println()
 	printCurrentPDAlerts(data.PdAlerts, data.pdServiceID)
+	fmt.Println()
 
 	if o.full {
-		_ = printHistoricalPDAlertSummary(data.HistoricalAlerts, data.pdServiceID, o.days)
+		printHistoricalPDAlertSummary(data.HistoricalAlerts, data.pdServiceID, o.days)
```
tnierman, line 222: Is there a reason we didn't handle the error from this function in the past? If not, I think we should start. If so, I think not intentionally assigning the output to the blank identifier might trigger gosec's compliance check?

clcollins: That's fair - I don't know why it was being ignored before.  The only possible error to be returned was a table.Flush() error, but we can still catch and handle that.

cmd/cluster/context.go
clcollins: This was also a table.Flush() error.  I'll update this as well.

cmd/cluster/context.go
clcollins: Alright, since the error return for each of the print* functions was just output from `table.Flush()`, I've changed all of them to be consistent with one another.  They will:

* no longer return errors, but
* if an error is encountered, print it to stderr, so the SRE will see the action-specific error printed under the action header, or, if sending stderr elsewhere, see the errors collected, with the action name + the error it encountered.

This should standardize how the errors are handled, raise them to the SREs with appropriate context, and remove the need to handle any return from the various printer functions.

cmd/cluster/context.go
clcollins: See the comments for the related change above.

cmd/cluster/context.go
tnierman: Agreed, it makes the most sense to complete the context evaluation even if an error is encountered rather than bail early and potentially miss out on important information :+1:

PR #405, opened 2023-07-10
Add flag to bypass aws_config value

AlexVulaj:
Adds a flag to bypass the `aws_config` value set in osdctl's config. The flag purposely does not provide a shorthand since we should be extremely deliberate in using this flag.

This PR comes as a result of this thread:
https://redhat-internal.slack.com/archives/CNJPJSG3Z/p1688745686279219

Example run:
```
$ ./osdctl account console --clusterID abc123 -p my_aws_profile --no-aws-proxy
Not adding proxy to AWS client due to presence of no-aws-proxy flag
The AWS Console URL is:
<console url>
```

I've confirmed that this will work off of the VPN as well.

### DIFF
--- a/pkg/provider/aws/client.go
+++ b/pkg/provider/aws/client.go
@@ -44,7 +44,10 @@ type AwsClientInput struct {
 	Region          string
 }
 
-const ProxyConfigKey = "aws_proxy"
+const (
+	ProxyConfigKey = "aws_proxy"
+	NoProxyFlag    = "no-aws-proxy"
+)
 
 // TODO: Add more methods when needed
 type Client interface {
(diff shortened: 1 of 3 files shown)


pkg/provider/aws/client.go
```diff
@@ -44,7 +44,10 @@ type AwsClientInput struct {
 	Region          string
 }
 
-const ProxyConfigKey = "aws_proxy"
+const (
+	ProxyConfigKey = "aws_proxy"
+	NoProxyFlag    = "no-aws-proxy"
```
dustman9000, line 49: Suggestion for consistency for other skip flags.
```suggestion
	NoProxyFlag    = "skip-aws-proxy-check"
```

PR #406, opened 2023-07-10
Enforce that payer accounts cannot be assigned to individual users

AlexVulaj:
This will prevent osdctl from assigning users to payer accounts regardless of the structure of the parent OUs. I use the same client that is initially set up with the provided payer account to call `GetCallerIdentity` and pull the account ID off of the result. Then we can make sure that an account with that ID is passed over during the assignment checks.

As for tests, I added a few cases and cleaned up nil-values in existing test cases.

I also reformatted some deprecated code using the `rand` package that I came across.

Resolves [OSD-16129](https://issues.redhat.com//browse/OSD-16129)

### DIFF
--- a/cmd/account/mgmt/account-assign.go
+++ b/cmd/account/mgmt/account-assign.go
@@ -2,7 +2,7 @@ package mgmt
 
 import (
 	"fmt"
-
+	"github.com/aws/aws-sdk-go/service/sts"
 	"math/rand"
 	"time"
 
@@ -214,8 +214,18 @@ func (o *accountAssignOptions) findUntaggedAccount(rootOu string) (string, error
 		return "", ErrNoUntaggedAccounts
 	}
 
+	identity, err := o.awsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	if err != nil {
+		return "", err
+	}
+
 	// Loop through accounts and check that it's untagged and assign ID to user
 	for _, a := range accounts.Accounts {
+		if *a.Id == *identity.Account {
+			// Don't allow the payer account to be assigned to an individual user
+			continue
+		}
+
 		isOwned, err := isOwned(*a.Id, &o.awsClient)
 		if err != nil {
 			return "", err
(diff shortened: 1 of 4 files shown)


cmd/account/mgmt/account-assign.go
```diff
@@ -214,8 +213,18 @@ func (o *accountAssignOptions) findUntaggedAccount(rootOu string) (string, error
 		return "", ErrNoUntaggedAccounts
 	}
 
+	identity, err := o.awsClient.GetCallerIdentity(nil)
```
mjlshen, line 216:
```suggestion
	identity, err := o.awsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
```
I think this is more clear, though the behavior should be identical

cmd/account/mgmt/account-assign.go
```diff
@@ -214,8 +213,18 @@ func (o *accountAssignOptions) findUntaggedAccount(rootOu string) (string, error
...
+	if err != nil {
+		return "", err
+	}
+
 	// Loop through accounts and check that it's untagged and assign ID to user
 	for _, a := range accounts.Accounts {
+		if *a.Id == *identity.Account {
```
mjlshen, line 223: My understanding of what this code does is that if you use AWS credentials for a specific AWS account, you cannot assign the same AWS account to yourself.

I think this generally works I think if SRE's are using AWS Profiles as intended, however I think the goal of this story is better suited as AWS-side RBAC vs client-side restrictions - special accounts should not allow SREs to tag them with:

```go
			{
				Key:   aws.String("owner"),
				Value: aws.String(o.username),
			},
```

mjlshen (changes requested):

cmd/account/mgmt/account-assign.go
AlexVulaj: The problem we run into is that for reasons out of our control, this _has_ happened in the past (see the ticket in the description). At the end of the day, we don't have full control over those AWS accounts, the OUs they belong to, their tags, etc. The decision to take on this effort was to control what we can control and prevent things as much as we can on our end.

mjlshen: I'm ok with merging this as a hack, but we do have full control over these accounts - they're the organization root accounts

AlexVulaj: More context here for any others following along:
https://redhat-internal.slack.com/archives/CNJPJSG3Z/p1689014783004559?thread_ts=1689003726.494479&cid=CNJPJSG3Z

AlexVulaj: I've created OSD-17462 as a follow up card as it will affect a different repository. In the mean time I'm in favor of merging this PR.

iamkirkbater: > I'm ok with merging this as a hack, but we do have full control over these accounts

@mjlshen - I think this is the exact complication that we have had with this account - EVERYONE has full access over these accounts, I think even some external to SD might have some billing and OU management access, so even if we modify our practices to say "don't do anything with this account" it can still get modified either intentionally or accidentally.

> could we tag the payer account with an owner tag as well? Just for safety?

@fahlmant - This is the exact approach we've already taken.

So the intention here is not so much just a hack in order to be the only "defense" against this account being assigned to a dev, but this allows us to also have a technical non-process related control against the same in case somehow the account gets moved and/or untagged accidentally or on purpose again. This change is meant to be just another layer of protection against us shooting ourselves in the foot.

fahlmant (approved):
/lgtm
/approved

PR #408, opened 2023-07-13
[OSD-17447] Gather hyperinfo from all associated AWS accounts

bergmannf:
This is a PR that will attempt to fetch resources from all AWS accounts associated with a hypershift accounts.

This is mostly, because it is really annoying to log into all those accounts one-by-one and finding if everything is there, so this should gather at least of the resources and print them in table form (and graphviz dot notation, as the tables are quite big).

Still in draft to get some feedback if this is considered useful.


cmd/hypershift/cmd.go
```diff
@@ -0,0 +1,20 @@
+package hypershift
+
+import (
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+// NewCmdAao implements the base aao command
```
clcollins, line 8: I think this comment is a copy/paste remnant :D

cmd/hypershift/cmd.go
```diff
@@ -0,0 +1,20 @@
...
+func NewCmdHypershift(streams genericclioptions.IOStreams) *cobra.Command {
```
clcollins, line 9: Is this intended to be `osdctl hypershift` for the command?  

If so, I think this would be better as a noun-verb subcommand for `cluster`, command, given the thing you're going for here.  Maybe something like `osdctl cluster get-cloud-resources` or something along those lines.

If it can be generic-ish, like this, someone else can do something similar for non-hypershift clusters.  I think there's benefits for both.  Obv not needing to gather management cluster info, etc, but getting the resources themselves would be an excellent advantage for both types!

cmd/hypershift/info.go
```diff
@@ -0,0 +1,691 @@
...
+	awsSessions, err := i.getAWSSessions(clusters)
+	if err != nil {
+		log.Fatal("Could not construct all AWS sessions: ", err)
+		return err
+	}
+	wg := sync.WaitGroup{}
+	wg.Add(3)
+
+	ai := aggregateClusterInfo{}
+	eC := make(chan error)
```
clcollins, line 162: Is there a benefit to using the channels to collect errors?  Not opposed, just curious on the thinking, for my own education.

cmd/hypershift/info.go
```diff
@@ -0,0 +1,691 @@
...
+	}
+	return nil
+}
+
+func baseApiUrl(c *v1.Cluster) string {
+	return strings.Replace(strings.Replace(strings.Replace(c.API().URL(), "https://api.", "", 1), c.Name(), "", 1), ":443", "", 1)
+}
+
+func (i *infoOptions) run() error {
+	log.Println("Getting hypershift info for cluster: ", i.clusterID)
```
clcollins, line 142: Would you be opposed to wrapping informational messages like these in a check for the verbose global flag? By default less is more, especially to allow folks to pipe output to other things.

cmd/hypershift/info.go
```diff
@@ -0,0 +1,691 @@
...
+	clusters, err := i.getClusters()
+	if err != nil {
+		return err
+	}
+	if i.awsRegion == "" {
+		i.awsRegion = clusters.customerCluster.Region().ID()
+		log.Println("Set aws region to: ", i.awsRegion)
+	}
+
+	log.Println("Constructing AWS sessions for all accounts")
```
clcollins, line 152: Is there a reason for using the `log` package here instead of `fmt`?  Most of the rest of osdctl isn't using log output, and while I'm not opposed, I'm not sure what the benefit of timestamps on a manually run too is.  It's valuable for actual log files if you're looking back at the history, but I think it just adds noise to the output for a CLI tool.

EDIT: Nevermind, looks like we're using log in a lot of places!  I take back my comment.

pkg/provider/aws/client.go
```diff
@@ -94,6 +101,9 @@ type Client interface {
 	DescribeRouteTables(*ec2.DescribeRouteTablesInput) (*ec2.DescribeRouteTablesOutput, error)
 	DescribeSubnets(*ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error)
 	DescribeVpcs(*ec2.DescribeVpcsInput) (*ec2.DescribeVpcsOutput, error)
+	DescribeVpcEndpoints(*ec2.DescribeVpcEndpointsInput) (*ec2.DescribeVpcEndpointsOutput, error)
```
clcollins, line 104: Just a note that there's an effort that's almost completed to port the AWS stuff to the v2 SDK: https://github.com/openshift/osdctl/pull/411

It might be worth updating this PR to that before merging.

clcollins (changes requested):
I love this PR - being able to visualize the resources for a cluster will be a great addition, especially for troubleshooting!

Just a couple small requests or requests for info for my own edification.

cmd/hypershift/cmd.go
bergmannf: Yes I totally agree, I was also thinking that having a similar command (or this command with the right flags) provide similar output for both - given that one of the cluster's that should be collected is the customer cluster anyhow that whole part should be 90% reusable (maybe except the VCP endpoints).

cmd/hypershift/info.go
bergmannf: Yea I guess I just looked at one of the commands that does, but I agree with the 'timestamps are not that useful' sentiment in this case - so I'm fine switching to just fmt printing, so we don't add more log code if we don't really want it.

cmd/hypershift/info.go
bergmannf: I think it's one of two ways (at least that I know of), to indicate an error during execution of the goroutine.
As the goroutine doesn't return a value at all, there must be another way to check for an error to have occurred.

I think the alternative would be use a single channel to also gather the output and use an aggregated type that wraps the expected return value & an error, so once the value is retrieved the caller can check if an error occurred:

```
type retrievable interface {
	clusterInfo | managementClusterInfo | privatelinkInfo
}

type ChanReturn[T retrievable] struct {
	Value T
	Error error
}
```

Then just use a `chan[ChanReturn]` and pass that to all async functions.

If there is a more synchronous-looking/behaving way I'd be happy to use that - the current mix between output-argument passed in and channel is quite ugly.

clcollins: Thank for the explanation!  This sounds good to me, and I didn't have any issues with it - just trying to learn more myself, thanks!

mjlshen (approved):
/approve

AlexVulaj:
/hold

Want to make sure this gets merged after the aws-sdk-v2 migration that should be merged very soon

cmd/cluster/graphviz.go
AlexVulaj, line 1: Can we move this file out of `cmd` since it doesn't hold a command struct? I also think it could have more general usage in the future. I'm thinking somewhere in `pkg`

cmd/cluster/hypershift_info.go
```diff
@@ -0,0 +1,773 @@
...
+
+// newCmdPool gets the current status of the AWS Account Operator AccountPool
+func NewCmdHypershiftInfo(streams genericclioptions.IOStreams) *cobra.Command {
+	ops := newInfoOptions(streams)
+	infoCmd := &cobra.Command{
+		Use:   "hypershift-info CLUSTERID",
+		Short: "Pull information about AWS objects from the cluster, the management cluster and the privatelink cluster",
+		Long: `This command aggregates AWS objects from the cluster, management cluster and privatelink for hypershift cluster.
+It attempts to render the relationships as graphviz if that output format is chosen or will simply print the output as tables.`,
+		Args: func(cmd *cobra.Command, args []string) error {
```
AlexVulaj, line 41: Could we shorten this to `cobra.ExactArgs(1)` ?

cmd/cluster/graphviz.go
bergmannf: I moved the rendering part out.

Generating the 'graph' structure is now in the hypershift_info.go file, because otherwise there would be a circular dependency - so every code that wants to render graphviz has to generate the graph and just pass it to the render function.

AlexVulaj: Love that approach!

AlexVulaj:
/lgtm

confirmed my last question with @bergmannf here: https://redhat-internal.slack.com/archives/CNJPJSG3Z/p1692189668859019

PR #409, opened 2023-07-13
OSD-15543 - Add 'package' subcommand to 'promote' commandset

tnierman:
https://issues.redhat.com/browse/OSD-15543

---

Adds logic to osdctl enabling package operator packages to be promoted via app-interface


mjlshen (approved):
/lgtm

PR #411, opened 2023-07-17
Migrate to aws-sdk-v2

AlexVulaj:
This PR migrates away all use of the old go aws sdk to the new v2 sdk.

Addresses [OSD-17454](https://issues.redhat.com//browse/OSD-17454)


go.mod
```diff
@@ -6,10 +6,18 @@ require (
 	cloud.google.com/go/compute v1.19.3
 	github.com/PagerDuty/go-pagerduty v1.5.1
 	github.com/andygrunwald/go-jira v1.16.0
-	github.com/aws/aws-sdk-go v1.44.110
```
AlexVulaj, line 9: 👋🏻

pkg/provider/aws/iam.go
```diff
@@ -25,17 +27,19 @@ func CreateIAMUserAndAttachPolicy(awsClient Client, username, policyArn *string)
...
 		}
 		return false, err
 	}
 	return true, nil
 }
 
+func Foo() int {
+	return 1
+}
+
```
mjlshen, line 42: typo?

cmd/account/console.go
```diff
@@ -168,8 +166,8 @@ func (o *consoleOptions) run() error {
 	consoleURL, err := aws.RequestSignInToken(
 		awsClient,
 		&o.consoleDuration,
-		awsSdk.String(sessionName),
-		awsSdk.String(targetRoleArn.String()),
+		&sessionName,
+		pointer.String(targetRoleArn.String()),
```
mjlshen, line 170: It's technically the same thing, but I'm wondering if it makes more sense for this to be `awsSdk.String` where `awsSdk` is now `github.com/aws/aws-sdk-go-v2/aws`

mjlshen (changes requested):

pkg/provider/aws/client.go
```diff
@@ -166,344 +159,326 @@ func addProxyConfigToSessionOptConfig(config *aws.Config) {
...
 	if _, err := awsClient.GetCallerIdentity(nil); err != nil {
-		if aerr, ok := err.(awserr.Error); ok {
-			switch aerr.Code() {
-			case "InvalidClientTokenId":
-				return nil, fmt.Errorf("`aws sts get-caller-identity --profile %s --region %s` failed - "+
-					"ensure the credentials in the profile are valid and the region is in the AWS partition you expect", profile, region)
-			default:
-				return nil, err
-			}
+		var iite *stsTypes.InvalidIdentityTokenException
```
mjlshen, line 212: InvalidIdentityTokenException != InvalidClientTokenId

pkg/provider/aws/iam.go
AlexVulaj: 😆 this was me testing something with my local config. good catch.

pkg/provider/aws/client.go
AlexVulaj: The new sts SDK doesn't seem to have an error type of "InvalidClientTokenId", and this was the only error I could find relating to tokens at all so I assumed it was what they renamed it to. The doc for InvalidIdentityTokenException reads:
```
// The web identity token that was passed could not be validated by Amazon Web
// Services. Get a new identity token from the identity provider and then retry the
// request.
```

Any thoughts?

pkg/osdCloud/aws.go
```diff
@@ -335,12 +336,13 @@ func (a *AwsCluster) GetAZs() []string {
 	return a.AZs
 }
 
-func (a *AwsCluster) GetAllVirtualMachines(region string) ([]VirtualMachine, error) {
+func (a *AwsCluster) GetAllVirtualMachines(string) ([]VirtualMachine, error) {
```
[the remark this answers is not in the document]
AlexVulaj: There's an interface in pkg/osdCloud/interfaces that AwsCluster is trying to satisfy. The region is needed for GCP clusters, so I decided to leave it in the interface and just make it unused (explicitly) for the AWS impl.

cmd/account/console.go
AlexVulaj: the v2 sdk got rid of that package entirely, so I just made these all `pointer.<>` everywhere applicable.

AlexVulaj: Sorry, my first comment misread this.

I'm okay with either, but my preference was to use the standard lib for an operation like this. To compare the functions' source code:

```
AWS SDK V2:
func String(v string) *string {
	return ptr.String(v)
}
// calls down to...
func String(v string) *string {
	return &v
}
```

```
Std Lib:
func String(s string) *string {
	return &s
}
```

So really the same thing is going on under the hood. I'll note that if you (or anyone else 😄 ) feel strongly one way or the other, it'd be extremely easy to do a quick find/replace for `pointer.` through this PR, just let me know!

mjlshen: Yeah, if it was in the stdlib, I'd be all for it haha. More that it's been changed to use `"k8s.io/utils/pointer"`?

pkg/provider/aws/client.go
mjlshen: Yeah, AWS doesn't have available types for all of their errors, in this case the two errors are not the same. `InvalidClientTokenId` is a common error to STS (https://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html#API_GetCallerIdentity_Errors) and can happen during `aws sts get-caller-identity` if the AWS Access Key ID is invalid, but `InvalidIdentityToken` specifically only happens during `aws sts assume-role-with-web-identity` https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html#API_AssumeRoleWithWebIdentity_Errors

cmd/account/console.go
AlexVulaj: This is a miss by me actually - for some reason I thought this was the `golang.org/x/tools/go/pointer` package 🤦🏻 

Let me revert all of these to use the aws sdk.

pkg/provider/aws/client.go
AlexVulaj: Thanks for the insight. In this case, I'm going to drop the special handling of this particular alert and return the error through from the AWS SDK.

pkg/provider/aws/client.go
```diff
@@ -166,344 +157,319 @@ func addProxyConfigToSessionOptConfig(config *aws.Config) {
...
-		if aerr, ok := err.(awserr.Error); ok {
-			switch aerr.Code() {
-			case "InvalidClientTokenId":
-				return nil, fmt.Errorf("`aws sts get-caller-identity --profile %s --region %s` failed - "+
-					"ensure the credentials in the profile are valid and the region is in the AWS partition you expect", profile, region)
-			default:
-				return nil, err
-			}
-		}
+		return nil, fmt.Errorf("error getting caller identity: %w", err)
```
mjlshen, line 208: Emphasis that this changed, personally I'm ok with it, but so other reviewers are aware

mjlshen (approved):
/approve

pkg/provider/aws/client.go
AlexVulaj: For context - https://github.com/openshift/osdctl/pull/411#discussion_r1268674566

mjlshen (approved):
/lgtm

PR #412, opened 2023-07-17
OSD-17508 Automatically send a service log after resizing infra nodes

yithian:
Once the infra nodes are successfully resized, send a service log message to notify the customer that the work was completed. This is a step in the SOP and is being included in the resizing command for ease of use.


yithian:
I tested this against osd-v4stg-aws successfully:

image

mjlshen (approved):
/approve

PR #415, opened 2023-07-18
automatically supply network verify-egress CA Bundle

mjlshen:
This commit automatically retrieves the additional trust bundle from Hive from its proxy SyncSet when needed instead of requiring users to manually obtain and supply it. We retrieve it from Hive instead of from the cluster itself to handle cases during cluster provisioning where it would not be accessible on the cluster yet.

OSD-17510


pkg/k8s/client.go
```diff
@@ -61,6 +61,10 @@ func (s *LazyClient) Status() client.StatusWriter {
 	return s.getClient().Status()
 }
 
+func (s *LazyClient) SubResource(subResource string) client.SubResourceClient {
```
mjlshen, line 64: This is just a side effect of satisfying the updated interface in controller-runtime v0.14.6

cmd/network/verification.go
```diff
@@ -464,6 +479,85 @@ func (e *EgressVerification) getSecurityGroupId(ctx context.Context) (string, er
...
+	// Register hivev1 for SelectorSyncSets
+	if err := hivev1.AddToScheme(scheme); err != nil {
+		return "", err
+	}
+
+	hive, err := utils.GetHiveCluster(e.cluster.ID())
+	if err != nil {
+		return "", err
+	}
+
```
typeid, line 502: **thought:** This is unfortunately now a difference in the effectivity of CAD running network verifier and manually running it. 

FYI @RaphaelBut

typeid (commented):
Apart from the failing unit tests for minor reasons and the lint job failing, it's looking good to me :) Thanks for adding this to once again make it much easier to run network verification.

cmd/network/verification.go
typeid: AFAIK - we don't. Right now, CAD just runs the network verifier pretty much in the same way osdctl did up until this point. We won't be able to get the CA as we have no hive access, unless it's saved in OCM as well?

cmd/network/verification.go
```diff
@@ -464,6 +482,85 @@ func (e *EgressVerification) getSecurityGroupId(ctx context.Context) (string, er
...
+	}})
+	if err != nil {
+		return "", err
+	}
+
+	if err := hc.List(ctx, nsList, &client.ListOptions{LabelSelector: clusterSelector}); err != nil {
+		return "", err
+	}
+
+	if len(nsList.Items) != 1 {
```
AlexVulaj, line 525: Extremely minor - we could save a `len` calculation/lookup here with a short statement and then reusing that in the error output.

PR #418, opened 2023-07-20
[OSD-16640] Updated the osdctl cluster health commend to work for hypershift clusters

aliceh:
In this commit, I updated the osdctl cluster health command to output
nodepools, and pod information from the management cluster for a hosted hypershift cluster ID.

https://issues.redhat.com/browse/OSD-16640


cmd/cluster/health.go
```diff
@@ -85,101 +102,158 @@ func (o *healthOptions) run() error {
...
-		defer clusterHealthClient.Close()
+
+		healthOutput, err := yaml.Marshal(&healthObject)
+		if err != nil {
+			log.Fatalf("error: %v", err)
+		}
+		fmt.Printf("\n \n")
+		fmt.Println(string(healthOutput))
 	} else {
-		return errors.New(fmt.Sprintf("Unknown cloud provider found: %s", cluster.CloudProvider().ID()))
```
fahlmant, line 123: What about handling GCP?

cmd/cluster/health.go
```diff
@@ -85,101 +102,158 @@ func (o *healthOptions) run() error {
...
-	err = clusterHealthClient.Login()
-	if err != nil {
-		return err
-	}
-	for _, zone := range clusterHealthClient.GetAZs() {
-		instances, err := clusterHealthClient.GetAllVirtualMachines(zone)
+
+		hsHealthObject := createHypershiftHealthObject(cluster)
+
+		//var clusterHealthClient osdCloud.ClusterHealthClient
```
fahlmant, line 204:
```suggestion

```

cmd/cluster/health.go
```diff
@@ -85,101 +102,158 @@ func (o *healthOptions) run() error {
...
+
+		if cluster.Nodes().Compute() != 0 {
+			hsHealthObject.Expected.Worker = int(cluster.Nodes().Compute())
+		}
+		mc := ""
+		if cluster.Hypershift().Enabled() {
+			// Ignorning the error here as this endpoint is behind a specific permissioon
```
fahlmant, line 211: What happens if there is an error? Shouldn't this command exit out since it will get an empty body back?

fahlmant (commented):
@aliceh Can you explain a bit more why the decision is get do an `oc get nodes, oc get pods` for this check? I don't see many details in the OSD card, nor the description. For Classic this command doesn't require any kube api access, so requiring it for the HCP equivalent makes me think this should be a separate command.

aliceh:
@fahlmant me and @mjlshen had a little discussion on what would be useful to output in case of hypershift. The nodepools check is the one that is required in the card.  I always check the pods as well, so I though this could be a useful addition.

fahlmant:
@aliceh 
Can you please place that discussion in the card?
Also, while those may be useful, I still think it needs to be a seperate command than the classic version. The classic command counts the AWS instances vs the OCM records, and no kube api is involved. The HCP version just uses a kube api, assuming you're logged into the correct cluster, and doesn't use the AWS api. I think it might make more sense to diverge here and have an AWS-side check for both HCP and Classic, and an "on-cluster" check for both HCP and classic.

cmd/cluster/health.go
```diff
@@ -194,8 +268,56 @@ func createHealthObject(cluster *v1.Cluster) *ClusterHealthCondensedObject {
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to execute:\n%s", strings.TrimSpace(string(nodepool)))
+	}
+	return nodepool, nil
+}
+func getPods(environment string, clusterID string, clusterName string) ([]byte, error) {
+
+	get_po := fmt.Sprintf("oc get po -n ocm-%v-%v-%v| grep -v \"Running\\|Completed\" ", environment, clusterID, clusterName)
+
+	pods, err := exec.Command("bash", "-c", get_po).CombinedOutput()
```
clcollins, line 306: If we're already using golang, would you be opposed to getting the pods with native go libs from the oc sdk?  The more self-contained command is going to be more predictable than calling out to a bash command with multiple pipes and tons of potential errors generated from outside the go libs.

cmd/cluster/health.go
```diff
@@ -85,101 +102,158 @@ func (o *healthOptions) run() error {
...
+		}
+		err = clusterHealthClient.Login()
 		if err != nil {
 			return err
 		}
-		ownedLabel = "kubernetes-io-cluster-" + infraID
-		defer clusterHealthClient.Close()
-	} else if cluster.CloudProvider().ID() == "aws" {
-		clusterHealthClient, err = osdCloud.NewAwsCluster(ocmClient, o.clusterID, o.awsProfile)
+		for _, zone := range clusterHealthClient.GetAZs() {
```
clcollins, line 147: For each of these requests that reach out (to OCM, via oc get, possibly to the cloud provider) would you be opposed to using a waitGroup (https://pkg.go.dev/sync#example-WaitGroup) to run them all async and then print the results when the last one is done?

There was great success with the cluster-context subcommand in speeding up the time it takes to gather the info.  This one is kinda on the line since there's less calls, but they're all reaching out to a remote api in one way or another, so there should still be a benefit.  It might be nice to incorporate it now so in the long run as this subcommand is expanded, folks will use the waitGroup by default.

Less time for SREs to sit an wait for output :)

clcollins (changes requested):
Generally LGTM!  Thanks for the update.  Just left a couple of comments.

PR #419, opened 2023-07-21
Check subscription status if no match for infra ID

geowa4:
We've seen a recent influx of alerts where the cluster has been deprovisioned before we get a chance to look at them. This change should make that situation apparent.

[OSD-17576](https://issues.redhat.com/browse/OSD-17576)

### DIFF
--- a/cmd/cluster/from_infra_id.go
+++ b/cmd/cluster/from_infra_id.go
@@ -3,6 +3,8 @@ package cluster
 import (
 	"encoding/json"
 	"fmt"
+	sdk "github.com/openshift-online/ocm-sdk-go"
+	amv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/openshift/osdctl/internal/utils/globalflags"
 	"github.com/openshift/osdctl/pkg/utils"
@@ -10,6 +12,7 @@ import (
 	"gopkg.in/yaml.v2"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"os"
+	"sort"
 	"strings"
 )
 
@@ -57,8 +60,8 @@ func (ops *fromInfraIdOptions) run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	defer func() {
-		if err := ocmClient.Close(); err != nil {
-			fmt.Printf("cannot close the ocmClient (possible memory leak): %q", err)
+		if errClose := ocmClient.Close(); errClose != nil {
+			fmt.Printf("cannot close the ocmClient (possible memory leak): %q", errClose)
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


geowa4:
Local tests passed. It looks like there was an issue downloading a dependency.

cmd/cluster/from_infra_id.go
```diff
@@ -79,10 +81,29 @@ func (ops *fromInfraIdOptions) run(cmd *cobra.Command, args []string) error {
 			return renderOutput(cluster, ops.globalOpts.Output)
 		}
 	}
-	_, err = fmt.Fprintf(os.Stderr, "No clusters found matching %s\n", infraId)
+	_, _ = fmt.Fprintf(os.Stderr, "No clusters found matching %s\n", infraId)
```
clcollins, line 84: Would you be opposed to wrapping this in a check for the `verbose` flag?  "Checking for" messages are useful, but implied by the fact that the next function you call will return that context.  Leaning toward UNIX "no news is good news".  

I'll agree that it's less likely with this particular scenario that folks might be piping to another command to do something with the data, but if they did, having informational info behind the verbose flag helps for parsing output.

geowa4: Hm, let me think about that. I like what you're saying. Right now, this prints to stderr so pipelines wouldn't ingest it. Maybe that's sufficient? 🤔

geowa4: Updated

cmd/cluster/from_infra_id.go
```diff
@@ -79,10 +82,38 @@ func (ops *fromInfraIdOptions) run(cmd *cobra.Command, args []string) error {
...
+	subResponse, err := ocmClient.AccountsMgmt().V1().Subscriptions().List().Search(fmt.Sprintf("display_name='%s' and managed='true'", clusterName)).Send()
+	if err != nil {
+		return nil, err
+	}
+	if subResponse.Size() == 0 {
+		return nil, fmt.Errorf("no subscriptions found for cluster name %s", clusterName)
+	}
+	subs := subResponse.Items().Slice()
+	sort.SliceStable(subResponse.Items().Slice(), func(i, j int) bool {
+		return subs[i].UpdatedAt().Before(subs[j].UpdatedAt())
```
AlexVulaj, line 110: Could we instead change this line to `return subs[i].UpdatedAt().After(subs[j].UpdatedAt())`, and then down on line 113 just grab `subs[0]` ?

geowa4: Huh, you highlighted a line with a bug. I'm not using `subs` in `SliceStable`'s first parameter. I'll make both changes while I'm there.

geowa4: Updated

cmd/cluster/from_infra_id.go
geowa4: We chatted offline and my latest push should have sufficiently informative output given the constraint that the display name we're searching on is not guaranteed to be unique.

cmd/cluster/from_infra_id.go
geowa4: Funny enough I ended up adding `len(subs)` to the output message, but I'm going to leave the sorting to allow us to pull the 0 index.

PR #422, opened 2023-07-25
Check for registry credentials in OCM, Part II

geowa4:
This check is standard procedure if the cluster is reported missing.

The first attempt identified an issue with our procedure. That has been fixed and this reflects the new steps that should work reliably.


geowa4:
It looks like fixing that panic in the existing code is going to require bumping some versions that will break at least one piece of existing code so this will have to be in another PR.

fahlmant (approved):
/lgtm
/approve

PR #424, opened 2023-07-28
Return all active clusters when using 'org clusters' command

AlexVulaj:
Resolves [OSD-17668](https://issues.redhat.com//browse/OSD-17668)

Rewrite the `org clusters ...` command to return all active clusters by default. Previously, this command would only return the first 100 results with no indication that the result set was incomplete, and no way to fetch more results.

Outward facing changes:
- Removed `--active` flag. This flag used to filter the result set down to only active clusters. This was no longer needed as this is now the default behavior
- Added `--all/-A` flag, which will return all clusters for an org regardless of status. This is to still allow for the old behavior, but respect the shift in default behavior as noted in the previous bullet point
- Updated `Long` description for the command and added multiple usage examples under `Example` that can be used by reviewers to test.
- Don't allow both orgId AND `--aws-profile` with `--aws-account-id` to be present - only one or the other.
- General logging improvements

Code Health changes:
- Refactored exported `SearchClusters` function to be more independent of Cobra CLI for easier testing and external calling
  - There were two flags used in the function related to AWS flags that were defined in a `common` file somewhere else. While we should absolutely stop doing this, I wasn't comfortable touching it at this time.
- Centralized call to `printClusters` to be in the command itself.
  - This is part of an effort I've been pushing to make all of our functions - even specific to a command - independent of Cobra itself. This allows functions to be re-used more freely and tested more easily.
- Removed custom structs in favor of ocm-sdk structs for calls to the accounts_management API.
- Refactored a branch of code that used to flow from "isAWSProfileSearch"
  - Prior to this, the old function `searchClustersByAWSProfile` would look up the org ID from the given AWS account info, and then call through to `searchClustersByOrg`, which was identical to the path taken when an org ID was present. I rewrote `searchClustersByAWSProfile` into `getOrganizationIdFromAWSProfile` which now only performs the one specific task it's named for. Then the two branches (aws vs. org ID) could be immediately consolidated back, making the flow of code easier to follow.
- Rewrote `searchclustersByOrg` into `searchAllClustersByOrg`. The function will now continuously page to the end of the result set, 100 items at a time.
- Rewrote `createGetClustersRequest` to use the more idiomatic notation for OCM API requests.
  - Instead of using a generic `Get` and manually providing an endpoint, search params, etc., we now chain function calls. This is more readable, easier to follow, and will return a specific API response of `*v1.SubscriptionsListRequest` instead of the generic `*sdk.Request`.
- Removed filtering from `printClusters`
  - a function named "printClusters" should do just that - print the clusters that it is given. It is left up to the caller to provide the filtered list that it wants printed.

### DIFF
--- a/cmd/org/aws-accounts.go
+++ b/cmd/org/aws-accounts.go
@@ -45,7 +45,7 @@ func init() {
 		"ou-id",
 		"",
 		"",
-		"specify orgnaization unit id",
+		"specify organization unit id",
 	)
 
 	AddOutputFlag(flags)
--- a/cmd/org/clusters.go
+++ b/cmd/org/clusters.go
@@ -1,13 +1,11 @@
 package org
 
 import (
-	"encoding/json"
 	"fmt"
-	"log"
+	accountsv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 	"os"
 
 	"github.com/aws/aws-sdk-go-v2/service/organizations"
-	"github.com/openshift-online/ocm-cli/pkg/arguments"
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	"github.com/openshift/osdctl/pkg/printer"
 	"github.com/openshift/osdctl/pkg/utils"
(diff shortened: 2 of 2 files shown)


cmd/org/clusters.go
```diff
@@ -20,39 +18,57 @@ const (
...
+`,
+		Args:          cobra.MaximumNArgs(1),
 		SilenceErrors: true,
 		Run: func(cmd *cobra.Command, args []string) {
-			cmdutil.CheckErr(SearchClusters(cmd, args))
+			var orgId string
+			if len(args) > 0 {
+				orgId = args[0]
+			} else {
+				orgId = ""
```
fahlmant, line 47: This is redundant; 43 already sets the orgId to ""

cmd/org/clusters.go
```diff
@@ -74,70 +90,77 @@ func init() {
...
+		return nil, fmt.Errorf("specify either an org id argument or --aws-profile, --aws-account-id arguments")
 	}
+
 	if isAWSProfileSearch() {
-		err = searchClustersByAWSProfile(cmd)
+		orgIdFromAws, err := getOrganizationIdFromAWSProfile()
+		orgId = *orgIdFromAws
+		if err != nil {
+			return nil, fmt.Errorf("failed to get org ID from AWS profile: %w", err)
+		}
```
typeid, line 107: **issue:** We're dereferencing the result (possibly a nil ptr) before checking `err != nil`

cmd/org/common.go
```diff
@@ -21,8 +21,8 @@ const (
 )
 
 var (
-	awsProfile string = ""
-	output     string = ""
+	awsProfile = ""
+	output     = ""
```
fahlmant, line 25: Is there a specific reason to remove string here?

typeid (changes requested):
One small issue, rest looks great to me. Good job on removing some unnecessary code like `hasOrgId`. :)

cmd/org/clusters.go
AlexVulaj: I personally like the explicit nature of this `else` so there's no question over the intention here. What I'll do instead of removing this line is:
```
orgId := ""
if len(args) > 0 {
	orgId = args[0]
}
```

fahlmant: line 50 below with status does the logic without the else

cmd/org/clusters.go
AlexVulaj: Great catch on this! Fixing now.

cmd/org/clusters.go
```diff
@@ -20,39 +18,55 @@ const (
...
+		Args:          cobra.MaximumNArgs(1),
 		SilenceErrors: true,
 		Run: func(cmd *cobra.Command, args []string) {
-			cmdutil.CheckErr(SearchClusters(cmd, args))
+			orgId := ""
+			if len(args) > 0 {
+				orgId = args[0]
+			}
+
+			var status string
```
fahlmant, line 48:
```suggestion
			status := ""
```
to match 43 since these use the exact same logic, default to empty unless something is set specifically.

cmd/org/common.go
fahlmant: It's more explicit, so easier to read and already there. I don't see a benefit either way to remove it.

cmd/org/aws-accounts.go
```diff
@@ -45,7 +45,7 @@ func init() {
 		"ou-id",
 		"",
 		"",
-		"specify orgnaization unit id",
+		"specify organization unit id",
```
AlexVulaj, line 48: I can't remember why I came across this as part of this work... but either way I fixed it in here.

PR #432, opened 2023-08-29
OSD-17941: Add etcd member replacement command

devppratik:
This PR aims to add a new command to osdctl i.e `osdctl cluster etcd-member-replacement --node <node-id>`

The command aims to automate the member replacement for etcd in case the state of unhealthy etcd member is `CrashLoopBackOff` or `Error` following the steps mentioned in the [doc](https://docs.openshift.com/container-platform/4.13/backup_and_restore/control_plane_backup_and_restore/replacing-unhealthy-etcd-member.html#restore-replace-crashlooping-etcd-member_replacing-unhealthy-etcd-member)

Assuming on of the pods is in `CrashLoopBackOff`
```
 oc -n openshift-etcd get pods -l k8s-app=etcd                                    
NAME                                READY   STATUS             RESTARTS       AGE
etcd-ip-10-0-154-180.ec2.internal   3/4     CrashLoopBackOff   5 (100s ago)   9m10s
etcd-ip-10-0-165-39.ec2.internal    4/4     Running            0              5m35s
etcd-ip-10-0-210-217.ec2.internal   4/4     Running            0              7m27s
```
 On running the command, we get
```
[INFO] This operation will run with the "backplane-cluster-admin" role

[INFO] Replacing pod etcd-ip-10-0-154-180.ec2.internal having member id 363b728d88e78c7a.
Continue? (y/N): y
Member 363b728d88e78c7a removed from cluster 8234ce6cfd037151

[INFO] Turning the quorum guard off
[INFO] Deleting the secrets for the unhealthy etcd member
[INFO] Forcing etcd redeployment
[INFO] Turning the quorum guard back on
The etcd member has been successfully replaced. Please verify by running health check after few minutes.
```

### DIFF
--- a/cmd/cluster/etcd_health.go
+++ b/cmd/cluster/etcd_health.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 
@@ -25,10 +26,11 @@ import (
 const (
 	EtcdNamespaceName       = "openshift-etcd"
 	EtcdMemberConditionType = "EtcdMembersAvailable"
-	ConatinerName           = "etcdctl"
+	ContainerName           = "etcdctl"
 	MasterNodeLabel         = "node-role.kubernetes.io/master"
 	EtcdPodMatchLabelName   = "k8s-app"
 	EtcdPodMatchValueName   = "etcd"
+	EtcdLabelSelector       = "k8s-app=etcd"
 )
 
 var etcdctlCmd = []string{
@@ -51,6 +53,34 @@ func (capture *logCapture) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
+func getKubeConfigAndClientSet() (*rest.Config, *kubernetes.Clientset, error) {
+	var kubeconfig *string
+
+	if home := homedir.HomeDir(); home != "" {
+		kubeconfig = flag.String("kubeconfig", filepath.Join(home, ".kube", "config"), "(optional) absolute path to the kubeconfig file")
+	} else {
+		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file")
+	}
+	flag.Parse()
+
+	// use the current context in kubeconfig
+	kconfig, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	kconfig.Impersonate = rest.ImpersonationConfig{
+		UserName: BackplaneClusterAdmin,
+	}
+	// create the clientset
+	clientset, err := kubernetes.NewForConfig(kconfig)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return kconfig, clientset, err
+}
+
 func newCmdEtcdHealthCheck(kubeCli client.Client, flags *genericclioptions.ConfigFlags) *cobra.Command {
 	return &cobra.Command{
 		Use:               "etcd-health-check",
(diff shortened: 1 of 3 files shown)


cmd/cluster/etcd_replace.go
```diff
@@ -0,0 +1,183 @@
...
+	output, err := Etcdctlhealth(kconfig, clientset, remove_cmd, pod)
+	if err != nil {
+		fmt.Println("Could not replace pod. Refer error below")
+		return err
+	}
+	fmt.Println(output)
+	return nil
+}
+
+func patchEtcdQuorum(kubeCli client.Client, flags *genericclioptions.ConfigFlags, patch string) error {
```
mrbarge, line 158: As this function is being used for more than just patching quorum, it could maybe be better renamed to something more generic.

cmd/cluster/etcd_replace.go
```diff
@@ -0,0 +1,183 @@
...
+					return err
+				}
+				return nil
+			}
+		}
+	}
+	return fmt.Errorf("none of the etcd members seems to be unhealthy. Please verify once again")
+}
+
+func ReplaceEtcdMember(kubeCli client.Client, kconfig *rest.Config, clientset *kubernetes.Clientset, flags *genericclioptions.ConfigFlags, opts flagOptions, pod string) error {
```
mrbarge, line 93: Some steps executed as part of this function do not log prior to their execution. It should be very clear to an SRE exactly what is happening throughout this process, so `removeEtcdMember` and `removeEtcdSecrets` should log prior to their call.

This is also important from the perspective of giving SRE context if one of these steps were to error. Currently, the error is passed right back up the chain and dumped out. It should be clear to an SRE to know exactly at what point in the member replacement process that error occurred and what step failed - because the SRE is likely going to have to manually take care of the rest of the steps, and it will help for them to not have to try and guess or figure out where they need to pick up from.

weherdh:
From the [doc](https://docs.openshift.com/container-platform/4.13/backup_and_restore/control_plane_backup_and_restore/replacing-unhealthy-etcd-member.html#restore-replace-crashlooping-etcd-member_replacing-unhealthy-etcd-member), it mentioned the first step we need to debug to the node and backup the manifest file, do we have this step? Or it is not very necessary?

devppratik:
Hey Wenqi
It was discussed not to include the backups as a part of the function and have the new member sync from the existing member

cmd/cluster/etcd_health.go
```diff
@@ -51,6 +53,34 @@ func (capture *logCapture) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
+func getKubeConfigAndClientSet() (*rest.Config, *kubernetes.Clientset, error) {
```
mrbarge, line 56: As I've noted in Slack, we should leverage the existing client that is supplied by `newCmdEtcdHealthCheck` rather than initialise a new one.

One of the main issues with the initialisation code as-is is that it does not respect the `KUBECONFIG` environment variable, which means that this code could be picking up a kubeconfig different to what the SRE expects.

As this is pretty important to prevent accidental client connection confusion, I'd like to add a hold until it's addressed somehow.

/hold

mjlshen: One option is to take the local kubeconfig out of the picture and supply a cluster id using one of these: https://github.com/openshift/osdctl/blob/34688d56ff7a79ca70b372e39a93ca21f4a2c756/pkg/k8s/client.go#L94-L120

It's utilized a lot in the infra node resize command https://github.com/openshift/osdctl/blob/master/cmd/cluster/resize/cmd.go since the automation needs to swap back and forth between a cluster and its Hive (and sometimes as backplane-cluster-admin)

mrbarge: Nice! That would do very well here.

devppratik: Updated with the suggestions

mrbarge:
Thanks for the changes @devppratik .

I tested this as follows:
- Entered a random etcd member's pod on the cluster
- `echo asdfadfasdfasf > /var/lib/etcd/member/snap/db` to cause a member to corrupt and crash
- Ran `osdctl cluster etcd-member-replace $CLUSTERID --node $NODE`
- Verified recovery of etcd

I also verified that running `osdctl cluster etcd-health-check` on the cluster also detected the unhealthy cluster and prompted to run `etcd-member-replace` with the correct arguments.

/lgtm

devppratik:
Thank you Matt for the review!

mjlshen (approved):
/approve

PR #433, opened 2023-08-31
remove docs and docs command

mjlshen:
Just my personal opinion that these are never referenced and we can depend on each command's `-h` flag for documentation. At minimum, these docs aren't kept up-to-date.

[OSD-18348](https://issues.redhat.com//browse/OSD-18348)


typeid:
+1 on this. Will leave it open for someone else to lgtm:)

iamkirkbater:
I would be opposed to this if we published these docs anywhere automatically, but as far as I know we have no plans on doing either of those things, so 👍🏼 from me.

/lgtm

PR #437, opened 2023-09-05
[OSD-17226] Only do a CPD if no SL was sent yet.

bergmannf:
In case a SL was already posted to the cluster, abort early.

### DIFF
--- a/cmd/cluster/cpd.go
+++ b/cmd/cluster/cpd.go
@@ -2,12 +2,16 @@ package cluster
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"sort"
 
 	awsSdk "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/openshift/osdctl/cmd/network"
+	"github.com/openshift/osdctl/cmd/servicelog"
+	sl "github.com/openshift/osdctl/internal/servicelog"
 	"github.com/openshift/osdctl/pkg/osdCloud"
 	"github.com/openshift/osdctl/pkg/provider/aws"
 	"github.com/openshift/osdctl/pkg/utils"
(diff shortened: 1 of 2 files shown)


cmd/cluster/cpd.go
```diff
@@ -75,6 +79,17 @@ func (o *cpdOptions) run() error {
 		return nil
 	}
 
+	fmt.Println("Checking if a servicelog has already been sent")
+	externalservicelogs, err := findExistingExternalServiceLogs(o.clusterID)
+	if err != nil {
+		fmt.Println("Could not retrieve servicelogs for cluster")
+	}
+	if len(externalservicelogs) > 0 {
+		fmt.Printf("A servicelog has already been sent on %s - stopping CPD analysis.\n", externalservicelogs[len(externalservicelogs)-1].CreatedAt.Format("2006-01-02"))
```
mjlshen, line 88: PR makes sense to me, what do you think about printing the SL message here?

cmd/cluster/cpd.go
```diff
@@ -75,6 +79,17 @@ func (o *cpdOptions) run() error {
...
+		fmt.Println("Check https://github.com/openshift/ops-sop/blob/master/v4/alerts/ClusterProvisioningFailure.md#toil-reduction to see if this check and servicelog sending can be automated.")
```
typeid, line 89: I'm a bit confused here, when does this happen?
If a SL is sent by OCM, I would expect the CPD to never reach SRE. Are we adding logic to catch OCM bugs? 

This message also confuses me - the SL is sent, what needs automation is that the CPD doesn't reach SRE, correct? IMO this message reads as "there is no SL automation yet".
```
		fmt.Println("Check https://github.com/openshift/ops-sop/blob/master/v4/alerts/ClusterProvisioningFailure.md#toil-reduction to see if this check and servicelog sending can be automated.")
```

bergmannf: I think this is more about not accidentally sending 2 SLs by hand, but in general I agree with the sentiment.
Maybe @iamkirkbater can shed some light about wanting this output as well.

iamkirkbater: Ahh, this looks like it was a bit of a misunderstanding of the original card, though I'm not opposed to also checking about the status of the servicelog here as well in order to not duplicate effort.

What we want to do is ensure that the CPD response does not continue when a non OCM-3999 error code is fired - as that means the customer has already been informed via the OCM console as to why the cluster installation failed.

Similar to this step of the SOP utility that has since been removed in favor of this script: https://github.com/openshift/ops-sop/blob/9ef748d233f810c98e8efadca0bb7348c8843b4c/v4/utils/cluster_provisioning_failure#L83-L95

typeid: > What we want to do is ensure that the CPD response does not continue when a non OCM-3999 error code is fired
I believe I had removed the check for non OCM-3999 here, based on [this discussion](https://github.com/openshift/configuration-anomaly-detection/pull/212#discussion_r1281800278). 
TLDR of the discussion was: non OCM-3999 does not mean SRE does not have to investigate. Anything reaching PD needs to be investigated. If this is not the case, we should still aim to fix this in OCM (do not trigger CPDs) instead of here IMO.

bergmannf: Hm there is no mentions of OCM-3999 specifics in the OSD card this is based on, so maybe that should be reworded, but IMO preventing any accidental sending of multiple SLs is useful.
I am wondering if it wouldn't be better to put this in the service log post command though, by checking 'was the last SL send within X minutes/hours' and then communicate this to the SRE.
That would make it generic and prevent flooding a customer with SLs in general?

typeid: Agreed that this should not be limited to this single command.

iamkirkbater: I'm fine with closing this PR if we don't think this is necessary, and I'm also fine with seeing the double SL prevention logic get pushed into the SL functionality.

bergmannf:
Closing this PR as per discussion - checking for double SL sending should be done in the normal SL post command.

PR #438, opened 2023-09-07
Adds env config loading for JIRA token

iamkirkbater:
Refactors the context command to load JIRA token from env or actually from the flag.

Tested by running `osdctl cluster context` and seeing it not load my JIRA token in my environment, then building this commit and doing the following:

* `export JIRA_API_TOKEN="my-api-token"; osdctl cluster context $CLUSTER_ID` - it works now
* `osdctl cluster context $CLUSTER_ID --jiratoken malformed-token` - it does not work due to a permissions error (expected, it's using "malformed-token" as the token from the flag, even though my env var is correctly set) 
* `export JIRA_TOKEN="$JIRA_API_TOKEN"; export JIRA_API_TOKEN=malformed-token; osdctl cluster context $CLUSTER_ID` - it does not work due to a permissions error (expected, it's using "malformed-token" from the env var because no flag is set.
* `osdctl cluster context $CLUSTER_ID --jiratoken $JIRA_TOKEN` - works (expected, because even though my env var is "malformed-token" and would fail, the correct token is passed in through the flag.

### DIFF
--- a/cmd/cluster/context.go
+++ b/cmd/cluster/context.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/openshift/osdctl/cmd/servicelog"
 	"os"
 	"os/exec"
 	"sort"
@@ -12,6 +11,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/openshift/osdctl/cmd/servicelog"
 
 	pd "github.com/PagerDuty/go-pagerduty"
 	"github.com/andygrunwald/go-jira"
@@ -24,6 +25,7 @@ import (
 	"github.com/openshift/osdctl/pkg/printer"
 	"github.com/openshift/osdctl/pkg/utils"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/cluster/context.go
```diff
@@ -162,9 +164,33 @@ func (o *contextOptions) complete(cmd *cobra.Command, args []string) error {
...
+	// The order in which these should be loaded would be:
+	// 1. config file
+	// 2. env var
+	// 3. command flag
+	// The highest number should be the one used. So if you have set a value in the config file but pass a different value in the flag, the flag value should be used.
+	// Viper should have already read in the config file at this point.
+
+	// We specifically _don't_ want to use viper's AutomaticEnv functionality for this because we want to leverage the same env vars that the external tooling uses instead of having to set separate env vars
+
+	jiraTokenEnv := os.Getenv("JIRA_API_TOKEN")
```
clcollins, line 183: Wouldn't this be more efficient to reverse the order and return if the highest prio one is found before moving on to load the others?  I know it's a trivial thing, but every little bit helps.

clcollins (changes requested):

cmd/cluster/context.go
geowa4: Proposal:
```
if o.jiratoken != "" {
	viper.Set(utils.JiraTokenConfigKey, o.jiratoken)
}
else if jiraTokenEnv, isSet := os.LookupEnv("JIRA_API_TOKEN"); isSet {
	viper.Set(utils.JiraTokenConfigKey, jiraTokenEnv)
}
```

PR #439, opened 2023-09-08
Update `org context` so it doesn't return when failing to gather context

tnierman:
Updates `org context` so that we don't return immediately after encountering an error when gathering cluster context. Instead, the command will report the problem, but still make a best-effort attempt at printing whatever context it did recover


iamkirkbater:
/approve

How can this be tested?

AlexVulaj:
Great work on this! Big fan of this change.

/lgtm

PR #440, opened 2023-09-08
Refactors the PD service to be testable.

iamkirkbater:
This refactors the PD service to be it's own tested package within pkg/providers instead of being an untested pkg/utils file.

Would love to keep doing this to the services like JIRA that are present.

Still a WIP for now - ~I still need to migrate the last function that's present in the `cluster context` command into the package itself because I didn't realize it was present until I went to delete the `pkg/utils/pagerduty.go` file.~

The last function has been migrated - now I just need to decide whether or not I want to try to shim in tests for that case or circle back to that specific function.

Knowing my time limitations I'm no longer marking this as a draft so that it can be pushed ahead, and maybe we can circle back to write tests for this later.


mjlshen (approved):
/lgtm

PR #441, opened 2023-09-08
pass customer-specified AWS tags to network verify-egress

mjlshen:
If a customer specifies custom AWS tags, they should be passed to the network verifier instance as the customer may be using tag-based enforcement in their AWS environment.

[OSD-18409](https://issues.redhat.com//browse/OSD-18409)

### DIFF
--- a/cmd/network/verification.go
+++ b/cmd/network/verification.go
@@ -266,7 +266,7 @@ func (e *EgressVerification) generateAWSValidateEgressInput(ctx context.Context,
 		}
 	}
 
-	input, err := defaultValidateEgressInput(ctx, region)
+	input, err := defaultValidateEgressInput(ctx, e.cluster, region)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/network/verification.go
```diff
@@ -571,13 +571,19 @@ func filtersToString(filters []types.Filter) string {
...
-	awsDefaultTags := map[string]string{
+// Tags from the cluster are passed to the network-verifier instance
+func defaultValidateEgressInput(ctx context.Context, cluster *cmv1.Cluster, region string) (*onv.ValidateEgressInput, error) {
+	networkVerifierDefaultTags := map[string]string{
 		"osd-network-verifier": "owned",
 		"red-hat-managed":      "true",
 		"Name":                 "osd-network-verifier",
 	}
 
+	// TODO: When this command supports GCP this will need to be adjusted
```
geowa4, line 582: @mjlshen would it make sense to wrap this loop in a conditional that prints a warn log if GCP?

PR #443, opened 2023-09-15
[OSD-18424] Check hibernation in cluster context

bergmannf:
Print out any hibernations that the cluster has performed.

This can make figuring out strange apiserver errors easier, as prolonged hibernation will force the internal certificates to expire.


bergmannf:
Instead of putting this check into osdctl i'll move this to CAD.

PR #445, opened 2023-09-19
[OSD-17502] Replace pull-secret as part of osdctl cluster transfer-owner

reedcort:
closes [OSD-17502](https://issues.redhat.com/browse/OSD-17502)

### DIFF
--- a/cmd/cluster/transferowner.go
+++ b/cmd/cluster/transferowner.go
@@ -25,6 +25,8 @@ import (
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 )
 
+const CheckSyncMaxAttempts = 24
+
 // transferOwnerOptions defines the struct for running transferOwner command
 type transferOwnerOptions struct {
 	output       string
@@ -81,7 +83,7 @@ func updatePullSecret(conn *sdk.Connection, kubeCli client.Client, clientset *ku
 
 	clusterDeployments := &hiveapiv1.ClusterDeploymentList{}
 	if err := kubeCli.List(context.TODO(), clusterDeployments, client.InNamespace(hiveNamespace)); err != nil {
-		return err
+		return fmt.Errorf("failed to list cluster deployments in namespace %v: %w", hiveNamespace, err)
 	}
 
 	if len(clusterDeployments.Items) == 0 {
@@ -92,7 +94,7 @@ func updatePullSecret(conn *sdk.Connection, kubeCli client.Client, clientset *ku
 	// Delete the secret
 	err := clientset.CoreV1().Secrets(hiveNamespace).Delete(context.TODO(), secretName, metav1.DeleteOptions{})
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to delete secret %v in namespacd %v: %w", secretName, hiveNamespace, err)
 	}
 
 	secret := &corev1.Secret{
@@ -107,27 +109,24 @@ func updatePullSecret(conn *sdk.Connection, kubeCli client.Client, clientset *ku
 	}
 	_, err = clientset.CoreV1().Secrets(hiveNamespace).Create(context.TODO(), secret, metav1.CreateOptions{})
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create new secret in namespace %v: %w", hiveNamespace, err)
 	}
 
-	//Create a syncset on hive to push the pull secret down
-	syncsetName := "pull-secret-replacement"
-
-	err = createSyncSet(syncsetName, hiveNamespace, cdName, kubeCli)
+	err = awaitPullSecretSyncSet(hiveNamespace, cdName, kubeCli)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to synchronize pull secret for Hive namespace '%s' and ClusterDeployment '%s': %w", hiveNamespace, cdName, err)
 	}
 
 	return nil
 
 }
 
-func createSyncSet(syncSetName string, hiveNamespace string, cdName string, kubeCli client.Client) error {
+func awaitPullSecretSyncSet(hiveNamespace string, cdName string, kubeCli client.Client) error {
 	ctx := context.TODO()
 
 	syncSet := &hiveapiv1.SyncSet{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      syncSetName,
+			Name:      "pull-secret-replacement",
 			Namespace: hiveNamespace,
 		},
 		Spec: hiveapiv1.SyncSetSpec{
@@ -156,12 +155,16 @@ func createSyncSet(syncSetName string, hiveNamespace string, cdName string, kube
 
 	err := kubeCli.Create(ctx, syncSet)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create SyncSet: %w", err)
 	}
 
-	fmt.Printf("SyncSet %s in namespace %s has been created.\n", syncSetName, hiveNamespace)
+	fmt.Printf("SyncSet pull-secret-replacement in namespace %s has been created.\n", hiveNamespace)
 
-	hiveinternalv1alpha1.AddToScheme(kubeCli.Scheme())
+	err = hiveinternalv1alpha1.AddToScheme(kubeCli.Scheme())
+	if err != nil {
+		return fmt.Errorf("failed to add scheme: %w", err)
+	}
+
 	searchStatus := &hiveinternalv1alpha1.ClusterSync{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      cdName,
@@ -170,14 +173,14 @@ func createSyncSet(syncSetName string, hiveNamespace string, cdName string, kube
 	}
 	foundStatus := &hiveinternalv1alpha1.ClusterSync{}
 	isSSSynced := false
-	for i := 0; i < 24; i++ {
+	for i := 0; i < CheckSyncMaxAttempts; i++ {
 		err = kubeCli.Get(ctx, client.ObjectKeyFromObject(searchStatus), foundStatus)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to get status for resource %s: %w", searchStatus.GetName(), err)
 		}
 
 		for _, status := range foundStatus.Status.SyncSets {
-			if status.Name == syncSetName {
+			if status.Name == "pull-secret-replacement" {
 				if status.FirstSuccessTime != nil {
 					isSSSynced = true
 					break
@@ -200,7 +203,7 @@ func createSyncSet(syncSetName string, hiveNamespace string, cdName string, kube
 	// Clean up the SS on hive
 	err = kubeCli.Delete(ctx, syncSet)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to delete SyncSet: %w", err)
 	}
 
 	return nil
@@ -212,13 +215,13 @@ func rolloutTelemeterClientPods(clientset *kubernetes.Clientset, namespace, sele
 		LabelSelector: selector,
 	})
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to list pods in namespace '%s' with label selector '%s': %w", namespace, selector, err)
 	}
 
 	for _, pod := range pods.Items {
 		err = clientset.CoreV1().Pods(namespace).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to delete pod '%s' in namespace '%s': %w", pod.Name, namespace, err)
 		}
 		fmt.Printf("Pod %s in namespace %s has been deleted.\n", pod.Name, namespace)
 	}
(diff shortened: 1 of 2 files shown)


cmd/cluster/transferowner.go
```diff
@@ -54,9 +68,200 @@ func newTransferOwnerOptions(streams genericclioptions.IOStreams, globalOpts *gl
...
+	hiveinternalv1alpha1.AddToScheme(kubeCli.Scheme())
+	searchStatus := &hiveinternalv1alpha1.ClusterSync{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cdName,
+			Namespace: hiveNamespace,
+		},
+	}
+	foundStatus := &hiveinternalv1alpha1.ClusterSync{}
+	isSSSynced := false
+	for i := 0; i < 6; i++ {
```
geowa4, line 199: I believe this will wait for at most 30s, correct? Could we extend this to at least 2 minutes?

cmd/cluster/transferowner.go
```diff
@@ -54,9 +68,200 @@ func newTransferOwnerOptions(streams genericclioptions.IOStreams, globalOpts *gl
...
+		}
+
+		fmt.Printf(".")
+		time.Sleep(time.Second * 5)
+	}
+	if !isSSSynced {
+		return fmt.Errorf("syncset failed to sync. Please verify")
+	}
+
+	// Clean up the SS on hive
```
geowa4, line 226: I was waffling on whether this should be done in a defer block or at the end like you have it. I _think_ I prefer how you have it so the syncset can be interrogated if it fails. In that case though, could you add a statement to the error message just above this to say that the syncset is still there and will need to be deleted manually after it is inspected?

geowa4 (commented):
One thing I don't see is the equivalent of `pull_secret=$(ocm post --body=/dev/null --header="Impersonate-User=$username" /api/accounts_mgmt/v1/access_token)`. I might just have Primary brain though and missed it.

cmd/cluster/transferowner.go
```diff
@@ -54,9 +68,200 @@ func newTransferOwnerOptions(streams genericclioptions.IOStreams, globalOpts *gl
...
 
+func generateServiceLog(clusterId string, template string) servicelog.PostCmdOptions {
+	return servicelog.PostCmdOptions{
+		Template:       template,
+		ClusterId:      clusterId,
+		TemplateParams: []string{fmt.Sprintf("DATE=%v", time.Now().Format(time.RFC3339))},
+	}
+}
+
+func getHiveKubeConfigAndClient(clusterID string) (client.Client, *rest.Config, *kubernetes.Clientset, error) {
```
rendhalver, line 79: This seems to be implemented in a lot of places.
Can we move this into the global utils?

geowa4:
We ran this today on a real cluster and it appears to have worked.

cmd/cluster/transferowner.go
```diff
@@ -54,9 +66,175 @@ func newTransferOwnerOptions(streams genericclioptions.IOStreams, globalOpts *gl
...
+	err = createSyncSet(syncsetName, hiveNamespace, cdName, kubeCli)
+	if err != nil {
+		return err
+	}
+
+	return nil
+
+}
+
+func createSyncSet(syncSetName string, hiveNamespace string, cdName string, kubeCli client.Client) error {
```
geowa4, line 125: This is tiny but since this is also deleting the syncset at the end, I might rename this to something like `awaitPullSecretSyncSet`.

cmd/cluster/transferowner.go
```diff
@@ -54,9 +66,175 @@ func newTransferOwnerOptions(streams genericclioptions.IOStreams, globalOpts *gl
...
+			".dockerconfigjson": pullsecret,
+		},
+	}
+	_, err = clientset.CoreV1().Secrets(hiveNamespace).Create(context.TODO(), secret, metav1.CreateOptions{})
+	if err != nil {
+		return err
+	}
+
+	//Create a syncset on hive to push the pull secret down
+	syncsetName := "pull-secret-replacement"
```
geowa4, line 114: If this variable isn't used anywhere, it probably can just be hard-coded in the `createSyncSet` function.

cmd/servicelog/post.go
```diff
@@ -72,6 +72,9 @@ func newPostCmd() *cobra.Command {
 }
 
 func (o *PostCmdOptions) Init() error {
+	userParameterNames = []string{}
+	userParameterValues = []string{}
+	filterParams = []string{}
```
geowa4, line 77: @AlexVulaj this is what I mentioned to you earlier. We had to do some shenanigans to send multiple service logs in the same execution.

geowa4 (changes requested):
Structurally, everything appears in order. I just had a couple small comments.

cmd/cluster/transferowner.go
```diff
@@ -54,9 +66,175 @@ func newTransferOwnerOptions(streams genericclioptions.IOStreams, globalOpts *gl
...
+	}
+
+	err := kubeCli.Create(ctx, syncSet)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("SyncSet %s in namespace %s has been created.\n", syncSetName, hiveNamespace)
+
+	hiveinternalv1alpha1.AddToScheme(kubeCli.Scheme())
```
AlexVulaj, line 164: AddToScheme returns an error to my knowledge, can we handle it here real quickly?

```suggestion
	err = hiveinternalv1alpha1.AddToScheme(kubeCli.Scheme())
	if err != nil {
		return fmt.Errorf("failed to add scheme: %w" err)
	}
```

cmd/cluster/transferowner.go
```diff
@@ -54,9 +66,175 @@ func newTransferOwnerOptions(streams genericclioptions.IOStreams, globalOpts *gl
...
+}
+
+func updatePullSecret(conn *sdk.Connection, kubeCli client.Client, clientset *kubernetes.Clientset, clusterID string, pullsecret []byte) error {
+	currentEnv := utils.GetCurrentOCMEnv(conn)
+	secretName := "pull"
+	hiveNamespace := "uhc-" + currentEnv + "-" + clusterID
+
+	clusterDeployments := &hiveapiv1.ClusterDeploymentList{}
+	if err := kubeCli.List(context.TODO(), clusterDeployments, client.InNamespace(hiveNamespace)); err != nil {
+		return err
```
AlexVulaj, line 84: Instead of blindly passing through the error, I personally prefer to add a _bit_ of high level context in our own words. This helps give a little more contextual detail, as well as helps devs find the specific line an error is being thrown on if they need to debug.

```suggestion
		return fmt.Errorf("failed to list cluster deployments in namespace %v: %w", hiveNamespace, err)
```

cmd/cluster/transferowner.go
```diff
@@ -54,9 +66,175 @@ func newTransferOwnerOptions(streams genericclioptions.IOStreams, globalOpts *gl
...
+
+	if len(clusterDeployments.Items) == 0 {
+		return fmt.Errorf("failed to retreive cluster deployments")
+	}
+	cdName := clusterDeployments.Items[0].ObjectMeta.Name
+
+	// Delete the secret
+	err := clientset.CoreV1().Secrets(hiveNamespace).Delete(context.TODO(), secretName, metav1.DeleteOptions{})
+	if err != nil {
+		return err
```
AlexVulaj, line 95: Similar comment as above
```suggestion
		return fmt.Errorf("failed to delete secret %v in namespacd %v: %w", secretName, hiveNamespace, err)
```

cmd/cluster/transferowner.go
```diff
@@ -54,9 +66,175 @@ func newTransferOwnerOptions(streams genericclioptions.IOStreams, globalOpts *gl
...
+			Namespace: hiveNamespace,
+		},
+		Type: corev1.SecretTypeDockerConfigJson,
+		Data: map[string][]byte{
+			".dockerconfigjson": pullsecret,
+		},
+	}
+	_, err = clientset.CoreV1().Secrets(hiveNamespace).Create(context.TODO(), secret, metav1.CreateOptions{})
+	if err != nil {
+		return err
```
AlexVulaj, line 110: Same as before
```suggestion
		return fmt.Errorf("failed to create new secret in namespace %v: %w", hiveNamespace, err)
```

cmd/cluster/transferowner.go
```diff
@@ -54,9 +66,175 @@ func newTransferOwnerOptions(streams genericclioptions.IOStreams, globalOpts *gl
...
+	hiveinternalv1alpha1.AddToScheme(kubeCli.Scheme())
+	searchStatus := &hiveinternalv1alpha1.ClusterSync{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cdName,
+			Namespace: hiveNamespace,
+		},
+	}
+	foundStatus := &hiveinternalv1alpha1.ClusterSync{}
+	isSSSynced := false
+	for i := 0; i < 24; i++ {
```
AlexVulaj, line 173: Where is this "24" coming from here?

cmd/cluster/transferowner.go
```diff
@@ -54,9 +66,175 @@ func newTransferOwnerOptions(streams genericclioptions.IOStreams, globalOpts *gl
...
+		err = kubeCli.Get(ctx, client.ObjectKeyFromObject(searchStatus), foundStatus)
+		if err != nil {
+			return err
```
AlexVulaj, line 176: Same as the few above
```suggestion
		return fmt.Errorf("failed to delete syncset from hive: %w", err)
```

cmd/cluster/transferowner.go
AlexVulaj: Actually I prefer @geowa4 's suggestion below - awaitPullSecretSyncSet

cmd/cluster/transferowner.go
```diff
@@ -54,9 +66,175 @@ func newTransferOwnerOptions(streams genericclioptions.IOStreams, globalOpts *gl
...
+	return nil
+}
+
+func rolloutTelemeterClientPods(clientset *kubernetes.Clientset, namespace, selector string) error {
+	// Delete pods with the specified label selector in the specified namespace.
+	pods, err := clientset.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: selector,
+	})
+	if err != nil {
+		return err
```
AlexVulaj, line 215: There's a few more of these, I'll leave it to you to determine what you think is the best error message to accompany 😄

cmd/cluster/transferowner.go
```diff
@@ -69,10 +247,63 @@ func (o *transferOwnerOptions) run() error {
...
+	hiveCluster, err := utils.GetHiveCluster(o.clusterID)
+	hiveKubeCli, _, hivecClientset, err := getKubeConfigAndClient(hiveCluster.ID())
+	if err != nil {
+		return err
+	}
 
-	cluster, err := utils.GetCluster(ocm, o.clusterID)
+	response, err := ocm.AccountsMgmt().V1().AccessToken().Post().Impersonate(userName).Parameter("body", "/dev/null").Send()
+	if err != nil {
+		return fmt.Errorf("Can't send request: %v", err)
```
AlexVulaj, line 269: Very nitpicky, but try to use %w for errors in formatting
https://stackoverflow.com/a/61287626

cmd/servicelog/post.go
AlexVulaj: All the more reason I want us to start getting rid of global/package scoped vars as we come across them 😞

AlexVulaj:
Time permitting I'd really like to see some tests for the newly added functions, and I'm more than happy to pair on those if you'd like. Let me know!

cmd/cluster/transferowner.go
geowa4: The script this is adapted from hardcoded a number of attempts. 24 * 5 is a few more attempts than the script made since I had an issue with the script just missing the sync.

AlexVulaj: Sounds good - @reedcort can we store this value as a constant at the top of the file just so it's clear what's going on here? Maybe something like `CheckSyncMaxAttempts`? (Do feel free to come up with a better name 😄 )

cmd/cluster/transferowner.go
```diff
@@ -54,27 +68,247 @@ func newTransferOwnerOptions(streams genericclioptions.IOStreams, globalOpts *gl
...
+	}
+	authsMap := map[string]map[string]string{}
+	for k, auth := range auths {
+		authsMap[k] = map[string]string{
+			"auth":  auth.Auth(),
+			"email": auth.Email(),
+		}
+	}
+
+	pullSecret, err := json.Marshal(map[string]map[string]map[string]string{
```
lnguyen1401, line 287: Where's this pull-secret from? In the[ SOP](https://github.com/openshift/ops-sop/blob/master/v4/howto/replace-pull-secret.md#verify-the-new-secret), this pull-secret is from the output of this command
`ocm post --body=/dev/null --header="Impersonate-User=${NEW_OWNER_USERNAME}" /api/accounts_mgmt/v1/access_token > pull-secret.json
`

reedcort: @lnguyen1401 [line 270](https://github.com/reedcort/osdctl/blob/OSD-17502/cmd/cluster/transferowner.go#L270-L274) is where the pull secret is taken from.

lnguyen1401: so is the output from 270 the same as the output from the ocm command above ?

reedcort: yes

PR #446, opened 2023-09-25
Get the port from the OCM info.

bergmannf:
Some clusters used 6443 as the api port and not 443, so don't hard code the port.


mjlshen (approved):
This is a Hypershift bug/regression and it's going back to 443, but I think this is a good idea regardless

/lgtm

PR #447, opened 2023-09-26
Use new servicelog endpoint for listing service logs

AlexVulaj:
Resolves https://issues.redhat.com/browse/OSD-18457. At a high level, this converts our API call from

```
/api/service_logs/v1/clusters/bar/cluster_logs
```
to
```
/api/service_logs/v1/clusters/cluster_logs?cluster_id="foo"&cluster_uuid="bar"
```
I refactored the command itself to be more inline with how we declare command structs and flags now.

I also refactored to use the data models directly from the ocm sdk, which led to the following two PRs that we'll have to wait on.

Model:
https://github.com/openshift-online/ocm-api-model/pull/844

SDK:
https://github.com/openshift-online/ocm-sdk-go/pull/847


AlexVulaj:
/hold

We'll need to wait for a new ocm-sdk to be built after https://github.com/openshift-online/ocm-sdk-go/pull/847/ is merged

cmd/servicelog/list.go
```diff
@@ -1,123 +1,123 @@
...
-
-	// prefer cluster external over cluster internal ID
-	var formatMessage string
-	if cluster.ExternalID() != "" {
-		formatMessage = fmt.Sprintf(`search=cluster_uuid = '%s'`, cluster.ExternalID())
-	} else {
-		formatMessage = fmt.Sprintf(`search=cluster_id = '%s'`, cluster.ID())
-	}
+// TODO uncomment these fields after https://github.com/openshift-online/ocm-sdk-go/pull/847 is merged
+type LogEntryView struct {
```
typeid, line 73: **Question:** Why can't we use the `slv1.LogEntry` type directly? Is it to have everything as string for a proper print format after `viewBytes, err := json.Marshal(view)`?

typeid (commented):
Looking good, thanks for refactoring the flags on the way :)

cmd/servicelog/list.go
AlexVulaj: We can't marshal it because the fields aren't exported unfortunately. This is the struct:
```go
type LogEntry struct {
	bitmap_        uint32
	id             string
	href           string
	clusterID      string
	clusterUUID    string
	description    string
	eventStreamID  string
	logType        LogType
	serviceName    string
	severity       Severity
	subscriptionID string
	summary        string
	timestamp      time.Time
	username       string
	internalOnly   bool
}
```

After [chatting with the OCM team,](https://redhat-internal.slack.com/archives/CBDNMS43V/p1695676203847559), this was determined to be the best path forward.

AlexVulaj:
/unhold

After updating ocm-api-model and ocm-sdk-go, we can now get the missing fields from the new endpoint.
PRs here for those interested:

https://github.com/openshift-online/ocm-api-model/pull/844
https://github.com/openshift-online/ocm-api-model/pull/845
https://github.com/openshift-online/ocm-sdk-go/pull/847
https://github.com/openshift-online/ocm-sdk-go/pull/848

typeid (approved):
/lgtm

PR #448, opened 2023-09-28
add new 'jira quick-task' command

AlexVulaj:
Creates a new, simple Jira ticket using a set of configuration values and a given ticket summary/title.

```bash
$ osdctl jira create "Fix a thing"
Successfully created ticket:
https://issues.redhat.com/browse/OSD-12345
```

Resolves https://issues.redhat.com/browse/OSD-18777

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"github.com/openshift/osdctl/cmd/jira"
 	"os"
 	"strings"
 
@@ -93,6 +94,7 @@ func NewCmdRoot(streams genericclioptions.IOStreams) *cobra.Command {
 	rootCmd.AddCommand(org.NewCmdOrg())
 	rootCmd.AddCommand(sts.NewCmdSts(streams, kubeFlags, kubeClient))
 	rootCmd.AddCommand(promote.NewCmdPromote(kubeFlags, globalOpts))
+	rootCmd.AddCommand(jira.Cmd)
 
 	// add completion command
 	rootCmd.AddCommand(newCmdCompletion(streams))
--- a/cmd/jira/cmd.go
+++ b/cmd/jira/cmd.go
@@ -1,0 +1,13 @@
+package jira
+
+import "github.com/spf13/cobra"
+
+var Cmd = &cobra.Command{
+	Use:   "jira",
+	Short: "Provides a set of commands for interacting with Jira",
+	Args:  cobra.NoArgs,
+}
+
+func init() {
+	Cmd.AddCommand(quickTaskCmd)
+}
--- a/cmd/jira/quick-task.go
+++ b/cmd/jira/quick-task.go
@@ -1,0 +1,126 @@
+package jira
+
+import (
+	"fmt"
+	"github.com/andygrunwald/go-jira"
+	"github.com/openshift/osdctl/pkg/utils"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"strings"
+)
+
+const (
+	TeamNameKey        = "jira_team"
+	TeamLabelKey       = "jira_team_label"
+	BoardIdLabel       = "jira_board_id"
+	DefaultDescription = ""
+	DefaultTicketType  = "Task"
+	DefaultProject     = "OSD"
+	SprintState        = "active"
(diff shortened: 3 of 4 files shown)


cmd/jira/create.go
```diff
@@ -0,0 +1,100 @@
...
+	TeamLabelKey       = "jira_team_label"
+	BoardIdLabel       = "jira_board_id"
+	DefaultDescription = ""
+	DefaultTicketType  = "Task"
+	DefaultProject     = "OSD"
+	SprintState        = "active"
+)
+
+var createCmd = &cobra.Command{
+	Use:   "create title",
```
tnierman, line 23:
```suggestion
	Use:   "create <title>",
```

cmd/jira/create.go
```diff
@@ -0,0 +1,100 @@
...
+
+	var activeSprint jira.Sprint
+	for _, sprint := range sprints.Values {
+		if strings.Contains(sprint.Name, teamName) {
+			activeSprint = sprint
+			break
+		}
+	}
+
+	_, err = jiraClient.Sprint.MoveIssuesToSprint(activeSprint.ID, []string{issue.ID})
```
tnierman, line 94: I imagine most tickets created w/ this command should be added to the backlog, no? Would it be worth adding a flag to toggle adding to current sprint vs backlogging?

geowa4: The use case we had during our sprint retro was that we had a lot of untracked work so we wanted to optimize to put this into the current sprint. I could get behind a `--skip-sprint` option that skips this step.

AlexVulaj: I'm indifferent, however if we are to add a flag here I'd prefer the opposite of what's being suggested. That is - by default we don't add the ticket to the caller's current sprint, but we add a flag `--add-to-sprint` or something similar that would take this action.

Thoughts?

typeid: +1 on default being no sprint.

typeid (commented):
Looking good to me, would remove the active sprint addition default though. 

Generally think this is a great feature but starting to feel a bit unsure how this and some other recent features fit in `osdctl`. Quite a few additions lately seem to have blurred the lines between what I would think is supposed to land in here and what shouldn't. 
This is just a personal feeling though, and as long as this reduces toil and makes everyone's work easier, I'm always +1 :)

AlexVulaj:
@typeid this is a good point, and definitely brings up the concern of osdctl expanding too much.

A lot of times, on-call SREs (or after shift) find themselves doing unplanned work. This could be updating an operator, fixing an SOP, or otherwise. The intent of this ticket was to allow the SRE to quickly make a Jira ticket to track that effort, all without leaving their terminal and having to go through Jira's web UI.

It's possible the title and description of this new function do a poor job selling that intent - the current name does indeed make it seem like this is just for generically creating any old Jira ticket. I'm open to any suggested changes, or if people feel that this doesn't belong in osdctl at all.

AlexVulaj:
/hold

Important discussion going on right now that I think should be resolved before we move forward

cmd/jira/create.go
```diff
@@ -0,0 +1,100 @@
...
+	TeamLabelKey       = "jira_team_label"
+	BoardIdLabel       = "jira_board_id"
+	DefaultDescription = ""
+	DefaultTicketType  = "Task"
+	DefaultProject     = "OSD"
+	SprintState        = "active"
+)
+
+var createCmd = &cobra.Command{
+	Use:   "create <title>",
```
geowa4, line 23: Maybe we could rename this to more align with the intent we had in our team discussion to easily track work that it otherwise untracked. I like the name "quick-task" since the issue type this creates is a Task.

tnierman: I think it makes sense to make a new `ticket` subcommand family that handles any create, delete, modify actions against individual jira issues. Users could then invoke them with 

```
$ osdctl jira ticket create
 ...
 $ osdctl jira ticket delete
 ...
 $ osdctl jira ticket modify
 ...
 # (etc)
 ```

geowa4: I like that for a next iteration that builds off this work. For now though, a generic `jira ticket create` will need to be less prescriptive than the intent behind the team's request.

AlexVulaj:
/unhold

We've decided to move forward with this change for now, and reevaluate the purpose and structure of osdctl going forward.

PR #455, opened 2023-10-11
Use built-in functions to generate osdctl completion for bash, zsh, and fish

mjlshen:
Don't need to pipe streams through to this command and can greatly simplify it via cobra built-ins. Don't need to maintain a hand-rolled zsh completion function and it's easy to add fish support if anyone's using that.

[OSD-19092](https://issues.redhat.com//browse/OSD-19092)


typeid:
Neat!
/lgtm

typeid (approved):

PR #460, opened 2023-10-12
Remove unused global flags

mjlshen:
`AddGoFlagSet()` results in these global flags being added to osdctl:

```
Flags:
      --alsologtostderr                   log to standard error as well as files
      --log_backtrace_at traceLocations   when logging hits line file:N, emit a stack trace
      --log_dir string                    If non-empty, write log files in this directory
      --log_link string                   If non-empty, add symbolic links in this directory to the log files
      --logbuflevel int                   Buffer log messages logged at this level or lower (-1 means don't buffer; 0 means buffer INFO only; ...). Has limited applicability on non-prod platforms.
      --logtostderr                       log to standard error instead of files
      --s2a_enable_appengine_dialer       If true, opportunistically use AppEngine-specific dialer to call S2A.
      --s2a_timeout duration              Timeout enforced on the connection to the S2A service for handshake. (default 3s)
      --stderrthreshold severityFlag      logs at or above this threshold go to stderr (default 2)
  -v, --v Level                           log level for V logs
      --vmodule vModuleFlag               comma-separated list of pattern=N settings for file-filtered logging
```

of which I am arguing that none of which are used. `-v` Level could be used with `klog`, though a search through the codebase shows that even `klog` log levels aren't being used in this codebase, so I swapped those calls out as well. I don't think we want to bring back these flags :D

[OSD-19092](https://issues.redhat.com//browse/OSD-19092)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -53,7 +53,6 @@ require (
 	k8s.io/apimachinery v0.26.3
 	k8s.io/cli-runtime v0.26.3
 	k8s.io/client-go v0.26.3
-	k8s.io/klog/v2 v2.90.1
 	k8s.io/kubectl v0.26.3
 	k8s.io/utils v0.0.0-20230313181309-38a27ef9d749
 	sigs.k8s.io/controller-runtime v0.14.6
@@ -184,6 +183,7 @@ require (
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 	k8s.io/component-base v0.26.3 // indirect
+	k8s.io/klog/v2 v2.90.1 // indirect
 	k8s.io/kube-openapi v0.0.0-20221012153701-172d655c2280 // indirect
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/kustomize/api v0.12.1 // indirect
--- a/internal/utils/globalflags/globalflags.go
+++ b/internal/utils/globalflags/globalflags.go
@@ -1,8 +1,6 @@
 package globalflags
 
 import (
-	"flag"
-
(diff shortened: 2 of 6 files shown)


mjlshen:
By removing this, the output of help is now:
```
❯ ./osdctl -h
CLI tool to provide OSD related utilities

Usage:
  osdctl [command]

Available Commands:
  aao               AWS Account Operator Debugging Utilities
  account           AWS Account related utilities
  cluster           Provides information for a specified cluster
  clusterdeployment cluster deployment related utilities
  completion        Output shell completion code for the specified shell (bash or zsh)
  cost              Cost Management related utilities
  env               Create an environment to interact with a cluster
  help              Help about any command
  jira              Provides a set of commands for interacting with Jira
  jumphost          
  network           network related utilities
  org               Provides information for a specified organization
  promote           Utilities to promote services/operators
  servicelog        OCM/Hive Service log
  sts               STS related utilities
  upgrade           Upgrade osdctl
  version           Display the version

Flags:
      --as string                        Username to impersonate for the operation. User could be a regular user or a service account in a namespace.
      --cluster string                   The name of the kubeconfig cluster to use
      --context string                   The name of the kubeconfig context to use
  -h, --help                             help for osdctl
      --insecure-skip-tls-verify         If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
      --kubeconfig string                Path to the kubeconfig file to use for CLI requests.
  -o, --output string                    Valid formats are ['', 'json', 'yaml', 'env']
      --request-timeout string           The length of time to wait before giving up on a single server request. Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests. (default "0")
  -s, --server string                    The address and port of the Kubernetes API server
      --skip-aws-proxy-check aws_proxy   Don't use the configured aws_proxy value
  -S, --skip-version-check               skip checking to see if this is the most recent release

Use "osdctl [command] --help" for more information about a command.
```

PR #462, opened 2023-10-16
OSD-18260: Force-drain and terminate infra nodes after appropriate timeout for infra resize command

clcollins:
[OSD-18260](https://issues.redhat.com//browse/OSD-18260) - Force drain infra nodes; terminate after a timeout during infra resize
    
This will kick off a drain of infrastructure nodes ahead of machine pool deletion, and will terminate the nodes in AWS after a timeout has been reached.

Force-termination does not work for GCP currently because we do not have a programmatic generator of backplane credentials for the GCP Project console, but a placeholder exists for the future.
    
REF: https://issues.redhat.com/browse/OSD-18260
    
### DIFF
--- a/cmd/cluster/resize/infra_node.go
+++ b/cmd/cluster/resize/infra_node.go
@@ -1,5 +1,7 @@
 package resize
 
+// cspell:ignore embiggen
+
 import (
 	"context"
 	"errors"
@@ -8,14 +10,18 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/aws/smithy-go"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	"github.com/openshift/osdctl/cmd/servicelog"
+	"github.com/openshift/osdctl/pkg/osdCloud"
 	"github.com/openshift/osdctl/pkg/utils"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -24,6 +30,8 @@ const (
 	twentyMinuteTimeout                = 20 * time.Minute
 	twentySecondIncrement              = 20 * time.Second
 	resizedInfraNodeServiceLogTemplate = "https://raw.githubusercontent.com/openshift/managed-notifications/master/osd/infranode_resized_auto.json"
+	infraNodeLabel                     = "node-role.kubernetes.io/infra"
+	temporaryInfraNodeLabel            = "osdctl.openshift.io/infra-resize-temporary-machinepool"
 )
 
 func newCmdResizeInfra() *cobra.Command {
@@ -79,6 +87,7 @@ func (r *Resize) RunInfra(ctx context.Context) error {
 	tempMp := newMp.DeepCopy()
 	tempMp.Name = fmt.Sprintf("%s2", tempMp.Name)
 	tempMp.Spec.Name = fmt.Sprintf("%s2", tempMp.Spec.Name)
+	tempMp.Spec.Labels[temporaryInfraNodeLabel] = ""
 
 	instanceType, err := getInstanceType(tempMp)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/provider/aws/client.go
mjlshen, line 1: I don't think these changes will be needed - are we using the osdctl aws client interface?

cmd/cluster/resize/infra_node.go
```diff
@@ -155,6 +228,35 @@ func (r *Resize) RunInfra(ctx context.Context) error {
 		return err
 	}
 
+	// Wait for original nodes to delete
+	if err := wait.PollImmediate(twentySecondIncrement, twentyMinuteTimeout, func() (bool, error) {
+		// Re-check for originalNodes to see if they have been deleted
+		return r.nodesMatchExpectedCount(ctx, originalNodeSelector, 0)
+	}); err != nil {
+		switch {
+		case err == wait.ErrWaitTimeout:
```
AlexVulaj, line 237: Fun fact about Go - comparing errors for equality with `==` will fail on wrapped errors. It's usually preferred to use `errors.Is`
```suggestion
		errors.Is(err, wait.ErrWaitTimeout):
```

pkg/provider/aws/client.go
clcollins: This was used in the termination of the cloud instances, here:

https://github.com/openshift/osdctl/pull/462/files#diff-60acf4ba40b2632d8613b384f0add52d190d27e014f395ae5b636fad91934e33R521-R525

cmd/cluster/resize/infra_node.go
clcollins: TIL, thanks, fixed in next push.

cmd/cluster/resize/infra_node.go
```diff
@@ -132,13 +156,62 @@ func (r *Resize) RunInfra(ctx context.Context) error {
...
+
+		// Unschedulable + noExecute taint should immediately cause nodes to drain
+		node.Spec.Taints = append(node.Spec.Taints, *unschedulableTaint)
+		node.Spec.Taints = append(node.Spec.Taints, *noExecuteTaint)
+		node.Spec.Unschedulable = true
+
+		if err = r.client.Update(ctx, &node); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
clcollins: It does, actually.  The "NoExecute" taint immediately evicts pods from the node, and "NoSchedule" prevents them from being reschedule on that node.  You can test it manually by just adding the any two keys, one with the "NoExecute" and one with "NoSchedule" as the effect, to any node. :)

But point taken, I think given the machinepool deletion it's not necessary.  I was thinking it might give us a jump on the draining, but since it ended up removing the pool _immediately_ after this, it's no longer needed.

I'll remove this stuff.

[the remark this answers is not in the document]
clcollins: > Typically an eviction resource needs to be created https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/

Ah, TIL, thanks!

With the two taints, the NodeController is evicting the pods, so I'm guess it's the one creating the eviction resource, but doing so due to the taints added.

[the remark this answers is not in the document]
mjlshen: Ah TIL myself :D https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions

pkg/provider/aws/client.go
mjlshen: Yupyup, so we're definitely calling ec2:TerminateInstances, but I don't think we're using the osdctl aws client interface (a good thing imo). I think your code will work without these additions in this file/the mock - `ec2.NewFromConfig()` is a function from aws-sdk-go-v2

https://github.com/openshift/osdctl/blob/master/pkg/provider/aws/client.go#L207 This function returns an osdctl AWS client interface.

cmd/cluster/resize/cmd.go
```diff
@@ -74,7 +74,7 @@ func (r *Resize) New() error {
 		return err
 	}
 
-	c, err := k8s.New(cluster.ID(), client.Options{Scheme: scheme})
+	c, err := k8s.NewAsBackplaneClusterAdmin(cluster.ID(), client.Options{Scheme: scheme})
```
mjlshen, line 77: What do you think about adding a clientAdmin and using it when we need to escalate to backplane-cluster-admin kind of like `hive` and `hiveAdmin`

clcollins: I *think* by removing the taint process we won't need the backplane elevation anymore, hopefully!

pkg/provider/aws/client.go
clcollins: Ok, thanks, after looking at this more I understand what you mean.  I was sure my IDE tried to tell me it needed to be updated, but I must have made that assumption myself, because it no longer complains now that I've removed it. I dunno what happened there.  Friday brain, I guess. :smile:

clcollins:
@mjshen @AlexVulaj - I've addressed your comments here, and this is ready for a once-over again, if you have time.

Thanks!

mjlshen (approved):
/lgtm

PR #463, opened 2023-10-17
Adds missing PagerDuty client error handling

clcollins:
This adds missing PagerDuty client build error handling in the cluster context command, which was allowing a nil pointer to a *client when improperly setting tokens.

This will also disable PagerDuty context gathering if the client has not properly been created.


iamkirkbater (approved):
/lgtm

PR #465, opened 2023-10-27
Add the VPCEndpoint connection from customer -> MC.

bergmannf:
Fix an issue, when a route has a transitgateway instead of NAT.


mjlshen (approved):
/lgtm

PR #469, opened 2023-11-02
Add description and examples to the clustersync-failures command

zmird-r:
* Adds longer description and some examples
* Adds a check on the kubeconfig presence instead of throwing a panic error


mjlshen (approved):
/lgtm

PR #472, opened 2023-11-06
Add Organization ID to SUPPORTEX query

davemulford:
Get SUPPORTEX tickets using the new Organization ID field. The Customer Name field is still used to to still show previously approved requests.


mjlshen (approved):
/lgtm

PR #474, opened 2023-11-07
Remove '-C' flag from 'account cli' subcommand

tnierman:
Removes the `-C` flag from the `account cli` subcommand, as it's no longer a viable option after a recent backplane update.

### DIFF
--- a/cmd/account/cli.go
+++ b/cmd/account/cli.go
@@ -3,7 +3,6 @@ package account
 import (
 	"fmt"
 
-	"github.com/aws/aws-sdk-go-v2/aws/arn"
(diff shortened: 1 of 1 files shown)


AlexVulaj:
As far as I can tell, the new backplane rollout shouldn't have had any impact on osdctl's ability to get AWS access to clusters using older account roles. I'd like to hold off on this and dig in a bit more together if you have some time tomorrow, but I also won't hold up this pr if others feel the flag should be removed.

AlexVulaj:
/lgtm

After further discussion, let's remove this.

PR #475, opened 2023-11-08
Add cluster orgId command

AlexVulaj:
With the new isolated backplane flow, it can be helpful to know the external org ID of a cluster in the event of backplane failures. This command makes that easy.

Resolves [OSD-19611](https://issues.redhat.com//browse/OSD-19611)

Example run:
```
$ ./osdctl -S cluster orgId 16708aq3bjcegto32sduqkinc3cd392j
{
  "external_id": "1234567",
  "internal_id": "1GZa79autPFQcwk3tN64c55KORz"
}
```

### DIFF
--- a/cmd/cluster/org_id.go
+++ b/cmd/cluster/org_id.go
@@ -1,0 +1,59 @@
+package cluster
+
+import (
+	"encoding/json"
+	"fmt"
+	ctlutil "github.com/openshift/osdctl/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+type OrgId struct {
+}
+
+type OrgIdOutput struct {
+	ExternalId string `json:"external_id"`
+	InternalId string `json:"internal_id"`
+}
+
+func newCmdOrgId() *cobra.Command {
+	o := &OrgId{}
+
+	orgIdCmd := &cobra.Command{
+		Use:   "orgId CLUSTER_ID",
+		Short: "Get the OCM org ID for a given cluster",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := o.Run(args[0]); err != nil {
+				return fmt.Errorf("error fetching OCM org ID for cluster %v", args[0])
+			}
+			return nil
+		},
+	}
+	return orgIdCmd
+}
+
+func (o *OrgId) Run(clusterID string) error {
+	if err := ctlutil.IsValidClusterKey(clusterID); err != nil {
+		return err
+	}
+
+	connection, err := ctlutil.CreateConnection()
+	if err != nil {
+		return err
+	}
+	defer connection.Close()
+
+	org, err := ctlutil.GetOrganization(connection, clusterID)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


cmd/jira/quick-task.go
```diff
@@ -27,10 +27,10 @@ var quickTaskCmd = &cobra.Command{
 The ticket will be assigned to the caller and added to their team's current sprint as an OSD Task.
 A link to the created ticket will be printed to the console.`,
 	Example: `#Create a new Jira issue
-osdctl jira create "Update command to take new flag"
+osdctl jira quick-task "Update command to take new flag"
```
AlexVulaj, line 30: Found these while creating an accompanying card for this effort and figured I'd fix the typos here real quick.

cmd/cluster/org_id.go
```diff
@@ -0,0 +1,59 @@
...
+	ExternalId string `json:"external_id"`
+	InternalId string `json:"internal_id"`
+}
+
+func newCmdOrgId() *cobra.Command {
+	o := &OrgId{}
+
+	orgIdCmd := &cobra.Command{
+		Use:   "orgId CLUSTER_ID",
+		Short: "Get the external org ID for a given cluster",
```
mjlshen, line 23:
```suggestion
		Short: "Get the OCM org ID for a given cluster",
```

cmd/cluster/org_id.go
```diff
@@ -0,0 +1,59 @@
...
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := o.Run(args[0]); err != nil {
+				return fmt.Errorf("error fetching external org ID for cluster %v", args[0])
```
mjlshen, line 27:
```suggestion
				return fmt.Errorf("error fetching OCM org ID for cluster %v", args[0])
```

pkg/utils/utils.go
```diff
@@ -188,6 +188,18 @@ func GetSubscription(connection *sdk.Connection, key string) (subscription *amv1
 	return
 }
 
+func GetOrganization(connection *sdk.Connection, key string) (*amv1.Organization, error) {
```
mjlshen, line 191: I think that it would be nice to add something like
```go
// GetOrganization returns an *amv1.Organization given an OCM cluster name, external id, or internal id as key
```

At least, it isn't clear to me, when reading this code, what I can safely supply to `key string`

cmd/cluster/org_id.go
```diff
@@ -0,0 +1,59 @@
+package cluster
+
+import (
+	"encoding/json"
+	"fmt"
+	ctlutil "github.com/openshift/osdctl/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+type OrgId struct {
```
mjlshen, line 10: You're using this so that adding flags in the future will just be stored in this struct?

mjlshen (commented):
looks good overall, just some small nits

cmd/cluster/org_id.go
AlexVulaj: Correct, plus the consistency of having the "Run" commands of our commands tied to a struct for others to use.

PR #479, opened 2023-11-14
Deprecate etcd-health-check and ask to use managed-script instead

feichashao:
The command `osdctl cluster etcd-health-check` will elevate to `backplane-cluster-admin` and exec into the etcd pod locally. 

As we have a managed-script to perform health-check https://github.com/openshift/managed-scripts/tree/89151d6963debd872e797d8b765f7b2bc7c878f8/scripts/SREP/etcd-complete-health-check , we might redirect users to use the managed-script instead.

### DIFF
--- a/cmd/cluster/etcd_health.go
+++ b/cmd/cluster/etcd_health.go
@@ -80,6 +80,7 @@ func newCmdEtcdHealthCheck() *cobra.Command {
 		Long:              `Checks etcd component health status for member replacement`,
 		Args:              cobra.ExactArgs(1),
 		DisableAutoGenTag: true,
+		Deprecated:        "please use the managed-script SREP/etcd-complete-health-check to avoid backplane elevation",
 		Run: func(cmd *cobra.Command, args []string) {
 			cmdutil.CheckErr(EtcdHealthCheck(args[0]))
 		},


devppratik:
I guess we can modify the way the kubeconfig method is called for the `osdctl cluster etcd-health-check` command. Instead of using 
```
kubeconfig, err := bplogin.GetRestConfigAsUser(bp, clusterID, "backplane-cluster-admin")
```

We can use
```
kubeconfig, err := bplogin.GetRestConfigAsUser(bp, clusterID, "")
```
This would remove the backplane-cluster-admin role

mjlshen (approved):
/lgtm

feichashao:
Hi @devppratik , exec into an etcd pod requires backplane-cluster-admin, for this sub-command to function we will need backplane-cluster-admin here. The goal of deprecating this sub-command is to ask users to use a managed-script instead, which doesn't need the user to exec into a pod by themselves, thus the user doesn't need elevation.

typeid:
@feichashao can we update the command to use the managed script itself here?

devppratik:
When working on one of the alerts yesterday, managed-script was unable to create a container and run the script as the cluster was under heavy load. It was stuck with ContainerCreating. 
However, I was able to run osdctl command and get the health info for etcd. 
IMO we should keep the osdctl command as well or figure out a way to elevate with reason

typeid:
> IMO we should keep the osdctl command as well or figure out a way to elevate with reason

Honestly, I think we should not enforce the usage for managed scripts on the SRE-P side. From my understanding managed scripts are there for people that would otherwise not have the permissions, a compliance alert should not be a reason to make things more complicated for ourselves.

PR #480, opened 2023-11-14
OSD-19636: Add Dynatrace URL option to osdctl cluster context

devppratik:
This PR closes [OSD-19636](https://issues.redhat.com//browse/OSD-19636).

- It updates the `osdctl cluster context` command to include the Dynatrace Environment URL when running for Management Cluster.
- Adds a new command - `osdctl cluster dynatrace <Cluster ID>` to directly fetch the DT Url without context command.

Example Output
```
>> Dynatrace Environment URL
https://hrm15629.live.dynatrace.com
```


cmd/cluster/dynatrace.go
mjlshen, line 1: Can we make a new `pkg/dynatrace/dynatrace.go` with these functions and skip `osdctl cluster dynatrace`? I'm thinking that we should iterate on `osdctl cluster context` first before going to `osdctl cluster dynatrace`

cmd/cluster/dynatrace.go
```diff
@@ -0,0 +1,130 @@
...
+		Namespace("dynatrace").
+		Resource("dynakubes").
+		DoRaw(context.TODO())
+	if err != nil {
+		return "", err
+	}
+	err = json.Unmarshal(data, &dynakubeCRD)
+	if err != nil {
+		return "", err
+	}
```
mjlshen, line 101: Let's use `k8s.New()` and build a scheme with the Dynatrace CRDs instead of relying on client-go and `interfaces{}`

Sample: https://github.com/openshift/osdctl/blob/a2f088e87b325fda832a84985bf02caf974ab941/cmd/cluster/resize/cmd.go#L77-L80

cmd/cluster/dynatrace.go
```diff
@@ -0,0 +1,130 @@
...
+	regionPrefix := strings.Split(regionID, "-")[0]
+	switch OCMEnv {
+	case "production":
+		return DynatraceURLProduction[regionPrefix]
+	case "stage":
+		return DynatraceURLStaging[regionPrefix]
+	default:
+		return ""
+	}
+}
```
mjlshen, line 130: I'm not really a fan of this fallback to be in code - I would prefer that these tenant/region combos are documented in https://github.com/openshift/ops-sop/tree/master/dynatrace#what-environments-are-there instead. If something ever changes about these URLs or regions that'll mean we'll need to rebuild and release a new version of the osdctl binary.

cmd/cluster/dynatrace.go
```diff
@@ -0,0 +1,130 @@
...
+	if !strings.HasPrefix(cluster.Name(), "hs-mc-") {
+		return fmt.Errorf("cluster is Not a Management Cluster")
+	}
+	url, err := GetDynatraceURLFromCluster(cluster.ID())
+	if err != nil {
+		OCMEnv := ocmutils.GetCurrentOCMEnv(connection)
+		url = getDynatraceURLFromRegion(region, OCMEnv)
+		if url == "" {
+			return fmt.Errorf("the Dynatrace Environemnt URL could not be determined")
+		}
```
mjlshen, line 65: Related to another comment, but could we just refer users to an SOP, e.g. https://github.com/openshift/ops-sop/blob/master/dynatrace/README.md ?

cmd/cluster/context.go
```diff
@@ -393,6 +400,25 @@ func (o *contextOptions) generateContextData() (*contextData, []error) {
...
+		if !strings.HasPrefix(cluster.Name(), "hs-mc-") {
+			data.DyntraceEnvURL = "cluster is Not a Management Cluster"
+			return
+		}
+
+		data.DyntraceEnvURL, err = GetDynatraceURLFromCluster(data.ClusterID)
+		if err != nil {
+			data.DyntraceEnvURL = getDynatraceURLFromRegion(cluster.Region().ID(), data.OCMEnv)
+			if data.DyntraceEnvURL == "" {
+				errors = append(errors, fmt.Errorf("the Dynatrace Environemnt URL could not be determined"))
```
mjlshen, line 417:
```suggestion
				errors = append(errors, fmt.Errorf("the Dynatrace Environemnt URL could not be determined: %v", err))
```

cmd/cluster/context.go
```diff
@@ -393,6 +400,25 @@ func (o *contextOptions) generateContextData() (*contextData, []error) {
...
+			fmt.Fprintln(os.Stderr, "Getting Dynatrace URL...")
+		}
+		if !strings.HasPrefix(cluster.Name(), "hs-mc-") {
+			data.DyntraceEnvURL = "cluster is Not a Management Cluster"
+			return
+		}
+
+		data.DyntraceEnvURL, err = GetDynatraceURLFromCluster(data.ClusterID)
+		if err != nil {
+			data.DyntraceEnvURL = getDynatraceURLFromRegion(cluster.Region().ID(), data.OCMEnv)
```
mjlshen, line 415: Similar to other comments, I don't think we should fallback to searching by region in osdctl

mjlshen (changes requested):
Nice work on this! Let's focus on getting the done criteria for the card met, which is `osdctl cluster context` and start using it to figure out what we may need next before jumping to `osdctl cluster dynatrace`

AlexVulaj:
I'm in agreement with @mjlshen 's comment about creating a new k8s client for fetching the dynatrace CRD.

As for adding the second command, I'm fine with it for now. Most of that code is required for the ticket's acceptance criteria anyway.

cmd/cluster/context.go
```diff
@@ -393,6 +400,25 @@ func (o *contextOptions) generateContextData() (*contextData, []error) {
 		}
 	}
 
+	GetDynatraceURL := func() {
+		defer wg.Done()
+		if o.verbose {
+			fmt.Fprintln(os.Stderr, "Getting Dynatrace URL...")
+		}
+		if !strings.HasPrefix(cluster.Name(), "hs-mc-") {
```
wanghaoran1988, line 408: We can call the OCM cluster api and check the labels on the cluster to see whether it's MC or not.
/api/clusters_mgmt/v1/clusters/<cluster_id>/external_configuration/labels

cmd/cluster/context.go
```diff
@@ -393,6 +400,25 @@ func (o *contextOptions) generateContextData() (*contextData, []error) {
...
+			data.DyntraceEnvURL = "cluster is Not a Management Cluster"
```
wanghaoran1988, line 409: If it's a HCP cluster, can we also return an dynatrace URL on the correct tenant with  some filter preset so it can save our DDU when doing the query? like we can append the filter log type, management cluster name, namespace name and status error as a default.

wanghaoran1988: For HCP, what we can do can be:
1. Check it's HCP
2. Find it's management cluster
3. Build k8s client to the mangement cluster and query the dynatrace CRD

cmd/cluster/context.go
devppratik: ack.
I will try to improve the check in that way

cmd/cluster/context.go
devppratik: @wanghaoran1988 This will be taken up in https://issues.redhat.com/browse/OSD-19655.

cmd/cluster/context.go
devppratik: Updated with same

cmd/cluster/dynatrace.go
devppratik: Removed this fallback. Added link to SOP incase we are unable to determine the URL

cmd/cluster/dynatrace.go
devppratik: Removed client-go and updated with k8s.New()

mjlshen (approved):
/lgtm

PR #486, opened 2023-11-30
Osd 19746 svclog

kamlendu1982:
Acceptance criteria:

    osdctl will check if another (non-automated) servicelog has been sent to the cluster already within the last hour.
    If another servicelog has been sent already it will as the user to confirm sending the additonal servicelog.
    If osdctl is not running in an interactive context this check is disabled.

### DIFF
--- a/cmd/servicelog/post.go
+++ b/cmd/servicelog/post.go
@@ -44,6 +44,8 @@ type PostCmdOptions struct {
 	failedClusters     map[string]string
 }
 
+const documentationBaseURL = "https://docs.openshift.com"
+
 func newPostCmd() *cobra.Command {
 	var opts = PostCmdOptions{}
 	postCmd := &cobra.Command{
@@ -193,11 +195,28 @@ func (o *PostCmdOptions) Run() error {
 		log.Fatal("servicelog post command terminated")
 	}()
 
+	// cluster type for which documentation link is provided in servicelog description
+	docClusterType := getDocClusterType(o.Message.Description)
+
 	for _, cluster := range clusters {
 		request, err := o.createPostRequest(ocmClient, cluster)
 		if err != nil {
 			o.failedClusters[cluster.ExternalID()] = err.Error()
 			continue
+		}
+
+		// if servicelog description contains a documentation link, verify that
+		// documentation link matches the cluster product (rosa, dedicated)
+		if !o.skipPrompts && docClusterType != "" {
+			clusterType := cluster.Product().ID()
+
+			if docClusterType != clusterType {
+				log.Warn("The documentation mentioned in the servicelog is for '", docClusterType, "' while the product is '", clusterType, "'.")
+				if !ocmutils.ConfirmPrompt() {
+					log.Info("Skipping cluster ID: ", cluster.ID(), ", Name: ", cluster.Name())
+					continue
+				}
+			}
 		}
 
 		response, err := ocmutils.SendRequest(request)
@@ -211,6 +230,28 @@ func (o *PostCmdOptions) Run() error {
 
 	o.printPostOutput()
 	return nil
+}
+
+// if servicelog description contains documentation link, parse and return the cluster type from the url
+func getDocClusterType(message string) string {
+
+	if strings.Contains(message, documentationBaseURL) {
+		pattern := `https://docs.openshift.com/([^/]+)/`
+		re := regexp.MustCompile(pattern)
+		match := re.FindStringSubmatch(message)
+		if len(match) >= 2 {
+			productType := match[1]
+			if productType == "dedicated" {
+				// the documentation urls for osd use "dedicated" as the differentiator
+				// e.g. https://docs.openshift.com/dedicated/welcome/index.html
+				// for proper comparison with cluster product types, return "osd"
+				// where "dedicated" is used in the documentation urls
+				productType = "osd"
+			}
+			return productType
+		}
+	}
+	return ""
 }
 
 func (o *PostCmdOptions) check(response *sdk.Response, clusterMessage servicelog.Message) {


cmd/servicelog/post.go
```diff
@@ -87,14 +89,39 @@ func (o *PostCmdOptions) Validate() error {
...
+	timeStampToCompare := time.Now().Add(-time.Hour * time.Duration(numberOfHours))
+	for _, svclog := range response.Items().Slice() {
+		timestamp := svclog.Timestamp()
+		if timestamp.After(timeStampToCompare.UTC()) {
+			return true // prompt message for re-confirm
+		}
+	}
+	return false
+}
+
```
typeid, line 109: In case there was at least one service log, we probably want to print it/them as it is necessary information for the decision whether a new SL should be sent or not.

cmd/servicelog/post.go
```diff
@@ -87,14 +89,39 @@ func (o *PostCmdOptions) Validate() error {
...
 	}
 	if err := o.Validate(); err != nil {
 		return err
 	}
-
+	if term.IsTerminal(int(os.Stdout.Fd())) && o.CheckServiceLogsLastHour() {
+		var resp string
+		fmt.Println("There is a service log within last 60 mins, are you sure you want to submit! Type YES to confirm")
+		fmt.Scan(&resp)
+		if resp != "YES" {
```
typeid, line 121: I believe we already have a helper function that could be re-used here: https://github.com/openshift/osdctl/blob/cdfaf84d3593afed1797b42758f657baebdf04aa/pkg/utils/utils.go#L249-L264

cmd/servicelog/post.go
```diff
@@ -87,14 +89,39 @@ func (o *PostCmdOptions) Validate() error {
...
+	numberOfHours := 1           // number of hours we need to wait for svc logs
+	response, err := FetchServiceLogs(o.ClusterId, getAllMessages, getInternalLogsOnly)
+	if err != nil {
+		log.Fatalf("failed to fetch service logs: %q", err)
+	}
+	timeStampToCompare := time.Now().Add(-time.Hour * time.Duration(numberOfHours))
+	for _, svclog := range response.Items().Slice() {
+		timestamp := svclog.Timestamp()
+		if timestamp.After(timeStampToCompare.UTC()) {
+			return true // prompt message for re-confirm
```
typeid, line 104: nitpick: this comment has nothing to do here, the `CheckServiceLogsLastHour()` logic doesn't know what's happening outside of it.

typeid (changes requested):

typeid:
Thanks for the contribution @kamlendu1982! I added some small comments, but generally this is already very close to the goal :)

cmd/servicelog/post.go
```diff
@@ -87,14 +89,37 @@ func (o *PostCmdOptions) Validate() error {
...
+	numberOfHours := 1           // number of hours we need to wait for svc logs
+	response, err := FetchServiceLogs(o.ClusterId, getAllMessages, getInternalLogsOnly)
+	if err != nil {
+		log.Fatalf("failed to fetch service logs: %q", err)
+	}
+	timeStampToCompare := time.Now().Add(-time.Hour * time.Duration(numberOfHours))
+	for _, svclog := range response.Items().Slice() {
+		timestamp := svclog.Timestamp()
+		if timestamp.After(timeStampToCompare.UTC()) {
+			fmt.Println("Below service Log has been subitted in last 60 minutes\nDescription: ", svclog.Description())
```
typeid, line 104: I think it would be cleaner to print this line just once in case we had service logs in the past hour, then print the list of service logs - but I'll leave this up to you whether you want to change this or not.

cmd/servicelog/post.go
```diff
@@ -87,14 +89,37 @@ func (o *PostCmdOptions) Validate() error {
 	return nil
 }
 
+func (o *PostCmdOptions) CheckServiceLogsLastHour() bool {
```
typeid, line 92: nitpick: this could be private

```suggestion
func (o *PostCmdOptions) checkServiceLogsLastHour() bool {
```

cmd/servicelog/post.go
```diff
@@ -87,14 +89,37 @@ func (o *PostCmdOptions) Validate() error {
...
+	getAllMessages := false      // we need just manual entries
+	getInternalLogsOnly := false // we need all messages
+	numberOfHours := 1           // number of hours we need to wait for svc logs
```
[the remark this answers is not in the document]
typeid: +1 Always good to have things configurable. :)

cmd/servicelog/post.go
typeid, line 104: Typo:
```suggestion
			fmt.Println("Below service log has been submitted in last 60 minutes\nDescription: ", svclog.Description())
```

kamlendu1982:
incorporated the recommend changes and addressed the review comments.
Can you please have a final review.

cmd/servicelog/post.go
```diff
@@ -87,14 +89,36 @@ func (o *PostCmdOptions) Validate() error {
...
+
 func (o *PostCmdOptions) Run() error {
 	if err := o.Init(); err != nil {
 		return err
 	}
 	if err := o.Validate(); err != nil {
 		return err
 	}
-
+	if term.IsTerminal(int(os.Stdout.Fd())) && o.CheckServiceLogsLastHour() {
```
bergmannf, line 117: Had a talk today about another card that would add a check like this.
It might be useful to still print that there was a warning even if it doesn't stop the script (e.g. to STDERR) just so it can be seen.

T0MASD:
tide/merge-method-squash

typeid:
/approve

@kamlendu1982 could you please mark the comments as resolved if they no longer apply?

PR #488, opened 2023-12-12
OSD-19605 Use managed-scripts for validate-pull-secret to avoid elevation

feichashao:
What this PR does:
- It uses the [managed-script](https://github.com/openshift/managed-scripts/tree/main/scripts/SREP/get-pull-secret-email) to get the pull-secret's email.
    - If the managed-script fails to run due to the pull-secret couldn't pull the managed-script image, it will send a SL.
- It moves the current existing logic to `--elevate`, which requires backplane elevation.

Why this PR:
- The current workflow gets the pull-secret to local, but we only need to know the email in the pull-secret for validation. We can use a managed-script to only get the email from the cluster, so we don't need to get any sensitive data locally.

Note:
- I didn't add related test cases since it is calling external commands which is not easy to mock. Appreciate if any reviewer can share some ideas.
- Calling the `ocm backplane managedjob` directly because it doesn't have easy library to call, yet.

Tests:
Case 1:  Normal case.
```
$ osdctl cluster validate-pull-secret xxxxxxxxxxxxx          
Getting email from OCM
email from OCM: siwu@redhat.com
Creating a managedjob to get pull-secret email in the cluster
managedjob id: openshift-job-8h6kd
Waiting for managedjob to finish
email from managedjob: siwu@redhat.com
Email addresses match.
```

Case 2: The email address doesn’t match.
```
$ osdctl cluster validate-pull-secret xxxxxxxxxxxxxxx
Getting email from OCM
email from OCM: siwu@redhat.com
Creating a managedjob to get pull-secret email in the cluster
managedjob id: openshift-job-vzm92
Waiting for managedjob to finish
email from managedjob: fake@redhat.com
Pull secret email doesn't match OCM user email. Sending service log.
INFO[0049] The following clusters match the given parameters: 
siwu-test           xxxxxxxxxxxxxxx   ready               4.14.4              aws                 us-west-2

INFO[0049] The following template will be sent:         
{
xxxxxxxxxx
}
Continue? (y/N): y
```

Case 3:  The pull secret fail authentication causing the managed-script fail running.
```
$ osdctl cluster validate-pull-secret xxxxxxxxxxxx
Getting email from OCM
email from OCM: siwu@redhat.com
Creating a managedjob to get pull-secret email in the cluster
managedjob id: openshift-job-2vc2v
Waiting for managedjob to finish
managedjob failed due to failed pulling image, run the below command for more detail:
oc describe pod openshift-job-2vc2v -n openshift-backplane-managed-scripts
Sending service log
xxxxxxxxxxxx
```

Case 4: The managed-script fails to run for other reasons.
```
$ osdctl cluster validate-pull-secret xxxxxxxxxxxx       
Getting email from OCM
email from OCM: siwu@redhat.com
Creating a managedjob to get pull-secret email in the cluster
managedjob id: openshift-job-th7ss
Waiting for managedjob to finish
error: managedjob timeout, try --elevate to validate pull-secret without a managed job
```

Case 5: Use `--elevate` to get pull secret.
```
$ osdctl cluster validate-pull-secret 2820u3k6rj9gv9kes1hk04mnacntgc1h --elevate
Getting email from OCM
email from OCM: siwu@redhat.com
Getting the pull-secret in the cluster with elevated permissions
email from cluster: siwu@redhat.com
Email addresses match.
```

### DIFF
--- a/cmd/cluster/validatepullsecret.go
+++ b/cmd/cluster/validatepullsecret.go
@@ -4,6 +4,10 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/exec"
+	"regexp"
+	"strings"
+	"time"
 
 	v1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
 	"github.com/openshift/osdctl/cmd/servicelog"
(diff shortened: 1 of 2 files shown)


cmd/cluster/validatepullsecret.go
```diff
@@ -11,93 +17,233 @@ import (
...
 
+	fmt.Println("Email addresses match.")
+	return nil
+}
+
+// getPullSecretFromCluster runs a managed-script to get the pull-secret
+// email from cluster without backplane elevation
+// it returns the email, error and done
+// done means a service log has been sent
+func getPullSecretFromCluster(clusterID string) (string, error, bool) {
```
xiaoyu74, line 103: Is that only myself feel this function is a bit too long to understand?  From it's logic, looks it contains: 
- Create Mamanged job
- wait for job to finish
- handle job errors
- Extrat email from job output.

Is that better we break this function into smaller ones ?

xiaoyu74:
Great work @feichashao - I think the main logic does makes much sense to me.

Regarding the test, yeah, it can be tricky when we use the external commands that don't have a direct lib interface. Personally, I think at this stage, as long as we tested with the external commands against actual clusters, should be good but in the future, may be request/add go SDK for testable interfaces.  

Open to get ideas from other reviewers.

cmd/cluster/validatepullsecret.go
```diff
@@ -11,93 +17,233 @@ import (
...
+			if err = postCmd.Run(); err != nil {
+				return "", err, true
+			}
+			return "", nil, true
+		}
+		return "", fmt.Errorf("managedjob timeout, try --elevate to validate pull-secret without a managed job"), false
+	}
+
+	// Get the output of the managed script
+	getJobResultCmd := fmt.Sprintf("ocm backplane managedjob logs %s", jobId)
```
NautiluX, line 162: I like the interaction with the backplane CLI, probably we could even use it as package instead of a shell-out, that would be great. Probably not here if that's a bigger thing, but I feel like we may want to add more such interactions in the future.

cmd/cluster/validatepullsecret.go
```diff
@@ -11,93 +17,233 @@ import (
...
 	}
 
+	fmt.Println("Email addresses match.")
+	return nil
+}
+
+// getPullSecretFromCluster runs a managed-script to get the pull-secret
+// email from cluster without backplane elevation
+// it returns the email, error and done
+// done means a service log has been sent
```
NautiluX, line 102: Consider naming the return values to make this clearer

cmd/cluster/validatepullsecret.go
NautiluX: At least I think the interactions with backplane managed scripts should be wrapped, so they don't clutter this function too much.

cmd/cluster/validatepullsecret.go
```diff
@@ -11,93 +17,233 @@ import (
...
+	fmt.Printf("email from cluster: %s\n", clusterPullSecretEmail)
 
-	if account.Email() != clusterPullSecretEmail {
-		_, _ = fmt.Fprintln(os.Stderr, "Pull secret email doesn't match OCM user email. Sending service log.")
+	return clusterPullSecretEmail, nil, false
+}
+
+// getPullSecretFromOCM gets the cluster owner email from OCM
+// it returns the email, error and done
+// done means a service log has been sent
```
NautiluX, line 201: Consider naming return values to make it clearer, specifically what done means.

cmd/cluster/validatepullsecret.go
feichashao: Thanks for reviewing! I split the function in the new commit.

cmd/cluster/validatepullsecret.go
feichashao: Yeah, having some library instead of direct shell call would make it more maintainable.  For this time I think we can use the shell call to deliver a MVP and enhance in the future.

I split the managedjob related operation to smaller functions in the new commit.

cmd/cluster/validatepullsecret.go
feichashao: In the new commit, I named it `sentSL` to specifically indicate a SL had been sent.

feichashao:
> I would like to note the discussion of this PR: #479
> 
> There are clear downsides to using managed scripts over direct elevated commands, we should consider them and evaluate if we want to move towards managed scripts in the first place.

I think it is true that we shouldn't force to use a managed script. This PR keeps the `--elevate` flag to execute the current routine, while moving the default routine to use the managed-script. 

I did a test to measure the time consumed:  
In the `--elevate` mode (the current routine), it takes 6s to complete.
In the default mode (the managed-script route), it takes 28s to complete.

With the 22s extra cost, we gain the benefits:  
- No need to touch the pull-secret in our local environment.
- One less compliance workflow.

If it brings trouble to oncall, we can flip the default behavior to the elevation one. 

WDYT?

feichashao:
@xiaoyu74 @NautiluX  thank you for reviewing and provide suggestions!  I put a new commit to address comments, could you help another look with you have time?

bmeng:
the change lgtm

feichashao:
Re-based.  @bmeng

NautiluX (commented):
/lgtm

PR #490, opened 2023-12-19
Fix yaml parsing to allow promoting MCC

NautiluX:
Also, add service name guessing so passing in `saas-` gets optional

### DIFF
--- a/cmd/promote/git/app_interface.go
+++ b/cmd/promote/git/app_interface.go
@@ -16,9 +16,9 @@ type Service struct {
 		Name    string `yaml:"name"`
 		URL     string `yaml:"url"`
 		Targets []struct {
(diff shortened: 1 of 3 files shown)


typeid:
@NautiluX For my understanding: what's preventing to promote MCC here? From the PR I would have thought the MCC saas file is called `managed-cluster-config` vs `saas-managed-cluster-config`, but that doesn't seem to be the case?

typeid:
/lgtm

To answer the above question, the yaml parsing failed, the change in the saas file name retrieval is unrelated:
```
Parameters map[string]interface{} `yaml:"parameters"`
```
Some parameters are wrapped by an additional array so the previous implementation did not work

PR #495, opened 2024-01-04
OSD-15383: osdctl LS delete without -i flag for single LS Reason

vibhuti-chalise:
in response to the issue raised in https://issues.redhat.com/browse/OSD-15383, this PR
-  allows cluster support delete to delete the "only" LS reason ID in the absence of the -i flag
-  and if there are multiple LS reason IDs then, it asks for either a specific LS reason ID or the -all flag

### DIFF
--- a/cmd/cluster/support/delete.go
+++ b/cmd/cluster/support/delete.go
@@ -9,6 +9,7 @@ import (
 	"github.com/openshift-online/ocm-cli/pkg/arguments"
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/gcp-project-operator/pkg/util/errors"
 	"github.com/openshift/osdctl/internal/support"
 	"github.com/openshift/osdctl/internal/utils/globalflags"
 	"github.com/openshift/osdctl/pkg/utils"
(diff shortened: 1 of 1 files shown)


cmd/cluster/support/delete.go
```diff
@@ -120,15 +116,22 @@ func (o *deleteOptions) run() error {
 	}
 
 	var limitedSupportReasonIds []string
+	limitedSupportReasons, err := getLimitedSupportReasons(o.clusterID)
+
+	if len(limitedSupportReasons) == 0 {
+		fmt.Fprintf(os.Stderr, "Cluster does not has any limited support reason ID \n")
+		os.Exit(1)
```
typeid, line 123: **typo:**
```
fmt.Fprintf(os.Stderr, "Cluster does not has any limited support reason ID \n")
```
We could re-phrase to: `Cluster is not in limited support.`

**nitpick:** exit in the main logic by returning the error instead of manually printing the error and calling `os.Exit(1)` here.

cmd/cluster/support/delete.go
```diff
@@ -120,15 +116,22 @@ func (o *deleteOptions) run() error {
...
 		for _, limitedSupportReason := range limitedSupportReasons {
 			limitedSupportReasonIds = append(limitedSupportReasonIds, limitedSupportReason.ID())
 		}
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Failed to get limited support reasons: %v\n", err)
-			return err
-		}
+	} else if len(limitedSupportReasons) > 1 && o.limitedSupportReasonID == "" {
+		fmt.Fprintf(os.Stderr, "This cluster has multiple limited support reason IDs.\nPlease specify the exact reason ID or the `all` flag \n")
+		os.Exit(1)
```
typeid, line 132: **nitpick:** same as above, return an error and let the main logic handle it.

cmd/cluster/support/delete.go
```diff
@@ -120,15 +116,22 @@ func (o *deleteOptions) run() error {
...
+	} else if len(limitedSupportReasons) == 1 && o.limitedSupportReasonID == "" {
+		limitedSupportReasonIds = append(limitedSupportReasonIds, limitedSupportReasons[0].ID())
 	} else {
 		limitedSupportReasonIds = append(limitedSupportReasonIds, o.limitedSupportReasonID)
```
typeid, line 136: **question:** this should be the same, right? 

I think we could do the following to make the logic easier to understand:
- `o.removeAll` and `(len(limitedSupportReasonIds) > 1 && o.limitedSupportReasonID == "")` is the same state, merge them
- merge the above two else's into one as well (they should be the same)
- move the deletion logic into a new function `deleteLimitedSupportReason` 

E.g. (pseudocode)
```
if len(limitedSupportReasonIds) == 0 {
    return errors.New("cluster is not in limited support")
}

if o.removeAll || (len(limitedSupportReasonIds) > 1 && o.limitedSupportReasonID == "") {
    if !o.removeAll && o.limitedSupportReasonID == "" {
        return errors.New("cluster has multiplate limited support reason IDs, please specify the exact reason ID or use the `all` flag.")
    }

    for _, id := range limitedSupportReasonIds {
        deleteLimitedSupportReason(connection, cluster, id)
    }
} else {
    limitedSupportReasonIds = append(limitedSupportReasonIds, o.limitedSupportReasonID)
    for _, id := range limitedSupportReasonIds {
        deleteLimitedSupportReason(connection, cluster, id)
    }
}
```

typeid (changes requested):

cmd/cluster/support/delete.go
vibhuti-chalise: The if..else is now consolidated and deleteLimitedSupportReason() is also implemented. 
However, I am still unable to delete the "only" LS reason ID in the absence of the -i flag.

typeid: At a quick glance I'm assuming you're running into an error in my pseudocode :D 
```
} else {
    limitedSupportReasonIds = append(limitedSupportReasonIds, o.limitedSupportReasonID)
    for _, id := range limitedSupportReasonIds {
        deleteLimitedSupportReason(connection, cluster, id)
    }
```

The limitedSupportReasonIds would be empty in that case, it needs to actually be filled with the single reason (we can add a check for safety that it indeed is only one). You can get that ID from `limitedSupportReasons, err := getLimitedSupportReasons(o.clusterID)` called earlier.

vibhuti-chalise: this works now.

devppratik:
Tested this PR. The changes summarized would be
For single LS, no `-i` flag is required to delete
For multiple LS, reason needs to be specified using `-i` or `--all` flag to delete all

typeid (approved):
/approve

PR #498, opened 2024-01-08
[WIP] - OSD-15091: Adding detachStuckVolume under cluster subcommand

TheUndeadKing:
The function of detachStuckVolume is to forcefully remove the volume of non running state pod. i.e, migrating Following script under osdctl cluster subcommand.

- https://github.com/openshift/ops-sop/blob/master/v4/utils/detach-stuck-volumes


cmd/cluster/detachstuckvolume.go
```diff
@@ -0,0 +1,166 @@
...
+type detachStuckVolumeOptions struct {
+	clusterID string
+	cluster   *cmv1.Cluster
+}
+
+// detachstuckvolumeCmd represents the detachstuckvolume command
+func newCmdDetachStuckVolume() *cobra.Command {
+	ops := newdetachStuckVolumeOptions()
+	detachstuckvolumeCmd := &cobra.Command{
+		Use:               "detachstuckvolume",
```
devppratik, line 31: `detach-stuck-volume`

cmd/cluster/detachstuckvolume.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	if len(Region) != 1 {
+		fmt.Printf("Got more than one region value: %v", len(Region))
+	}
+	fmt.Println(Region[0])
+
+	fmt.Println(VolumeId)
+	// aws ec2 detach-volume --volume-id $VOLUME_ID --region $REGION --force
+	// WiP - Need to convert above cmd to function once volIdRegion gets completed
```
devppratik, line 100: Is this still a WIP?

cmd/cluster/detachstuckvolume.go
```diff
@@ -0,0 +1,166 @@
...
+	if len(Region) == 0 && len(VolumeId) == 0 {
+		return fmt.Errorf("there's no pv utilized by non running state pod in cluster: %s\nNo action prequired", d.clusterID)
+	}
+
+	if len(Region) != 1 {
+		fmt.Printf("Got more than one region value: %v", len(Region))
+	}
+	fmt.Println(Region[0])
+
+	fmt.Println(VolumeId)
```
devppratik, line 98: If multiple regions are present and it is an error, it would make sense to return from here rather then print

cmd/cluster/detachstuckvolume.go
```diff
@@ -0,0 +1,166 @@
...
+
+	for _, pod := range pods.Items {
+		//if pod.Status.Phase != "Running" { // IMP: We only need to pass the following condition for non running state pods. Will un-comment once the testing is complete.
+		for _, volume := range pod.Spec.Volumes {
+			if volume.PersistentVolumeClaim != nil {
+				for _, pVol := range pV.Items {
+					if pVol.Spec.AWSElasticBlockStore != nil {
+						// Most of the cluster return AWSElasticBlockStore as nil. Could write code, not sure what'll be actual response.
+						fmt.Println("Gathering info from AWSElastic")
+						// Logic -
```
devppratik, line 129: Is this the actual logic or something else needs to be done

cmd/cluster/detachstuckvolume.go
```diff
@@ -0,0 +1,166 @@
...
+	bucket := make(map[string]bool)
+	var result []string
+	for _, str := range s {
+		if _, ok := bucket[str]; !ok {
+			bucket[str] = true
+			result = append(result, str)
+		}
+	}
+	return result
+}
```
devppratik, line 166: This is not a good logic to use. Instead try & figure out to limit the duplicate entry being added

devppratik (changes requested):

cmd/cluster/detachstuckvolume.go
TheUndeadKing: yes

cmd/cluster/detachstuckvolume.go
TheUndeadKing: The jira ticket migration script detach AWSElasticBlockStore id's but as per Florian mentioned in thread conversion that go with cis volume most of the cluster doesn't have AWSElasticBlockStore. Hence i just left blank

cmd/cluster/detachstuckvolume.go
```diff
@@ -0,0 +1,190 @@
...
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/osdctl/pkg/osdCloud"
+	"github.com/openshift/osdctl/pkg/utils"
+	"github.com/spf13/cobra"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+const NameSpace = "openshift-monitoring"
```
supreeth7, line 19:
```suggestion
const Namespace = "openshift-monitoring"
```

cmd/cluster/detachstuckvolume.go
```diff
@@ -0,0 +1,190 @@
...
+
+var Region []string
+var VolumeId []string
```
supreeth7, line 22: Better to amalgamate these into a struct.

cmd/cluster/detachstuckvolume.go
```diff
@@ -0,0 +1,190 @@
...
+
+type detachStuckVolumeOptions struct {
+	clusterID string
+	cluster   *cmv1.Cluster
+}
+
+// detachstuckvolumeCmd represents the detachstuckvolume command
```
supreeth7, line 29: Self-explanatory, not-needed

cmd/cluster/detachstuckvolume.go
```diff
@@ -0,0 +1,190 @@
...
+
+	}
+	pV, err := clientset.CoreV1().PersistentVolumes().List(context.TODO(), v1.ListOptions{})
+
+	if err != nil {
+		return fmt.Errorf("failed to list pv in namespace '%s'", NameSpace)
+
+	}
+
+	for _, pod := range pods.Items {
```
supreeth7, line 144: Should modify the logic - O(n^4) time complexity

supreeth7 (changes requested):
This would require a complete overhaul of the business logic, starting with better variable naming, proper object definitions, and modular functions.

cmd/cluster/detachstuckvolume.go
TheUndeadKing: Performed some modification by only getting required obj using field selector

cmd/cluster/detachstuckvolume.go
TheUndeadKing: Removed above func & replace with **contain** func available in slices pkg

cmd/cluster/detachstuckvolume.go
```diff
@@ -0,0 +1,229 @@
...
+		Short:             "Detach openshift-monitoring namespace's volume from a cluster forcefully",
+		Args:              cobra.NoArgs,
+		DisableAutoGenTag: true,
+		Run: func(cmd *cobra.Command, args []string) {
+			cmdutil.CheckErr(ops.complete(cmd, args))
+			cmdutil.CheckErr(ops.run())
+		},
+	}
+	detachstuckvolumeCmd.Flags().StringVarP(&ops.clusterID, "cluster-id", "c", "", "The internal ID of the cluster to perform actions on")
+	detachstuckvolumeCmd.MarkFlagRequired("cluster-id")
```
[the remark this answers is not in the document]
devppratik: This is because we will require the cluster id anyhow

cmd/cluster/detachstuckvolume.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	defer connection.Close()
+	cluster, err := utils.GetCluster(connection, o.clusterID)
+	if err != nil {
+		return err
+	}
+	o.cluster = cluster
+	o.clusterID = cluster.ID()
+	if strings.ToUpper(cluster.CloudProvider().ID()) != "AWS" {
+		return errors.New("this command is only available for AWS clusters")
```
devppratik, line 78: `return fmt.Errorf("this command is only applicable for ROSA clusters")`

cmd/cluster/detachstuckvolume.go
```diff
@@ -0,0 +1,229 @@
...
+
+func (o *detachStuckVolumeOptions) run() error {
+
+	_, _, clientset, err := getKubeConfig(o.clusterID)
+
+	if err != nil {
+		return fmt.Errorf("failed to retrieve Kubernetes configuration and client for cluster with ID %s: %w", o.clusterID, err)
+	}
+
+	err = volIdRegion(clientset, Namespace, "")
```
devppratik, line 92: Variable naming for `volIdRegion` could be improved

cmd/cluster/detachstuckvolume.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// aws ec2 detach-volume --volume-id $VOLUME_ID --region $REGION --force
+	// WiP - Need to convert above cmd to function once volIdRegion gets completed
+	// Tested till line 107 - Couldn't test below aws function getting priv issue. gig acc doesn't have nessary priv
+
+	ocmClient, err := utils.CreateConnection()
+	if err != nil {
+		return err
+	}
+	defer ocmClient.Close()
```
devppratik, line 121: We already have a client intially in the `complete` function
Instead of starting a new one, we can use the exisitng one for the entire command

cmd/cluster/detachstuckvolume.go
devppratik: The change looks good

PR #501, opened 2024-01-16
OSD-20407: Fix osdctl cluster etcd related commands

devppratik:
Closes - https://issues.redhat.com/browse/OSD-20407

This PR changes
1. `etcd-health-check` is reverted from deprecation
2.  `etcd-health-check` does not elevate to cluster admin
3. `etcd-member-replace` autopopulates the reason for elevation

### DIFF
--- a/cmd/cluster/etcd_health.go
+++ b/cmd/cluster/etcd_health.go
@@ -51,13 +51,17 @@ func (capture *logCapture) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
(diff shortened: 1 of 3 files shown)


mjlshen (approved):
/lgtm

Tof1973:
Honestly, I do not think we should have reason provided without a ticket number...
I'm currently working on another PR in this direction for all command, so here the new change I will do:
https://github.com/openshift/osdctl/pull/502/files#diff-17b060d34ea482394204edb6b428cf9991e98c98c7f287415bdfacb66b14edbf

PR #502, opened 2024-01-18
OSD-20024: Add mandatory reason argument when elevation is required for osdctl

Tof1973:
The goal of this PR is to add reason argument to commands that require elevation.
--as backplane-cluster-admin will not be required in that case as is part of elevation procedure
When osdctl is just displaying comand oc with --as backplane-cluster-admin, those commands are replaced by ocm-backplane elevate.
On top of that this PR will do some cleanup to not propagate flags to commands, as this is not used anymore and could end in "dirty" usage.
Finally, cluster resize infra will now display proper old instance type

### DIFF
--- a/cmd/cluster/access/access.go
+++ b/cmd/cluster/access/access.go
@@ -52,7 +52,7 @@ var (
 func NewCmdAccess(streams genericclioptions.IOStreams, client *k8s.LazyClient) *cobra.Command {
 	ops := newClusterAccessOptions(client, streams)
 	accessCmd := &cobra.Command{
-		Use:               "break-glass <cluster identifier> --reason <OHSS or PD ticket>",
+		Use:               "break-glass <cluster identifier>",
 		Short:             "Emergency access to a cluster",
 		Long:              "Obtain emergency credentials to access the given cluster. You must be logged into the cluster's hive shard",
 		Args:              cobra.ExactArgs(1),
--- a/cmd/cluster/access/access.go
+++ b/cmd/cluster/access/access.go
@@ -49,23 +49,22 @@ var (
 )
 
 // NewCmdCluster implements the 'cluster access' subcommand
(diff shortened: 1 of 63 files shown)


Tof1973:
I put this PR in draft for the moment, as I did not test all modified commands yet, but I already publish it in order to have comments if needed...
Here are commands already successfully tested:
- osdctl cluster resize infra
- osdctl cluster transfer-owner

cmd/cluster/resizecontrolplanenode.go
```diff
@@ -187,6 +197,7 @@ func withRetrySkipCancelOption(fn func() error, procedure string) (err error) {
 	if err == nil {
 		return nil
 	}
+	fmt.Println(err)
```
typeid, line 200: Leftover from debugging?

typeid (commented):
This is a pretty big change. It generally looks good to me, I would be happy if someone else would take a second look.

typeid:
Thanks @Tof1973, I know you put a lot of time into testing it. In the future, it would be great if we could break down bigger PRs if possible, it would help with the review speed. 

/approve

cmd/cluster/access/access.go
```diff
@@ -49,23 +49,22 @@ var (
 )
 
 // NewCmdCluster implements the 'cluster access' subcommand
-func NewCmdAccess(streams genericclioptions.IOStreams, flags *genericclioptions.ConfigFlags) *cobra.Command {
+func NewCmdAccess(streams genericclioptions.IOStreams, client *k8s.LazyClient) *cobra.Command {
+	ops := newClusterAccessOptions(client, streams)
 	accessCmd := &cobra.Command{
-		Use:               "break-glass <cluster identifier>",
+		Use:               "break-glass <cluster identifier> --reason <OHSS or PD ticket>",
```
tnierman, line 55: Flags shouldn't be included in a command's `Use` field

Tof1973: removed

PR #504, opened 2024-01-22
promote: Allow providing app-interface checkout directory

NautiluX:
* Add argument to pass in app-interface checkout directory
* Use default location if none is passed in
* Don't silently fail if osdctl is called outside of a git repository

Jira: [OSD-20520](https://issues.redhat.com//browse/OSD-20520)

### DIFF
--- a/cmd/promote/git/app_interface.go
+++ b/cmd/promote/git/app_interface.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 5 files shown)


cmd/promote/git/app_interface.go
```diff
@@ -23,21 +24,49 @@ type Service struct {
 	} `yaml:"resourceTemplates"`
 }
 
-func BootstrapOsdCtlForAppInterfaceAndServicePromotions() {
-	_, err := getBaseDir()
-	if err != nil {
-		log.Fatal(err)
+type AppInteface struct {
```
AlexVulaj, line 27: small typo

NautiluX: thanks
fixed the type typo

PR #505, opened 2024-01-31
Update backplane-cli, osd-network-verifier

mjlshen:
* backplane-cli v0.1.23 makes assume-initial-arn optional
* osd-network-verifier v0.4.5 Reduces the risk of leaking security groups

[OSD-20606](https://issues.redhat.com//browse/OSD-20606)


typeid (approved):
/lgtm
/approve

PR #506, opened 2024-02-02
Fix bug to allow posting bulk servicelogs

mjlshen:
* Don't append cluster-id searches to ocm filter searches
* Added examples and documentation to the command
* Adjusted the command to return errors instead of relying on log.Fatal
* Only check for recent servicelogs when there's only one cluster

Current state:
```bash
❯ osdctl servicelog post -q "cloud_provider.id is 'gcp' and ccs.enabled is 'true' and managed='true' and state is 'ready'" -t tmp.json                           
error: no cluster identifier has been found 
```
which was happening because the command was attempting to `CheckServiceLogsLastHour()` even though no cluster id was supplied to the command. So I moved this check to after the list of clusters have been identified and only check if we're sending servicelogs to one cluster because the output would be too much for bulk service logs.

After:
```bash
❯ ~/git/openshift/osdctl/osdctl servicelog post -q "cloud_provider.id is 'gcp' and ccs.enabled is 'true' and managed='true' and state is 'ready'" -t tmp.json
The current version () is different than the latest released version (v0.24.0).It is recommended that you update to the latest released version to ensure that no known bugs or issues are hit.
Continue? (y/N): y
INFO[0003] The following clusters match the given parameters: 
${CLUSTERS}

INFO[0003] The following template will be sent:
${TEMPLATE}      
```

Noticed while working on [OSD-20579](https://issues.redhat.com//browse/OSD-20579)

### DIFF
--- a/cmd/servicelog/post.go
+++ b/cmd/servicelog/post.go
@@ -11,7 +11,6 @@ import (
 	"strings"
 	"time"
 
-	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/utils/strings/slices"
 
 	"github.com/openshift-online/ocm-cli/pkg/arguments"
@@ -35,6 +34,7 @@ type PostCmdOptions struct {
 	TemplateParams  []string
 	filterFiles     []string // Path to filter file
 	filtersFromFile string   // Contents of filterFiles
+	filterParams    []string
 	isDryRun        bool
 	skipPrompts     bool
 	clustersFile    string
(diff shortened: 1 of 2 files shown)


cmd/servicelog/post.go
```diff
@@ -146,54 +154,69 @@ func (o *PostCmdOptions) Run() error {
...
+		o.filterParams = append(o.filterParams, filters)
 	}
+
+	// Combine existing OCM filters with any cluster id-related flags
 	var queries []string
 	if o.clustersFile != "" {
 		contents, err := o.accessFile(o.clustersFile)
 		if err != nil {
-			log.Fatalf("Cannot read file %s: %q", o.clustersFile, err)
+			return fmt.Errorf("cannot read file %s: %v", o.clustersFile, err)
```
typeid, line 171: Any specific reason to not wrap the error here?
```suggestion
			return fmt.Errorf("cannot read file %s: %w", o.clustersFile, err)
```

cmd/servicelog/post.go
```diff
@@ -146,54 +154,69 @@ func (o *PostCmdOptions) Run() error {
...
 		}
-		err = o.parseClustersFile(contents)
-		if err != nil {
-			log.Fatalf("Cannot parse file %s: %q", o.clustersFile, err)
+		if err := o.parseClustersFile(contents); err != nil {
+			return fmt.Errorf("cannot parse file %s: %v", o.clustersFile, err)
```
typeid, line 174: Same as above
```suggestion
			return fmt.Errorf("cannot parse file %s: %w", o.clustersFile, err)
```

cmd/servicelog/post.go
```diff
@@ -146,54 +154,69 @@ func (o *PostCmdOptions) Run() error {
...
+			if !ocmutils.ConfirmPrompt() {
+				return nil
+			}
+		}
 	}
 
 	log.Infoln("The following template will be sent:")
 	if err := o.printTemplate(); err != nil {
-		log.Errorf("Cannot read generated template: %q", err)
+		return fmt.Errorf("cannot read generated template: %v", err)
```
typeid, line 219:
```suggestion
		return fmt.Errorf("cannot read generated template: %w", err)
```

cmd/servicelog/post.go
mjlshen: Ah, just that we don't normally unwrap errors, but no real reason not to prevent it :D

typeid (approved):
/lgtm
/approve

PR #508, opened 2024-02-06
Added a check to "osdctl network verify-egress" for public subnets supplied by user with --subnet-id

aliceh:
In this commit, I added a check when a subnet is supplied by a user with --sibnet-id. In the code, I check whether a subnet is public with isSubnetPublic function that checks if there is an internet gateway attached to the subnet and whether the RouteTable for the subnet has a default route to 0.0.0.0/0.


aliceh:
cancel /hold

cmd/network/verification.go
```diff
@@ -429,6 +439,101 @@ func (e *EgressVerification) getSubnetIds(ctx context.Context) ([]string, error)
 	// For non-PrivateLink BYOVPC clusters, provided subnets are 50/50 public/private subnets, so make the user decide for now
 	// TODO: Figure out via IGW/NAT GW/Route Tables
 	return nil, fmt.Errorf("unable to determine which non-PrivateLink BYOVPC subnets are private yet, please check manually and provide the --subnet-id flag")
+
+}
+
+// This function checks the gateway attached to the subnet and returns true if the subnet starts with igw- (for InternetGateway) and has a route to 0.0.0.0/0
+func (e *EgressVerification) isSubnetPublic(ctx context.Context, subnetID string) (bool, error) {
```
fahlmant, line 446: Some of the logic here looks exactly the same as the logic in https://github.com/fahlmant/osdctl/blob/master/cmd/cluster/cpd.go#L125

Is it possible to pull out the shared logic into a single function?

fahlmant, line 446: This code seems to be similar to https://github.com/openshift/osdctl/blob/master/cmd/cluster/cpd.go#L125

Can you pull out the common logic into a separate shared function?

fahlmant: I linked the other function above and that function and this function have most of the same code. I would recommend creating a third function, something called `findRouteTableForSubnet` and placing all the shared code there, then calling it within isSubnetPublic and the other fucntion I linked

aliceh: Where should I put this shared function? A new file in /pkg/utils ?  Just want to make sure I understand what the ask is.

aliceh: I am thinking /pkg/utils/utils

abyrne55:
Ran into this while on-call today. Any chance we can get this merged soon? @aliceh

aliceh:
I'll look at it today.

cmd/network/verification.go
```diff
@@ -131,6 +131,7 @@ func NewCmdValidateEgress() *cobra.Command {
 type egressVerificationAWSClient interface {
 	DescribeSubnets(ctx context.Context, params *ec2.DescribeSubnetsInput, optFns ...func(options *ec2.Options)) (*ec2.DescribeSubnetsOutput, error)
 	DescribeSecurityGroups(ctx context.Context, params *ec2.DescribeSecurityGroupsInput, optFns ...func(options *ec2.Options)) (*ec2.DescribeSecurityGroupsOutput, error)
```
AlexVulaj, line 133: From the looks of this code, I think we can blow away this function from the interface.

cmd/network/verification.go
AlexVulaj: I spent some time digging into this issue this morning and speaking with @mjlshen in [this slack thread](https://redhat-internal.slack.com/archives/C04PUGBE7JN/p1715951689864719)- I think the right path forward here is to leave this for now and have a future effort to clean up all of the functions in [this interface](https://github.com/openshift/osdctl/blob/f475456b95544fce30378f4a7684903e0626efb1/pkg/provider/aws/client.go#L47) such that they don't hide the `context` parameter. Then our wrapper interface will be directly compatible with base AWS clients while still allowing all of the helper functions we provide still.

Curious to hear other thoughts, as I think a refactor here would be a bit large for this PR.

AlexVulaj: All the better - I'm also perfectly happy deprecating the wrapper interface as a whole and directly using clients from the aws go sdk :)

cmd/network/verification.go
aliceh: Thanks, I will remove it!

AlexVulaj: Looks like I missed that somehow - disregard this.

cmd/network/verification_test.go
```diff
@@ -142,6 +147,18 @@ func Test_egressVerificationGenerateAWSValidateEgressInput(t *testing.T) {
 							},
 						},
 					},
+					describeRouteTablesResp: &ec2.DescribeRouteTablesOutput{
```
AlexVulaj, line 150: I think a few other tests also need this `describeRouteTablesResp` property defined now so they can pass.

aliceh: Done✅

typeid:
As an FYI, this is the CAD implementation: https://github.com/openshift/configuration-anomaly-detection/pull/276/files

PR #509, opened 2024-02-07
OSD-20697: Fetch Dynatrace URL via API Call

devppratik:
We now have a label in the OCM Get API call for MC Clusters which point to the DT Tenant being used. 
This change adds the ability to directly fetch the Dynatrace Tenant URL without logging into the cluster

Note: The original method to fetch the URL is moved as a fallback method


NautiluX (commented):
/lgtm

PR #515, opened 2024-02-19
OSD-21061: fix `osdctl cluster resize-control-plane-node` when KUBECONFIG is not set

Tof1973:
When running `osdctl cluster resize-control-plane-node` without KUBECONFIG variable set, it will end with an error:
```
Force draining node... This might take a minute or two...
failed to force drain:
no current kubeconfig context 
```
This PR will fix this issue.

In the future we will be able to remove this 'hack' when backplane-cli will not override KUBECONFIG variable anymore...

### DIFF
--- a/cmd/cluster/resizecontrolplanenode.go
+++ b/cmd/cluster/resizecontrolplanenode.go
@@ -105,7 +105,7 @@ func (o *resizeControlPlaneNodeOptions) complete(cmd *cobra.Command, _ []string)
 		machine type doesn't exist and can be re-run.
 	*/
 
-	// As RunElevate is unsetting KUBECONFIG, we need to store its value in order to redefine it if it was defined
+	// As RunElevate is overriding KUBECONFIG, we need to store its value in order to redefine it to its original (value or unset)
 	o.kubeconfig, o.kubeconfigDefined = os.LookupEnv("KUBECONFIG")
 
 	return nil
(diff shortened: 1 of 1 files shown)


Tof1973:
Tested in staging ;)

fahlmant:
Can you update the description?

cmd/cluster/resizecontrolplanenode.go
```diff
@@ -253,6 +253,8 @@ func (o *resizeControlPlaneNodeOptions) forceDrainNode(nodeID string, reason str
 	// As RunElevate is unsetting KUBECONFIG, we need to redefined it if it was defined
```
fahlmant, line 253: Could `os.Unsetenv("KUBECONFIG")` be placed above all the if statements?

Tof1973: Sorry, I do not see the point of unsetting the variable if we set it to a value?
And I guess it is more readable, as:
 if variable was set,
 then we set it to old value,
 else we unset it to retrieve the original configuration...

typeid, line 253: **Question:** Why do we need to unset it if `!o.kubeconfigDefined`? The comment isn't clear on that. 

nitpick: (repeating):
```suggestion
	// As RunElevate is unsetting KUBECONFIG, we need to redefine it if it was defined
```

typeid (changes requested):
PR does not make it clear to me yet why we run into the bug. Could we be more precise in the code comments, so that we don't accidentally throw out the seemingly NOOP logic?

typeid (approved):
/lgtm
/approve

cmd/cluster/resizecontrolplanenode.go
fahlmant: Instead of copying the same line of code three places, you can use it once and get rid of unnecessary elses.

PR #518, opened 2024-02-20
OSD-16848: Migrate ops-sop alert script as osdctl alert command

itsmitul9:
hiding the cluster id and useinfo for security reason 

$ osdctl alert list cluster-id
No alert level specified. Defaulting to 'all'.
Alert Information:
  AlertName:  Watchdog
  Severity:   none
  State:      active
  Message:    An alert that should always be firing to certify that Alertmanager is working properly.

$ osdctl alert list cluster-id --level all
Alert Information:
  AlertName:  Watchdog
  Severity:   none
  State:      active
  Message:    An alert that should always be firing to certify that Alertmanager is working properly.

$ osdctl alert list cluster-id --level info
No such Alert found with requested "info" severity

$ osdctl alert silence add cluster-id --alertname Watchdog --duration 1h --comment "Add silence for 1h" 
Alert "Watchdog" has been silenced with id "97151424-cc13-4703-9506-ee295b1ac3c8 "for duration of 1h by user "hide-user-id"

$ osdctl alert silence list cluster-id
Silence Information:
-------------------------------------------
  SilenceID:		97151424-cc13-4703-9506-ee295b1ac3c8
  Created By:		nobody
  Starts At:		2024-03-06T14:26:38.851Z
  Ends At:		2024-03-06T15:26:38.844Z
  Comment:		Add silence for 1h
  AlertName:		Watchdog
---------------------------------------------
$ osdctl alert silence expire cluster-id
No valid option specified. Using a default option
SilenceID "97151424-cc13-4703-9506-ee295b1ac3c8" expired successfully.
All SilenceID expired successfully

$ osdctl alert silence add cluster-id --all --duration 1h --comment "Add silence for 1h"                                   
All alerts for cluster mituxx has been silenced with id "053cd06d-70a2-4216-9017-2fff43b6af6e "for a duration of 1h by user <hide-user-id>

$ osdctl alert silence list cluster-id
Silence Information:
-------------------------------------------
  SilenceID:		053cd06d-70a2-4216-9017-2fff43b6af6e
  Created By:		nobody
  Starts At:		2024-03-06T14:30:03.533Z
  Ends At:		2024-03-06T15:30:03.516Z
  Comment:		Add silence for 1h
  AlertName:		<cluster-id>
---------------------------------------------
osdctl % ./osdctl alert silence expire cluster-id --all
 
SilenceID "053cd06d-70a2-4216-9017-2fff43b6af6e" expired successfully.
All SilenceID expired successfully.

$ osdctl alert silence expire cluster-id --silence-id f11a11fb-9591-4335-a771-e449d9284f3c
Requested SilenceID "f11a11fb-9591-4335-a771-e449d9284f3c" expired successfully.

$ osdctl alert silence list cluster-id       
 Silence Information:
No silences found, all silence has been cleared


Tof1973:
Shouldn't we have the old command set to deprecated instead of just deleted, like 
[@kirk -vvv](https://redhat-internal.slack.com/team/USPJDCA5T)
 was mentioning here: https://redhat-internal.slack.com/archives/C04PUGBE7JN/p1708463884513639?thread_ts=1708463665.277959&cid=C04PUGBE7JN ?

iamkirkbater:
/label ok-to-test

I'd love to see some unit tests added to these commands if there's any possibility to have that happen :)

As for the deprecation, we'd need to deprecate the old command in the SOP repo, and I think at this point until we can confirm a 1:1 functionality match we should leave both in place until we are confident that this is bug free. We can add a WARNING output to the sop script, though, :D

With ALL of that said I took a quick glance at the code and from a high-level pass this looks great! I wasn't able to get too deep into the details, so I'll leave any approvals for others to review as well.

Thanks so much for this contribution!

cmd/alerts/add_silence.go
```diff
@@ -0,0 +1,49 @@
...
+	"log"
+
+	"github.com/spf13/cobra"
+)
+
+func NewCmdAddSilence() *cobra.Command {
+	return &cobra.Command{
+		Use:   "add-silence <cluster-id> <alertname1> <alertname2>... <comment>",
+		Short: "add a new silence",
+		Long:  `add new silence`,
```
AlexVulaj, line 14: For the long description, I'd like to see some more details about the command. What does it mean to "add a new silence"? How long does it last for? Etc.

cmd/alerts/add_silence.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"fmt"
+	"log"
+
+	"github.com/spf13/cobra"
+)
+
+func NewCmdAddSilence() *cobra.Command {
+	return &cobra.Command{
+		Use:   "add-silence <cluster-id> <alertname1> <alertname2>... <comment>",
```
AlexVulaj, line 12: For data like this, I'd prefer we use flags instead of multiple args. You can even declare a "slice" flag that takes multiple values for `alertname`

cmd/alerts/add_silence.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/spf13/cobra"
+)
+
+func NewCmdAddSilence() *cobra.Command {
+	return &cobra.Command{
```
AlexVulaj, line 11: Can we add `DisableAutoGenTag: true,` to this command for consistency with the others?

cmd/alerts/common.go
```diff
@@ -0,0 +1,95 @@
...
+
+type logCapture struct {
+	buffer bytes.Buffer
+}
+
+func (capture *logCapture) GetStdOut() string {
+	return capture.buffer.String()
+}
+
+func (capture *logCapture) Write(p []byte) (n int, err error) {
```
AlexVulaj, line 32: This function returns an error, but it can only ever be nil. My vote would be to either

a.) Get the error from the `WriteString` call and check that
b.) not return an error value at all

cmd/alerts/common.go
```diff
@@ -0,0 +1,95 @@
...
+
+	clientset, err := kubernetes.NewForConfig(kubeconfig)
+	if err != nil {
+		log.Fatalf("failed to create clientset : %v", err)
+	}
+
+	return kubeconfig, clientset, err
+}
+
+func ExecInPod(kubeconfig *rest.Config, clientset *kubernetes.Clientset, localHostUrl string, cmd []string, podName string) (string, error) {
```
AlexVulaj, line 58: I notice that "localHostUrl" and "podName" aren't used here - is this function still in progress or can we remove them?

cmd/alerts/list_alerts.go
```diff
@@ -0,0 +1,124 @@
...
+			alertCmd.clusterID = args[0]
+			ListAlerts(alertCmd)
+		},
+	}
+
+	newCmd.Flags().StringVarP(&alertCmd.alertLevel, "level", "l", "", "Alert level [warning, critical, firing, pending, all]")
+	return newCmd
+}
+
+func alertLevel(level string) string {
```
AlexVulaj, line 56: Is this function going to be used eventually?

cmd/alerts/list_alerts.go
```diff
@@ -0,0 +1,124 @@
...
+	case "firing":
+		levelCmd = "firing"
+	case "pending":
+		levelCmd = "pending"
+	case "all":
+		levelCmd = "all"
+	default:
+		log.Fatalf("Invalid alert level: %s\n", level)
+		return ""
+	}
```
AlexVulaj, line 71: This is completely personal, subjective preference, but if you wanted you could rewrite this switch as follows.
```suggestion
	switch level {
	case "warning", "critical", "firing", "pending", "all":
		levelCmd = level
	default:
		log.Fatalf("Invalid alert level: %s\n", level)
		return ""
	}
```

cmd/alerts/list_alerts.go
```diff
@@ -0,0 +1,124 @@
...
+	var alerts []Alert
+
+	defer func() {
+		if err := recover(); err != nil {
+			log.Fatal("error : ", err)
+		}
+	}()
+
+	clusterID := cmd.clusterID
+	levelcmd := cmd.alertLevel
```
AlexVulaj, line 85: nitpicky - I recommend declaring variables no sooner than they're needed. For example, the instantiation of `levelcmd` here could be moved down to line 99/100. It doesn't make as much of a difference with static values like this, but it's a good habit to get into for assignments that could have side effects.

cmd/alerts/list_silence.go
```diff
@@ -0,0 +1,70 @@
...
+		fmt.Println("Error in unmarshaling the data", err)
+	}
+
+	for _, s := range silence {
+		id, matchers := s.ID, s.Matchers
+		for _, matcher := range matchers{
+			fmt.Printf("Found %v %v with silence id %s\n",matcher.Name, matcher.Value, id) 
+		}
+	}
+	return id
```
AlexVulaj, line 69: `id` will always be empty-string here, as the `id` on line 64 is a new variable that shadows the original initialization.

devppratik:
/retitle OSD-16848: Migrate ops-sop alert script as osdctl alert command

cmd/alerts/add_silence.go
itsmitul9: Added 
```
cmd.Flags().StringSliceVar(&addSilenceCmd.alertID, "alertname", []string{}, "alertname (comma-separated)")
cmd.Flags().StringVarP(&addSilenceCmd.comment, "comment","c","","add comment about alertname")
```

cmd/alerts/common.go
itsmitul9: @AlexVulaj Does it looks good ?

`func (capture *logCapture) Write(p []byte) (n int, err error) {
	a := string(p)
	_, err = capture.buffer.WriteString(a)
	return len(p), err
}`

cmd/alerts/list_silence.go
itsmitul9: @AlexVulaj I was planning to use this `id` in clear_silence.go file , feed the output from id to clear silence function, seems that's not working, will find alternative.

cmd/alerts/list_alerts.go
itsmitul9: @AlexVulaj I don't get it the full context here, `clusterID` and `levelcmd` as argument in next line, if we declare in line 99/100 that would be calling function before declaration and code will through warning ?

cmd/alerts/list_silence.go
```diff
@@ -51,16 +58,47 @@ func ListSilence(clusterID string){
...
+			fmt.Printf("	SilenceID:	%s\n", id)
+			fmt.Printf("	Status:	%s\n", status)
+			fmt.Printf("	Created By:	%s\n", created)
+			fmt.Printf("	Starts At:	%s\n", starts)
+			fmt.Printf("	Ends At:	%s\n", end)
+			fmt.Printf("	Comment:	%s\n", comment)
+			fmt.Printf("	AlertName:	%s\n", matcher.Value)
+		}
+	fmt.Println("--------------------------")
+}
```
NautiluX, line 103: can we remove this function if it's not used?

cmd/alerts/list_alerts.go
```diff
@@ -63,12 +64,14 @@ func ListAlerts(cmd *alertCmd) {
...
+	if levelcmd == "warning" || levelcmd == "critical" || levelcmd == "firing" || levelcmd == "pending" {
 		levelCmd = levelcmd
+	} else if levelcmd == "all" {
+		levelCmd = "all"
 	} else {
 		log.Fatalf("Invalid alert level: %s\n", levelcmd)
-	}	
-		
+	}
+
```
NautiluX, line 74: Any chance to simplify this if-statement to make it more readable?

cmd/alerts/add_silence.go
```diff
@@ -3,53 +3,99 @@ package alerts
...
 		log.Fatal(err)
 	}
 
+	if all {
+		AddAllSilence(clusterID, duration, comment, username, kubeconfig, clientset)
+	} else if len(alertID) > 0 {
+		AddAlertNameSilence(alertID, duration, comment, username, kubeconfig, clientset)
+	} else {
+		fmt.Println("No valid option specified. Use --all or --alertname.")
+	}
```
NautiluX, line 65: could we provide a good default here? --all?

iamkirkbater (changes requested):
I'm wondering if we want to make some usability considerations here.

I'm only requesting these changes to invoke the discussion, if we want to push forward with this as is I won't hold it up, but I'd like to start adding some consistency to osdctl, even though it hasn't "officially" been documented anywhere.

I also want to add here, before I begin, that this is not a critique on the code written, but the organization of the command structure. The code as written looks great from my quick review.

But I specifically want to call out our flag and command structure strategy here.

I'd much rather see `--silence-id` than `--silenceID` as that's more "conventional" when it comes to flags. To be explicitly clear, having the variables named silenceID within the go code is fine, I'm looking at the flags we expose to the users of the CLI tool.

Additionally, could we organize the command into a structure more like the following:

* `osdctl alert silence add`
* `osdctl alert silence list`
* `osdctl alert silence expire`

There are two reasons here:
1. This allows us to expand the `silence` set of actions in the future if necessary without having to continually add additional `verb-silence` commands. While this may not be explicitly necessary today, it's an overall better "pattern" IMO to follow.
2. I'd like to try to remain as close to the `amtool` commands as possible with our conventions. Since we're wrapping that functionality I'd like to make it as similar as possible so the two commands are both more intuitive to use if you're familiar with one or the other already.

This is really great so far! Thanks!

cmd/alerts/list_alerts.go
bergmannf: As it's related to `levelcmd` - I dislike that this function as essentially the same variable with 2 different capitalizations being the only difference:
- `levelCmd`
- `levelcmd`
These also seem to be used to store the same thing (the level of alerts to list), so please only use a single variable for it.
I would also recommend calling it `alertLevel` instead maybe, as the `cmd` prefix made me think it would store some shell-command to execute.

cmd/alerts/list_alerts.go
```diff
@@ -0,0 +1,117 @@
...
+		fmt.Println("No alert level specified. Defaulting to 'all'")
+		levelcmd = "all"
+	} else if levelcmd == "warning" || levelcmd == "critical" || levelcmd == "firing" || levelcmd == "pending" || levelcmd == "info" || levelcmd == "none" || levelcmd == "all" {
+		levelCmd = levelcmd
+	} else {
+		fmt.Printf("Invalid alert level \"%s\" \n", levelcmd)
+		return
+	}
+
+	ListAlertCmd := []string{"amtool", "--alertmanager.url", silence.LocalHostUrl, "alert", "-o", "json"}
```
bergmannf, line 77: Starting with a capital letter makes this look like a global variable - which it doesn't seem to be.

cmd/alerts/list_alerts.go
```diff
@@ -0,0 +1,117 @@
...
+		if err := recover(); err != nil {
+			log.Fatal("error : ", err)
+		}
+	}()
+
+	clusterID := cmd.clusterID
+	levelcmd := cmd.alertLevel
+
+	if levelcmd == "" {
+		fmt.Println("No alert level specified. Defaulting to 'all'")
```
bergmannf, line 68: IMO debug style messages should be output to stderr instead of stdout.
We might also want to use some kind of verbosity level to not print in all cases.

cmd/alerts/silence/clear_silence.go
```diff
@@ -0,0 +1,129 @@
...
+		if countsilence == 0 {
+			fmt.Println()
+			fmt.Printf("All SilenceID expired successfully.\n")
+			return
+		}
+	}
+}
+
+func ClearSilenceByID(silenceID []string, kubeconfig *rest.Config, clientset *kubernetes.Clientset) {
+	for _, silenceId := range silenceID {
```
bergmannf, line 112: Might be nicer to call the argument `silenceIDs` instead of just switching capitalization to get two variables with nearly the same name.

cmd/alerts/silence/common.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+const (
+	AccountNamespace = "openshift-monitoring"
+	ContainerName    = "alertmanager"
+	LocalHostUrl     = "http://localhost:9093"
+	PodName          = "alertmanager-main-0"
+)
+
+type logCapture struct {
```
bergmannf, line 24: Could this be a type alias instead? I don't see the wrapping adding much value.

cmd/alerts/silence/common.go
```diff
@@ -0,0 +1,95 @@
...
+	buffer bytes.Buffer
+}
+
+func (capture *logCapture) GetStdOut() string {
+	return capture.buffer.String()
+}
+
+func (capture *logCapture) Write(p []byte) (n int, err error) {
+	a := string(p)
+	_, err = capture.buffer.WriteString(a)
```
bergmannf, line 34: Is there a reason this is not just using the bytes.Buffer [Write](https://pkg.go.dev/bytes#Buffer.Write) that writes the []bytes directly?

devppratik:
All changes look good. Tested as well on a stg cluster
Thank you so much for your contribution @itsmitul9!!

/lgtm

NautiluX (commented):
/approve

PR #519, opened 2024-02-20
requires reason when using validate-pull-secret with elevate flag

Tof1973:
Following [PR#488](https://github.com/openshift/osdctl/pull/488), that introduce the new elevate flag for validate-pull-secret, we should ask for the reason if the elevate flag is used.

### DIFF
--- a/cmd/cluster/validatepullsecret.go
+++ b/cmd/cluster/validatepullsecret.go
@@ -49,8 +49,9 @@ In case the managed-script fails, --elevate can be added to get the pull-secret 
 			cmdutil.CheckErr(ops.run())
 		},
 	}
-	validatePullSecretCmd.Flags().BoolVar(&ops.elevate, "elevate", false, "get pull-secret with backplane-cluster-admin without running a managed-script")
-	validatePullSecretCmd.Flags().StringVar(&ops.reason, "reason", "", "The reason for this command, which requires elevation, to be run (usualy an OHSS or PD ticket)")
+	validatePullSecretCmd.Flags().BoolVar(&ops.elevate, "elevate", false, "get pull-secret with backplane-cluster-admin without running a managed-script, mandatory when a reason is provided")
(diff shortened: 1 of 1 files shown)


cmd/cluster/validatepullsecret.go
```diff
@@ -94,6 +95,9 @@ func (o *validatePullSecretOptions) run() error {
 
 func (o *validatePullSecretOptions) getPullSecretFromCluster() (email string, err error, sentSL bool) {
 	if o.elevate {
+		if len(o.reason) == 0 {
+			return "", errors.New("please provide a reason when using elevate"), false
+		}
```
devppratik, line 100: Can we show a prompt something like
`Reason not provided. Continue? (y/n)`

Tof1973: That would not be consistent with other commands that requires reason, and what would be the benefit of not providing the reason in that case ? We just need to provide the ref to the PD/OHSS ticket ;)

cmd/cluster/validatepullsecret.go
```diff
@@ -94,6 +95,9 @@ func (o *validatePullSecretOptions) run() error {
 
 func (o *validatePullSecretOptions) getPullSecretFromCluster() (email string, err error, sentSL bool) {
 	if o.elevate {
+		if len(o.reason) == 0 {
```
AlexVulaj, line 98: Instead of doing this check here, I believe we could just add `validatePullSecretCmd.MarkFlagsRequiredTogether(...)` after line 54. That way we let Cobra handle the logic and keep our code cleaner. Thoughts?

Tof1973: Great, that was what I was looking for, but did not find it! Will change that ;) Thanks...
/hold

Tof1973: Implemented as you suggested.
The only drawback is that if we specify --elevate=false, then --reason will also be mandatory, but I guess we will ever do that, and it is not really an issue...

PR #521, opened 2024-02-21
Fix OSD-21049 by updating the hive api dependencies.

wanghaoran1988:
This fix OSD-21049 by updating the hive api dependencies.


typeid (approved):
/lgtm
/approve

PR #522, opened 2024-02-21
remove the hack for RunElevate as it is fixed in backplane-api

Tof1973:
A hack was introduced to workaround a bug in RunElevate from backplane-cli.
The bug is now fixed (https://github.com/openshift/backplane-cli/pull/346), so we can remove the hack safely.

Jira: [OSD-21061](https://issues.redhat.com//browse/OSD-21061)


Tof1973:
For sure it has been tested with KUBECONFIG variable set and not set ;)

NautiluX (commented):
/lgtm

PR #528, opened 2024-02-29
OSD-20998 make infra resize more resilient to temporary error

Tof1973:
When doing resize, we identified, in some circonstances, that some checks in waiting loops may fail. Those check errors are not a real pb and we can just display and skip them as we are in a retry loop.
This PR is doing this.
I have tested the resize, but were not able to reproduce a chek error, as I guess it happen when we are in heavy loaded clusters, but I guess this change will avoid most (if not all) issues we had so far...

### DIFF
--- a/cmd/cluster/resize/infra_node.go
+++ b/cmd/cluster/resize/infra_node.go
@@ -122,7 +122,8 @@ func (r *Resize) RunInfra(ctx context.Context) error {
 		nodes := &corev1.NodeList{}
 
 		if err := r.client.List(ctx, nodes, &client.ListOptions{LabelSelector: selector}); err != nil {
-			return false, err
+			log.Printf("error retrieving nodes list, continuing to wait: %s", err)
+			return false, nil
 		}
 
 		readyNodes := 0
@@ -193,7 +194,8 @@ func (r *Resize) RunInfra(ctx context.Context) error {
 			if apierrors.IsNotFound(err) {
 				return true, nil
 			}
-			return false, err
+			log.Printf("error retrieving machines list, continuing to wait: %s", err)
+			return false, nil
 		}
 
 		log.Printf("original machinepool %s/%s still exists, continuing to wait", originalMp.Namespace, originalMp.Name)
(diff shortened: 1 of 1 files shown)


cmd/cluster/resize/infra_node.go
```diff
@@ -536,3 +543,30 @@ func (r *Resize) nodesMatchExpectedCount(ctx context.Context, labelSelector labe
...
+// in case we have a function that return a bool statis and an error, we can use following helper
+// f beeing a function returning (bool, error), replace
+//
+//	return f(...)
+//
+// by
+//
+//	return skipError(wrapResult(f(...)), "message to context the error"
+//
+// and then the return will always have error set to nil, but a continuing message will be dispalyed in case of error
```
tnierman, line 557:
```suggestion
// having an error when being in a rety loop, should not be handled as an error, and we should just display it and continue
// in case we have a function that return a bool status and an error, we can use following helper
// f being a function returning (bool, error), replace
//
//	return f(...)
//
// by
//
//	return skipError(wrapResult(f(...)), "message to context the error"
//
// and then the return will always have error set to nil, but a continuing message will be dispalyed in case of error
```

cmd/cluster/resize/infra_node.go
```diff
@@ -536,3 +543,30 @@ func (r *Resize) nodesMatchExpectedCount(ctx context.Context, labelSelector labe
...
+func wrapResult(condition bool, err error) result {
+	return result{condition, err}
+}
+
+func skipError(res result, msg string) (bool, error) {
+	if res.err != nil {
+		log.Printf("%s, continuing to wait: %s", msg, res.err)
+	}
+	return res.condition, nil
+}
```
tnierman, line 572: Would it make sense to combine `wrapResult` and `skipError`, if the recommendation is to always call it like 

```golang
return skipError(wrapResult(f(...)), "message to context the error"
```

Tof1973: No, this is not really possible as goal of wrapResult is to convert the multi value return of f function in a struct in order to pass it to skipError, that takes another string argument.
Now, if there is better solution, I would be happy to learn, because indeed I do not like it much ;)

PR #530, opened 2024-03-01
Print version warnings to stderr

geowa4:
This change prints the version check warnings to stderr so it isn't swallowed in a pipeline.


fahlmant (approved):
/lgtm

PR #534, opened 2024-03-14
OSD-19982: osdctl improve limited support workflow

vibhuti-chalise:
corresponds to Jira issue https://issues.redhat.com/browse/OSD-19982
Summary of the changes are as below:
1. The current LS workflow requires 3 flags (problem, resolution & misconfiguration) to be compulsory. The flag evidence is option. Based on the description for [OSD-19882](https://issues.redhat.com/browse/OSD-19982), this PR extends to use templates like [ScheduledChaosTest](https://github.com/openshift/managed-notifications/blob/master/osd/limited_support/scheduled_chaos_test.json) or DR Test to be directly used without all problem, resolution,  misconfiguration & evidence flag. 
2. Based on AC for [OSD-19470](https://issues.redhat.com/browse/OSD-19470), we will also add a template that SRE can use in general to add the LS (Templates can be added separately)
3. This does not remove the current way the command works. It is still valid and we will look for suggestions from SRE what to keep/change.
4. User cannot use problem, resolution, evidence & misconfiguration flags  when using the --template flag.


devppratik:
/rename OSD-19982: osdctl improve limited support workflow

devppratik:
/retitle OSD-19982: osdctl improve limited support workflow

cmd/cluster/support/post.go
```diff
@@ -58,33 +74,66 @@ The cluster has a second failing ingress controller, which is not supported and
...
+	case ".", "?", "!":
+		return errors.New("--resolution should not end in punctuation")
+	}
+	return nil
 }
 
-func validateResolutionString(res string) error {
-	if res[len(res)-1:] == "." {
-		return errors.New("resolution string should not end with a `.` as it is already added in the email template")
+func (p *Post) check() error {
```
devppratik, line 109:
```suggestion
func validateResolutionString(res string) error {
	if res[len(res)-1:] == "." {
		return errors.New("resolution string should not end with a `.` as it is already added in the email template")
	}
	return nil
}

func (p *Post) check() error {
```

cmd/cluster/support/post.go
```diff
@@ -58,33 +74,66 @@ The cluster has a second failing ingress controller, which is not supported and
...
+			return fmt.Errorf("\nIf Template flag is present, all three --problem, --resolution and --misconfiguration flags cannot be used")
+		}
+		p.parseUserParameters() // parse all the '-p' user flags
+
+		// For every '-p' flag, replace its related placeholder in the template
+		for k := range userParameterNames {
+			p.replaceFlags(userParameterNames[k], userParameterValues[k])
+		}
+	} else {
+		if p.Problem == "" || p.Resolution == "" || p.Misconfiguration == "" {
```
devppratik, line 121:
```suggestion
	} else {
			if err := validateResolutionString(p.Resolution); err != nil {
			return err
		}
		if p.Problem == "" || p.Resolution == "" || p.Misconfiguration == "" {
```

cmd/cluster/support/post.go
```diff
@@ -173,6 +229,90 @@ func (p *Post) buildLimitedSupport() (*cmv1.LimitedSupportReason, error) {
...
+
+func (p *Post) readTemplate() *Templatetest {
+	if p.Template == "" {
+		log.Fatalf("Template file is not provided. Use '-t' to fix this.")
+	} else {
+		templateObj, err := p.accessFile(p.Template)
+		if err != nil { //check the presence of this URL or file and also if this can be accessed
+			log.Fatal(err)
+		}
+		fmt.Printf("file = ", string(templateObj))
```
devppratik, line 269: We donot require to print the file

cmd/cluster/support/common.go
```diff
@@ -3,9 +3,14 @@ package support
...
+	"os"
+
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	ctlutil "github.com/openshift/osdctl/pkg/utils"
-	"os"
+)
+
+var (
+	userParameterNames, userParameterValues []string
 )
```
devppratik, line 14: If this is specific to post.go, can this be moved into that file?

devppratik (changes requested):

cmd/cluster/support/scheduled_chaos_test.json
devppratik, line 1: This is additional and can be removed

cmd/cluster/support/common.go
```diff
@@ -3,9 +3,14 @@ package support
...
 	"fmt"
+	"os"
+
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	ctlutil "github.com/openshift/osdctl/pkg/utils"
-	"os"
+)
+
+var (
+	userParameterNames, userParameterValues []string
```
AlexVulaj, line 13: I'd prefer if these variables were scoped to the specific function in which they're needed instead of being package scoped. Package scoped variables can easily bleed state between commands and make issues harder to debug.

cmd/cluster/support/scheduled_chaos_test.json
vibhuti-chalise: removed

cmd/cluster/support/post.go
```diff
@@ -83,8 +117,37 @@ func validateResolutionString(res string) error {
...
+func (p *Post) check() error {
+	if p.Template != "" {
+		if p.Problem != "" || p.Resolution != "" || p.Misconfiguration != "" {
+			return fmt.Errorf("\nIf Template flag is present, all three --problem, --resolution and --misconfiguration flags cannot be used")
+		}
+		p.parseUserParameters() // parse all the '-p' user flags
+
+		// For every '-p' flag, replace its related placeholder in the template
+		for k := range userParameterNames {
+			p.replaceFlags(userParameterNames[k], userParameterValues[k])
```
devppratik, line 129: We need to parse after reading the template
Ref -https://github.com/openshift/osdctl/blob/2bd6c403d30ce1a06c2c24e30a5b54f3098a9184/cmd/servicelog/post.go#L139

cmd/cluster/support/post.go
```diff
@@ -173,6 +243,89 @@ func (p *Post) buildLimitedSupport() (*cmv1.LimitedSupportReason, error) {
 	return limitedSupport, nil
 }
 
+func (p *Post) buildLimitedSupportTemplate() (*cmv1.LimitedSupportReason, error) {
+	t := p.readTemplate() // parse the given JSON template provided via '-t' flag
+
```
devppratik, line 248:
```suggestion
	t := p.readTemplate() // parse the given JSON template provided via '-t' flag
        p.parseUserParameters() // parse all the '-p' user flags

		// For every '-p' flag, replace its related placeholder in the template
		for k := range userParameterNames {
			p.replaceFlags(userParameterNames[k], userParameterValues[k])
```

devppratik: This will ensure that we read the template and parse correctly

cmd/cluster/support/post.go
```diff
@@ -173,6 +243,89 @@ func (p *Post) buildLimitedSupport() (*cmv1.LimitedSupportReason, error) {
...
+	if flagValue == "" {
+		log.Fatalf("The selected template is using '%[1]s' parameter, but '%[1]s' flag was not set. Use '-p %[1]s=\"FOOBAR\"' to fix this.", flagName)
+	}
+
+	found := false
+
+	if !found {
+		log.Fatalf("The selected template is not using '%s' parameter, but '--param' flag was set. Do not use '-p %s=%s' to fix this.", flagName, flagName, flagValue)
+	}
+}
```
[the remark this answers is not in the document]
vibhuti-chalise: implemented this change

fahlmant (approved):
/approve

PR #535, opened 2024-03-15
Sort service logs by timestamp descending

AlexVulaj:
[OSD-21624](https://issues.redhat.com/browse/OSD-21624)

Currently, osdctl servicelog list will output results from oldest to newest. This is great when there are only a few service logs for a cluster, as the most recent messages will be at the bottom of the console output.

Where we run into a problem is when a cluster has a large number of service logs to the point where they don't all fit on one page. In these instances, the newest service logs will not be output at all.

This PR updates the command to return the service logs from OCM ordered by descending timestamp, but still output the results such that the newest service logs are at the bottom of the output.


mjlshen (approved):
/lgtm

PR #539, opened 2024-03-22
Add error message if running against cluster using isolated backplane flow

AlexVulaj:
This PR acts as a temporary bandaid for [OSD-21613](https://issues.redhat.com/browse/OSD-21613).

### DIFF
--- a/cmd/cluster/cpd.go
+++ b/cmd/cluster/cpd.go
@@ -3,6 +3,10 @@ package cluster
 import (
 	"context"
 	"fmt"
+	"github.com/aws/aws-sdk-go-v2/aws/arn"
+	sdk "github.com/openshift-online/ocm-sdk-go"
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/backplane-cli/pkg/ocm"
 
 	awsSdk "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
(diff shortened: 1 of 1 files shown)


cmd/cluster/cpd.go
```diff
@@ -89,32 +94,38 @@ func (o *cpdOptions) run() error {
...
-	awsClient, err := osdCloud.GenerateAWSClientForCluster(o.awsProfile, o.clusterID)
-	if err != nil {
-		fmt.Println("PLEASE CONFIRM YOUR CREDENTIALS ARE CORRECT. If you're absolutely sure they are, send this Service Log https://github.com/openshift/managed-notifications/blob/master/osd/aws/ROSA_AWS_invalid_permissions.json")
-		fmt.Println(err)
-		return err
-	}
+	if isolated, err := isIsolatedBackplaneAccess(cluster, ocmClient); err != nil {
+		return fmt.Errorf("unable to determine which backplane flow this cluster is using: %w.\nNeeds manual investigation:\nocm backplane cloud console -b %s", err, o.clusterID)
+	} else if isolated {
+		return fmt.Errorf("this command doesn't support the isolated backplane flow yet. Needs manual investigation:\nocm backplane cloud console -b %s", o.clusterID)
```
tnierman, line 100: Would it be possible for us to somehow still invoke the old flow in this case? Or does this require the new flow?

AlexVulaj: Invoking the old flow on a cluster using the isolated flow is guaranteed to fail, which could lead us to erroneously telling the SRE to send a service log to the customer.

PR #540, opened 2024-03-27
[OSD-20175] - DQL in OSDCTL

Tafhim:
This PR adds a `logs` sub-command under `cluster` command which help us pull logs from Dynatrace.

### DIFF
--- a/cmd/cluster/dynatrace/dtquery.go
+++ b/cmd/cluster/dynatrace/dtquery.go
@@ -1,0 +1,150 @@
+package dynatrace
+
+import (
+	"fmt"
+	"strings"
+)
+
+type DTQuery struct {
+	fragments  []string
+	finalQuery string
+}
+
+func (q *DTQuery) Init(hours int) *DTQuery {
+	q.fragments = []string{}
+
+	q.fragments = append(q.fragments, fmt.Sprintf("fetch logs, from:now()-%dh \n| filter matchesValue(event.type, \"LOG\")", hours))
+
+	return q
+}
+
+func (q *DTQuery) Cluster(mgmtClusterName string) *DTQuery {
+	q.fragments = append(q.fragments, fmt.Sprintf(" and matchesPhrase(dt.kubernetes.cluster.name, \"%s\")", mgmtClusterName))
+
+	return q
+}
+
+func (q *DTQuery) Namespaces(namespaceList []string, clusterID string, hcp bool) *DTQuery {
+	var nsQuery string
+	finalQuery := ""
+
+	if !hcp {
+		nsQuery = " and ("
+	}
+
+	if hcp && len(namespaceList) > 0 {
+		finalQuery += fmt.Sprintf(" and ( matchesPhrase(k8s.namespace.name, \"ocm-production-%s\")", clusterID)
+		nsQuery = " or "
+	}
+
+	for i, ns := range namespaceList {
+		nsQuery += fmt.Sprintf("matchesValue(k8s.namespace.name, \"%s\")", ns)
+		if i < len(namespaceList)-1 {
+			nsQuery += " or "
+		}
+	}
+	nsQuery += ")"
+	finalQuery += nsQuery
+
+	q.fragments = append(q.fragments, finalQuery)
+
+	return q
+}
+
+func (q *DTQuery) Nodes(nodeList []string) *DTQuery {
+	var nodeQuery string
+
+	nodeQuery = " and ("
+	for i, node := range nodeList {
+		nodeQuery += fmt.Sprintf("matchesValue(k8s.node.name, \"%s\")", node)
+		if i < len(nodeList)-1 {
+			nodeQuery += " or "
+		}
+	}
+	nodeQuery += ")"
+	q.fragments = append(q.fragments, nodeQuery)
+
+	return q
+}
+
+func (q *DTQuery) Pods(podList []string) *DTQuery {
+	var podQuery string
+
+	podQuery = " and ("
+	for i, pod := range podList {
+		podQuery += fmt.Sprintf("matchesValue(k8s.pod.name, \"%s\")", pod)
+		if i < len(podList)-1 {
+			podQuery += " or "
+		}
+	}
+	podQuery += ")"
+	q.fragments = append(q.fragments, podQuery)
+
+	return q
+}
+
+func (q *DTQuery) Containers(containerList []string) *DTQuery {
+	var containerQuery string
+
+	containerQuery = " and ("
+	for i, container := range containerList {
+		containerQuery += fmt.Sprintf("matchesValue(k8s.container.name, \"%s\")", container)
+		if i < len(containerList)-1 {
+			containerQuery += " or "
+		}
+	}
+	containerQuery += ")"
+	q.fragments = append(q.fragments, containerQuery)
+
+	return q
+}
+
+func (q *DTQuery) Status(statusList []string) *DTQuery {
+	var statusQuery string
+
+	statusQuery = " and ("
+	for i, status := range statusList {
+		statusQuery += fmt.Sprintf("matchesValue(status, \"%s\")", status)
+		if i < len(statusList)-1 {
+			statusQuery += " or "
+		}
+	}
+	statusQuery += ")"
+	q.fragments = append(q.fragments, statusQuery)
+
+	return q
+}
+
+func (q *DTQuery) ContainsPhrase(phrase string) *DTQuery {
+	q.fragments = append(q.fragments, " and contains(content,\""+phrase+"\")")
+
+	return q
+}
+
+func (q *DTQuery) Sort(order string) (query *DTQuery, error error) {
+	validOrders := []string{
+		"asc",
+		"desc",
+	}
+
+	for _, or := range validOrders {
+		if or == order {
+			q.fragments = append(q.fragments, fmt.Sprintf("\n| sort timestamp %s", order))
+			return q, nil
+		}
+	}
+
+	return q, fmt.Errorf("no valid sorting order specified. valid order are %s. given %v", strings.Join(validOrders, ", "), order)
+}
+
+func (q *DTQuery) Limit(limit int) *DTQuery {
+	q.fragments = append(q.fragments, "\n| limit "+fmt.Sprint(limit))
+
+	return q
+}
+
+func (q *DTQuery) Build() string {
+	q.finalQuery = strings.Join(q.fragments[:], "")
+
+	return q.finalQuery
+}
--- a/cmd/cluster/dynatrace/rootCmd.go
+++ b/cmd/cluster/dynatrace/rootCmd.go
@@ -1,0 +1,25 @@
+package dynatrace
+
+import (
+	"github.com/spf13/cobra"
+)
+
+const (
+	DynatraceTenantKeyLabel    string = "sre-capabilities.dtp.tenant"
+	HypershiftClusterTypeLabel string = "ext-hypershift.openshift.io/cluster-type"
+)
+
+func NewCmdDynatrace() *cobra.Command {
+	dtCmd := &cobra.Command{
+		Use:               "dynatrace",
+		Aliases:           []string{"dt"},
+		Short:             "Dynatrace related utilities",
+		Args:              cobra.NoArgs,
+		DisableAutoGenTag: true,
+	}
+
+	dtCmd.AddCommand(NewCmdLogs())
+	dtCmd.AddCommand(newCmdURL())
+
+	return dtCmd
+}
--- a/cmd/cluster/dynatrace/urlCmd.go
+++ b/cmd/cluster/dynatrace/urlCmd.go
@@ -1,0 +1,25 @@
+package dynatrace
+
+import (
+	"fmt"
+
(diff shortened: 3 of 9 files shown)


cmd/cluster/logs/cluster.go
```diff
@@ -0,0 +1,111 @@
...
+	sdk "github.com/openshift-online/ocm-sdk-go"
+	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	ocmutils "github.com/openshift/osdctl/pkg/utils"
+	"os"
+)
+
+func fetchDetails(clusterKey string) (string, string, string) {
+	if err := ocmutils.IsValidClusterKey(clusterKey); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
typeid, line 14: **chore (repeating)**: Can we bubble up the error instead of manually calling exit here?

cmd/cluster/logs/cluster.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"fmt"
+	sdk "github.com/openshift-online/ocm-sdk-go"
+	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	ocmutils "github.com/openshift/osdctl/pkg/utils"
+	"os"
+)
+
+func fetchDetails(clusterKey string) (string, string, string) {
```
typeid, line 11: **Suggestion:** name parameters or return a struct, it's hard to tell what this function returns from the signature `string, string, string`.

cmd/cluster/logs/cmd.go
```diff
@@ -0,0 +1,145 @@
...
+func NewCmdLogs() *cobra.Command {
+	logsCmd := &cobra.Command{
+		Use:   "logs",
+		Short: "Fetch logs from Dynatrace",
+		Args:  cobra.NoArgs,
+		RunE:  Run,
+	}
+
+	logsCmd.Flags().BoolVar(&dryRun, "dry-run", false, "Only builds the query without fetching any logs from the tenant")
+	logsCmd.Flags().IntVar(&tail, "tail", 10, "Last 'n' logs to fetch")
```
typeid, line 42: **Question:** is a tail of 10 per default not a bit low? I don't think I've ever looked at only the last 10 log lines.

cmd/cluster/logs/cmd.go
```diff
@@ -0,0 +1,145 @@
...
+	logsCmd.Flags().IntVar(&hours, "hours", 1, "Number of hours back to search")
```
typeid, line 43: **nitpick:** would renaming this to `since` make more since? Then it would match other cmd line tools we use.

cmd/cluster/logs/cmd.go
```diff
@@ -0,0 +1,145 @@
...
+	logsCmd.Flags().StringVar(&contains, "contains", "", "Include logs which contain a phrase")
+	logsCmd.Flags().StringVar(&cluster, "cluster", "", "Cluster name")
```
typeid, line 45: **suggestion:** IMO this should accept cluster internal, external id or name.

cmd/cluster/logs/cmd.go
```diff
@@ -0,0 +1,145 @@
...
+	logsCmd.Flags().StringVar(&sortOrder, "sort", "", "Sort the results by timestamp in either ascending or descending order")
```
typeid, line 46: **nitpick:** It's not clear to me what values this parameter allows and what the default is/does (empty string?). 
I suggest we default to either `asc` or `desc` and the possible values should be `asc` and `desc`.

cmd/cluster/logs/cmd.go
```diff
@@ -0,0 +1,145 @@
...
+	logsCmd.Flags().StringSliceVar(&containerList, "container", []string{}, "Container name(s) (comma-separated)")
+	logsCmd.Flags().StringSliceVar(&statusList, "status", []string{}, "Status(Info/Warn/Error) (comma-separated)")
+
+	return logsCmd
+}
+
+func errorExit(err error) {
+	fmt.Println(err)
+	os.Exit(1)
+}
```
typeid, line 60: **suggestion:** As mentioned above, bubble up errors in a go-like manner. We don't need this function, we should always return errors to the main logic. We're using `RunE` so the error won't get lost:

```
	logsCmd := &cobra.Command{
		Use:   "logs",
		Short: "Fetch logs from Dynatrace",
		Args:  cobra.NoArgs,
		RunE:  Run,
	}
```

cmd/cluster/logs/cmd.go
```diff
@@ -0,0 +1,145 @@
...
+
+func getBase64Url(query string) string {
+	return base64.StdEncoding.EncodeToString([]byte(query))
+}
```
typeid, line 64: **nitpick:** This function doesn't get a base 64 url, it encodes a string to base 64. We could just remove it as calling it has the same length as actually calling what it does internally.

cmd/cluster/logs/cmd.go
```diff
@@ -0,0 +1,145 @@
...
+
+func getLinkToWebConsole(dtUrl string, hours int, base64Url string) string {
+	return fmt.Sprintf("\nLink to Web Console - \n%sui/apps/dynatrace.classic.logs.events/ui/logs-events?gtf=-%dh&gf=all&sortDirection=desc&advancedQueryMode=true&isDefaultQuery=false&visualizationType=table#%s\n\n", dtUrl, hours, base64Url)
+}
+
+func Run(cmd *cobra.Command, args []string) error {
+	if cluster == "" {
+		return fmt.Errorf("Cluster name cannot be left blank")
+	}
```
typeid, line 73: **suggestion:** We can make the parameter required in that case, then we don't have to verify that it isn't blank. 

https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#required-flags

cmd/cluster/logs/cmd.go
```diff
@@ -0,0 +1,145 @@
...
+
+	if hours <= 0 {
+		return fmt.Errorf("Invalid time duration")
+	}
+
+	clusterInternalID, mgmtClusterName, DTURL := fetchDetails(cluster)
```
typeid, line 79: **thought:** I'm certain we already have functions fetching internal id, mgmt cluster, dt url somewhere in osdctl. Maybe worth looking around a bit so we don't duplicate.

cmd/cluster/logs/vault.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		fmt.Printf("Error setting environment variable: %v\n", err)
+		return "", err
+	}
+	loginCommand := exec.Command("vault", "login", "-method=oidc", "-no-print")
+	loginCommand.Stdout = nil
+	loginCommand.Stderr = nil
+	if err := loginCommand.Run(); err != nil {
+		fmt.Println("Error running 'vault login':", err)
+		return "", nil
```
typeid, line 27: **chore (repeating):** Return the error

cmd/cluster/logs/requests.go
```diff
@@ -0,0 +1,194 @@
...
+func getAccessToken() (string, error) {
+	clientSecret, err := getSecretFromVault()
+	if err != nil {
+		return "", err
+	}
+
+	reqData := url.Values{
+		"grant_type":    {"client_credentials"},
+		"scope":         {"storage:logs:read storage:buckets:read"},
+		"client_id":     {"dt0s02.5QMWDNHL"},
```
typeid, line 70: **Question:** is this something we want publicly exposed?

cmd/cluster/logs/requests.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+type LogContent struct {
+	Content string `json:"content"`
+}
+
+type ExecuteResponse struct {
+	RequestToken string `json:"requestToken"`
+}
+
```
typeid, line 126: **Question:** Do these need to be public?

cmd/cluster/logs/vault.go
```diff
@@ -0,0 +1,53 @@
+package logs
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+type VaultResponse struct {
```
typeid, line 10: **Question:** Does this need to be public?

cmd/cluster/logs/dtquery.go
```diff
@@ -0,0 +1,140 @@
+package logs
```
typeid, line 1: **thought:** It's a bit unfortunate dynatrace doesn't provide a go sdk, it would save us a lot of work here.

typeid (changes requested):
Good work @Tafhim! Logic makes sense to me. Added a few nitpicks and change requests on the error handling. 

I'm also +1 to merging this and the existing dynatrace commands into a new dynatrace command family, I'm sure we will notice that we duplicated a few functions once we do that.

cmd/cluster/logs/cmd.go
devppratik: The variable works with  cluster internal, external id or name.
I think the Flag Description can be changed

cmd/cluster/logs/cmd.go
devppratik: +1
@typeid Do you think 100 lines of logs would be good as the intial logs?

typeid: Sounds good to me.

cmd/cluster/logs/requests.go
devppratik: I think this got missed out and was moved directly from DT Logs. We can move this to the osdctl config.

cmd/cluster/dynatrace/dtquery.go
```diff
@@ -0,0 +1,140 @@
...
+	q.fragments = append(q.fragments, " and contains(content,\""+phrase+"\")")
+
+	return q
+}
+
+func (q *DTQuery) Sort(order string) *DTQuery {
+	q.fragments = append(q.fragments, fmt.Sprintf("\n| sort timestamp %s", order))
+
+	return q
+}
```
typeid, line 128: **Suggestion:** check for allowed strings. What if the user enters something that isn't "asc" or "desc"?

cmd/cluster/dynatrace/vault.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		fmt.Printf("Error setting environment variable: %v\n", err)
+		return "", err
+	}
+	vaultCmd := exec.Command("vault", "login", "-method=oidc", "-no-print")
+	vaultCmd.Stdout = nil
+	vaultCmd.Stderr = nil
+	if err = vaultCmd.Run(); err != nil {
+		fmt.Println("Error running 'vault login':", err)
+		return "", nil
```
typeid, line 32: **nitpick:** here and following, there's still errors that are just printed but not returned. Is there a reason we're not returning them to let them be returned in the main logic?

cmd/cluster/logs/vault.go
typeid: @Tafhim this comment is still open.

cmd/cluster/logs/vault.go
typeid: This comment is still open. Is there a reason we're not returning them to let them be returned in the main logic?

cmd/cluster/dynatrace/rootCmd.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+const (
+	DynatraceTenantKeyLabel    string = "sre-capabilities.dtp.tenant"
+	HypershiftClusterTypeLabel string = "ext-hypershift.openshift.io/cluster-type"
+)
+
+func NewCmdDynatrace(streams genericclioptions.IOStreams, client *k8s.LazyClient) *cobra.Command {
```
typeid, line 14: **chore:** if I'm not mistaken, these two parameters are unused, can we remove them?

cmd/cluster/dynatrace/rootCmd.go
```diff
@@ -0,0 +1,26 @@
...
+	dtCmd := &cobra.Command{
+		Use:               "dynatrace",
+		Short:             "Provides information for Dynatrace",
```
typeid, line 17:
```suggestion
		Short:             "Dynatrace related utilities",
```

cmd/cluster/dynatrace/urlCmd.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/spf13/cobra"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+func newCmdURL() *cobra.Command {
+	urlCmd := &cobra.Command{
+		Use:   "url CLUSTER_ID",
+		Short: "Get the Dyntrace Tenant URL for a given MC or HCP cluster",
+		Args:  cobra.ExactArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
```
typeid, line 15: **Suggestion:** use `RunE` here as well. This is not code you touched, but it would still be good to change this if it's minor. :)

typeid (approved):
/lgtm
/approve

PR #548, opened 2024-04-16
OSD-20175: Minor fixes to dynatrace logs commands

devppratik:
- Removes vault path & address from code and uses them from config
- Makes cluster ID as argument not flag


typeid (approved):
/lgtm
/approve

PR #550, opened 2024-04-18
avoid nil exception when awsMachine do not have InstanceIds

Tof1973:
In some cases, awsMachines can do not have any InstanceID (seems when they fail to create), and in that case we should just ship those machine to avoid nil pointer exception.


typeid (approved):
/lgtm
/approve

PR #551, opened 2024-04-18
Improve messaging around cluster uninstallation expectations

mjlshen:
The cluster will only be uninstalled if it already was in an error/uninstalling state. The command does not initiate uninstalls by itself, so this is mostly to clear up any confusion that may arise.

[OHSS-34009](https://issues.redhat.com//browse/OHSS-34009)


typeid (approved):
/lgtm
/approve

PR #553, opened 2024-04-19
terminate only running and stopped instances when cleaning up leaked ec2 

Tof1973:
When runing the cleanup_leaked_ec2 we were not checking the state of the detected leaked instance, so we may detect leaked instances and terminate them, some instance in state terminated or pending (for instance that just have been added - I saw it).

So, in that PR we will just search for instances in running and stopped state. Like that we also exlude stopping and shuting-down states, that we should leave and take...

Also we now get the list of awsMachine after having retrieve the list of Instances to avoid to have an instance created after we have retrieved the list of awsMachines.

ps: replacing #552 which was failing test

### DIFF
--- a/cmd/cluster/cleanup_leaked_ec2.go
+++ b/cmd/cluster/cleanup_leaked_ec2.go
@@ -124,21 +124,6 @@ func (c *cleanup) Run(ctx context.Context) error {
 }
 
 func (c *cleanup) RemediateOCPBUGS23174(ctx context.Context) error {
-	awsmachines := &capav1beta2.AWSMachineList{}
-	if err := c.client.List(ctx, awsmachines, client.MatchingLabels{
-		"cluster.x-k8s.io/cluster-name": c.cluster.ID(),
-	}); err != nil {
-		return err
-	}
-
-	expectedInstances := map[string]bool{}
-	for _, awsmachine := range awsmachines.Items {
-		if awsmachine.Spec.InstanceID != nil {
-			expectedInstances[*awsmachine.Spec.InstanceID] = true
-		}
-	}
-	log.Printf("expected instances: %v", expectedInstances)
-
 	resp, err := c.awsClient.DescribeInstances(ctx, &ec2.DescribeInstancesInput{
 		Filters: []types.Filter{
 			{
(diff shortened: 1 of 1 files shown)


cmd/cluster/cleanup_leaked_ec2.go
```diff
@@ -124,23 +124,14 @@ func (c *cleanup) Run(ctx context.Context) error {
...
-	}
-	log.Printf("expected instances: %v", expectedInstances)
-
 	resp, err := c.awsClient.DescribeInstances(ctx, &ec2.DescribeInstancesInput{
 		Filters: []types.Filter{
+			{
+				// We don't want to terminate pending because they would just be started and might not have been picked up by the MC yet.
+				// Importantly we also don't want to count terminated instances as leaked, as it causes confusion.
+				Name:   aws.String("instance-state-name"),
+				Values: []string{"running", "stopped"},
```
mjlshen, line 133: Optional nit: there are constants available for these states if you would like to use them https://github.com/aws/aws-sdk-go-v2/blob/6dbfda7e9981fc0f5f03e6001526c7b8b5938257/service/ec2/types/enums.go#L2916-L2924

mjlshen (approved):
/approve

cmd/cluster/cleanup_leaked_ec2.go
Tof1973: types.InstanceStateName does not seem to be designed to be used for types.Filter :(
When doing so I have:
cannot use []types.InstanceStateName{…} (value of type []"github.com/aws/aws-sdk-go-v2/service/ec2/types".InstanceStateName) as []string value in struct literal

mjlshen: It would still be `[]string` as it is a type alias for `string`

Tof1973: Well, golang is not as much permissive as this, and it returns an error, because it is expecting an string and not a InstanceStateName :(

mjlshen:
```suggestion
				Values: []string{string(types.InstanceStateNameRunning), string(types.InstanceStateNameStopped)},
```

Tof1973: Ok, indeed, forcing the string type, it works... Not sure it is very "elegant" but if you prefer ;)

typeid:
We're not going to be actively running this, but it doesn't hurt to fix it. :) 

/lgtm

typeid (approved):

PR #555, opened 2024-04-24
[OSD-20552] - Gather Pod logs using DQL through OSDCTL

Tafhim:
This PR adds the functionality to gather logs from HCPs using DQL.
Additionally it also gather Pod YAMLs.

```
osdctl cluster dt hcp-logs roja-s-use1-c3
WARN: The current version (v0.28.0-next) is different than the latest released version (v0.30.0). It is recommended that you update to the latest released version to ensure that no known bugs or issues are hit.
Continue? (y/N): y
Successfully authenticated with DynaTrace
INFO[0020] Backplane URL retrieved via OCM environment: https://api.backplane.openshift.com
INFO[0022] No PagerDuty API Key configuration available. This will result in failure of `ocm-backplane login --pd <incident-id>` command.
INFO[0025] Backplane URL retrieved via OCM environment: https://api.backplane.openshift.com
INFO[0027] No PagerDuty API Key configuration available. This will result in failure of `ocm-backplane login --pd <incident-id>` command.
Using HCP Namespace ocm-production-2aliimms9cvlju07vdt6rere6drogse4-roja-s-use1-c3
[1/83] Pod logs for audit-webhook-54fbb54d88-54gx5
[2/83] Pod logs for audit-webhook-54fbb54d88-884fw
[3/83] Pod logs for aws-ebs-csi-driver-controller-574bbd478d-h9qkw
[4/83] Pod logs for aws-ebs-csi-driver-operator-847fd99977-hnrgr
[5/83] Pod logs for capi-provider-5d487574fc-5dn7n
[6/83] Pod logs for catalog-operator-6fb865b89b-6rcg2
[7/83] Pod logs for certified-operators-catalog-8698dfbc84-vhtgb
[8/83] Pod logs for cloud-network-config-controller-5577866944-zp9j
```


mjlshen (commented):
Overall the intent makes sense, but we should ask the OCP HyperShift team if this is useful for them. One of the main reasons to take a must-gather is when filing an OCPBUGs

cmd/cluster/dynatrace/cluster.go
```diff
@@ -148,3 +153,36 @@ func GetDynatraceURLFromManagementCluster(clusterID string) (string, error) {
...
+	connection, err := utils.CreateConnection()
+	if err != nil {
+		return nil, err
+	}
+	defer connection.Close()
+
+	return connection, nil
+}
+
+func getClientsetFromClusterID(connection *sdk.Connection, clusterID string) (cset *kubernetes.Clientset, error error) {
```
AlexVulaj, line 167: the "connection" parameter isn't used at all here - let's remove it.

cmd/cluster/dynatrace/hcpMgCmd.go
```diff
@@ -0,0 +1,302 @@
...
+		f.Close()
+
+		eventsFilePath := filepath.Join(eventsDirPath, "events.log")
+		f, err = os.OpenFile(eventsFilePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0655)
+		if err != nil {
+			return err
+		}
+
+		err = getEvents(DTURL, accessToken, eventsRequestToken, f)
+		defer f.Close()
```
AlexVulaj, line 137: Not something to hold the PR up for, but it's generally considered risky to put a `defer` in a for-loop because the resources won't be released until the entire loop is completed.

Some workarounds I've seen for this include wrapping the code in an anonymous function, or creating a new named function to call.

cmd/cluster/dynatrace/requests.go
```diff
@@ -186,35 +205,93 @@ func getRequestToken(query string, dtURL string, accessToken string) (requestTok
...
+
+		if dtPollRes.State == "RUNNING" {
+			continue
+		}
+
+		if dtPollRes.State == "SUCCEEDED" {
+			return resp, nil
+		}
+
+		if dtPollRes.State != "RUNNING" && dtPollRes.State == "SUCCEEDED" {
```
AlexVulaj, line 247: This if-statement can be simplified to just the second clause
```suggestion
		if dtPollRes.State == "SUCCEEDED" {
```

cmd/cluster/dynatrace/cluster.go
```diff
@@ -148,3 +153,36 @@ func GetDynatraceURLFromManagementCluster(clusterID string) (string, error) {
...
+
+func getConnection() (conn *sdk.Connection, error error) {
+	connection, err := utils.CreateConnection()
+	if err != nil {
+		return nil, err
+	}
+	defer connection.Close()
+
+	return connection, nil
+}
```
typeid, line 165: **Question:** Does this work? It looks wrong to me: if I'm not mistaken, we would be closing the connection at the end of the `getConnection()` before returning it.

cmd/cluster/dynatrace/cluster.go
```diff
@@ -148,3 +153,36 @@ func GetDynatraceURLFromManagementCluster(clusterID string) (string, error) {
...
+
+func getClientsetFromClusterID(clusterID string) (cset *kubernetes.Clientset, error error) {
+	_, _, clientset, err := common.GetKubeConfigAndClient(clusterID, "", "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve Kubernetes configuration and client for cluster with ID %s: %w", clusterID, err)
+	}
+
+	return clientset, nil
+}
```
typeid, line 174: **nitpick:** this seems like a bit of a bloating wrapper.  It's just a function calling `common.GetKubeConfigAndClient` without any further logic. 

Why not directly have the following in the main logic?
```
	_, _, clientset, err := common.GetKubeConfigAndClient(clusterID, "", "")
	if err != nil {
		return nil, fmt.Errorf("failed to retrieve Kubernetes configuration and client for cluster with ID %s: %w", clusterID, err)
	}
```

cmd/cluster/dynatrace/hcpMgCmd.go
```diff
@@ -0,0 +1,300 @@
...
+		Run: func(cmd *cobra.Command, args []string) {
+			err := mustGather(args[0])
+			if err != nil {
+				cmdutil.CheckErr(err)
+			}
+		},
+	}
+
+	hcpMgCmd.Flags().IntVar(&since, "since", 10, "Number of hours (integer) since which to pull logs and events")
+	hcpMgCmd.Flags().IntVar(&tail, "tail", 100, "Last 'n' logs and events to fetch ")
```
typeid, line 34: **issue:** I don't think we need the tail flag if we have a `since` flag. I would expect all logs in the time interval to be gathered when there's a `since` flag. This would cause confusion in my opinion.

**issue:** 100 lines seems way too low as default. A standard must-gather will get all the available logs, can we do this here as well?

cmd/cluster/dynatrace/hcpMgCmd.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	hcpNS, err := GetHCPNamespaceFromInternalID(clientset, clusterInternalID)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(fmt.Sprintf("Using HCP Namespace %v", hcpNS))
+
+	gatherNamespaces := []string{hcpNS, "hypershift", "cert-manager", "redhat-cert-manager-operator"}
```
typeid, line 67: **issue:** We're forgetting the nodepool namespace here.

typeid (changes requested):
Good stuff @Tafhim, this is a great starting point to allow us properly opening OCPBUGS with disabled debug handlers. 

I'm wondering about the scope of the must-gather command. The way I understood it on slack it is meant to only gather logs, now it also gathers some yamls and events. What do you suggest the process to open an OCPBUG looks like? 
We would still need quite a few other yamls that are not included here, so as it stands we would still need to run a classic hypershift must-gather just to get all the necessary yamls on top of this, leading to duplicated yamls.

cmd/cluster/dynatrace/hcpMgCmd.go
typeid, line 67: We're probably also forgetting the kubelet namespace, can we check what else a hypershift dump gathers?

typeid:
One more thing pending to be clarified: does this gather all container's logs, or just the default container's logs?

typeid (approved):
/lgtm
/approve

Thanks @Tafhim - this is good to go IMO, we can always improve the format and add more things in the future.

PR #560, opened 2024-05-05
[SDE-3246] Cloudtail write-events feature

TomNelsonTembo:
This pull request introduces a new command, osdctl cloudtrail write-events, enhancing the capabilities of osdctl to effectively query CloudTrail logs for write events related to a specific cluster.

### DIFF
--- a/cmd/cloudtrail/write-events.go
+++ b/cmd/cloudtrail/write-events.go
@@ -200,7 +200,7 @@ func filterUsers(lookupOutputs []*cloudtrail.LookupEventsOutput, Ignore []string
 
 				}
 
-				if matchesUsername || matchesArn {
+				if matchesUsername || matchesArn || userArn == "" && event.Username == nil {
 					continue
 					// skips entry
 				}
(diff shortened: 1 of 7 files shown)


cmd/cloudtrail/write-events.go
```diff
@@ -0,0 +1,341 @@
...
+
+		printEvents(*filteredEvents, o.printEventUrl, o.printRawEvents)
+		fmt.Println("")
+		return err
+	}
+
+	arn, accountId, err := whoami(*sts.NewFromConfig(cfg))
+	fmt.Printf("[INFO] Checking write event history since %v for AWS Account %v as %v \n", startTime, accountId, arn)
+	cloudTrailclient := cloudtrail.NewFromConfig(cfg)
+	defaultCloudtrailClient := cloudtrail.New(cloudtrail.Options{
```
typeid, line 323: Let's keep this variable scoped to the minimum necessary. We only need it in case:  `if defaultConfig.Region != cfg.Region {`, so we can move it to that block.

cmd/cloudtrail/write-events.go
```diff
@@ -0,0 +1,341 @@
...
+		return err
+	}
+
+	startTime, err := parseDurationToUTC(o.startTime)
+	if err != nil {
+		return err
+	}
+
+	fetchFilterPrintEvents := func(client cloudtrail.Client, startTime time.Time, o *LookupEventsOptions) error {
+
```
typeid, line 305:
```suggestion
```

cmd/cloudtrail/write-events.go
```diff
@@ -0,0 +1,341 @@
...
+					fmt.Print("\n")
+				}
+
+				if printUrl && event.CloudTrailEvent != nil {
+					if err == nil {
+						fmt.Printf("EventLink: %v\n\n", generateLink(*rawEventDetails))
+					} else {
+						fmt.Println("EventLink: <not available>")
+					}
+				}
```
typeid, line 252: Let's use a stringbuilder kind of pattern here. See `strings.Builder`.

cmd/cloudtrail/write-events.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+// PrintEvents prints the details of each event in the provided slice of events.
+// It takes a slice of types.Event
+func printEvents(filteredEvent []types.Event, printUrl bool, raw bool) {
+	for _, event := range filteredEvent {
+		if raw {
+			if event.CloudTrailEvent != nil {
+				fmt.Printf("%v \n\n", *event.CloudTrailEvent)
+			}
```
typeid, line 222: We can return early, then we don't need the extra else.

cmd/cloudtrail/write-events.go
```diff
@@ -0,0 +1,341 @@
...
+	cfg, err := osdCloud.CreateAWSV2Config(connection, cluster)
+	if err != nil {
+		return err
+	}
+	DefaultRegion := "us-east-1"
+	configOptions := config.WithRegion(DefaultRegion)
+	defaultConfig, err := config.LoadDefaultConfig(context.Background(), configOptions)
+	if err != nil {
+		return err
+	}
```
typeid, line 297: We should only touch the `us-east-1` part once we know the cluster's region is different, as commented below, move this to the minimal scope.

cmd/cloudtrail/write-events.go
```diff
@@ -0,0 +1,341 @@
...
+				}
+
+			}
+		}
+	}
+
+}
+
+func (o *LookupEventsOptions) run() error {
+
```
typeid, line 261:
```suggestion
```

typeid (changes requested):

cmd/cloudtrail/write-events.go
```diff
@@ -0,0 +1,340 @@
...
+	if strings.ToUpper(cluster.CloudProvider().ID()) != "AWS" {
+		return fmt.Errorf("[ERROR] this command is only available for AWS clusters")
+	}
+
+	Ignore, err := envConfig.LoadCloudTrailConfig()
+	if err != nil {
+		return fmt.Errorf("[ERROR] error Loading cloudtrail configuration file: %w", err)
+	}
+	if len(Ignore) == 0 {
+		fmt.Println("\n[WARNING] No filter list DETECTED!!")
```
typeid, line 284:
```suggestion
		fmt.Println("\n[WARNING] No filter list detected! If you want intend to apply user filtering for the cloudtrail events, please add cloudtrail_cmd_lists to your osdctl configuration file.")
```

PR #562, opened 2024-05-09
Add admin-username to rotate-secret command

AlexVulaj:
Looking at [OHSS-34220](https://issues.redhat.com//browse/OHSS-34220) I found an issue where osdctl would fail to rotate secrets if the customer had an `osdManagedAdmin*` user that didn't match our expected pattern due to having multiple clusters in their AWS account. This new flag will allow SREs to manually override osdctl and provide an admin username in this instance as long as it still starts with `osdManagedAdmin`.

Corresponding SOP update: https://github.com/openshift/ops-sop/pull/3155

### DIFF
--- a/cmd/account/rotate-secret.go
+++ b/cmd/account/rotate-secret.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	awsSdk "github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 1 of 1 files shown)


cmd/account/rotate-secret.go
```diff
@@ -208,8 +215,12 @@ func (o *rotateSecretOptions) run() error {
...
 
 	// Update osdManagedAdmin secrets
-	// Username is osdManagedAdmin-aaabbb
-	osdManagedAdminUsername := common.OSDManagedAdminIAM + "-" + accountIDSuffixLabel
+	var osdManagedAdminUsername string
+	if o.osdManagedAdminUsername != "" {
+		osdManagedAdminUsername = o.osdManagedAdminUsername
+	} else {
+		osdManagedAdminUsername = common.OSDManagedAdminIAM + "-" + accountIDSuffixLabel
+	}
```
NautiluX, line 223: I feel like this could be simplified and get rid of the else... something like the following probably?
```suggestion
	osdManagedAdminUsername := o.osdManagedAdminUsername
	if osdManagedAdminUsername == "" {
		osdManagedAdminUsername = common.OSDManagedAdminIAM + "-" + accountIDSuffixLabel
	}
```

NautiluX (commented):
/lgtm

PR #571, opened 2024-05-21
Remove completion command

AlexVulaj:
Per [Cobra's documentation](https://github.com/spf13/cobra/blob/main/site/content/completions/_index.md), this functionality is automatically provided and we don't need our own custom command for it:

> Cobra will automatically provide your program with a fully functional completion command, similarly to how it provides the help command.


mjlshen (approved):
/lgtm

PR #578, opened 2024-05-31
OSD-23744: make context cmd quicker

Tof1973:
- Some actions which were done in sequence are now done in parallel
- We are now retrieving the cluster from OCM only one time in the process while it was done twice
- before running the OCM query to retrieve cluster, we now detect first what is the cluster identifier, in order to make the query quicker
- Add verbose output with delay for each action in order to be able to troubleshoot easily when command is taking time

With those changes we move from a frequent context of more than 10s (and sometime much more) to a context of less than 5s most of the time ;)


pkg/utils/ocm.go
```diff
@@ -143,7 +144,14 @@ func ApplyFilters(ocmClient *sdk.Connection, filters []string) ([]*cmv1.Cluster,
 
 // GenerateQuery returns an OCM search query to retrieve all clusters matching an expression (ie- "foo%")
 func GenerateQuery(clusterIdentifier string) string {
```
AlexVulaj, line 146: Great work with these regexes - do you mind writing a couple of simple unit test cases for this function?

Tof1973: test added

PR #579, opened 2024-06-03
Add warning to network verification that HCPs are unsupported.

RaphaelBut:
Its not widely known, that network verifier is unable to run against HCPs.
This can lead to red herring situations and a warning is needed.

This is a follow up item from PMR 75 https://issues.redhat.com/browse/OSD-23785

### DIFF
--- a/cmd/network/verification.go
+++ b/cmd/network/verification.go
@@ -3,6 +3,7 @@ package network
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"os"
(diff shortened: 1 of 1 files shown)


cmd/network/verification.go
```diff
@@ -218,6 +219,12 @@ func (e *EgressVerification) setup(ctx context.Context) (*aws.Config, error) {
 		e.log.Debug(ctx, "cluster %s found from OCM: %s", e.ClusterId, cluster.ID())
 		e.cluster = cluster
 
+		if e.cluster.Hypershift().Enabled() {
+			return nil, errors.New("failed to run network verifier." +
+				" Hosted control plane clusters are not supported." +
+				" Instead, please use the network verifier script in ops-sop/hypershift/utils/verify-egress.sh")
+		}
```
mjlshen, line 226: If you agree, I think it would be better to print a warning message that SRE has to confirm to bypass. Something like "Generaly, SRE has insufficient AWS permissions to run network verifier on hosted control plane clusters, run anyway?"

cmd/network/verification.go
```diff
@@ -218,6 +219,12 @@ func (e *EgressVerification) setup(ctx context.Context) (*aws.Config, error) {
 		e.log.Debug(ctx, "cluster %s found from OCM: %s", e.ClusterId, cluster.ID())
 		e.cluster = cluster
 
+		if e.cluster.Hypershift().Enabled() {
```
AlexVulaj, line 222: Small conversation about this here: https://redhat-internal.slack.com/archives/CMD1L32M9/p1717424955110269

We may want to additionally

1. Take that bash script and just make it a new osdctl command (for another PR, not here 😄 )
2. Leave comments here explaining _why_ HCPs aren't supported, and when we can remove this block of code

RaphaelBut: Good suggestions thanks! Will add the info from the thread.

PR #580, opened 2024-06-05
OSD-22394 osdctl alert silence --all does not silence

itsmitul9:
- Fixed issue for making all alerts silenced using --all 
- Refactored code 
- Silence list display namespace in AlertName
- Using namespace matcher to alert all silence


cmd/alerts/silence/add_silence.go
```diff
@@ -65,27 +67,71 @@ func AddSilence(cmd *addSilenceCmd) {
...
+			return
+		}
 
+		formattedOutput := strings.Replace(output, "\n", " ", -1)
+
+		fmt.Printf("Alert %s has been silenced with id \"%s\" for a duration of %s by user \"%s\" \n", alert.Labels.Alertname, formattedOutput, duration, username)
+	}
+}
+
+func fetchAllAlerts(clusterID string, kubeconfig *rest.Config, clientset *kubernetes.Clientset) []utils.Alert {
```
AlexVulaj, line 95: is this clusterID param used anywhere? If not let's get rid of it.

cmd/alerts/silence/add_silence.go
```diff
@@ -65,27 +67,71 @@ func AddSilence(cmd *addSilenceCmd) {
...
+
+	err = json.Unmarshal([]byte(output), &fetchedAlerts)
+	if err != nil {
+		log.Fatal("Error in unmarshaling the alerts", err)
+	}
+
+	return fetchedAlerts
+}
+
+/*
```
AlexVulaj, line 112: If we don't need to keep this commented out section around, I'd say delete it entirely to remove confusion.

cmd/alerts/utils/utils.go
```diff
@@ -15,14 +16,35 @@ const (
...
+	var err error
 
-	req := clientset.CoreV1().RESTClient().Post().Resource("pods").Name(PodName).
-		Namespace(AccountNamespace).SubResource("exec")
+	// Attempt to execute with the primary pod
+	cmdOutput, err = ExecWithPod(kubeconfig, clientset, PrimaryPod, cmd)
+	if err == nil {
+		return cmdOutput, nil // Successfully executed
+	}
+
```
AlexVulaj, line 33: We should log the above error here before informing the user that we're attempting to fallback to the secondary pod.

AlexVulaj:
Some thoughts as I look over this PR:

- `ExecWithPod` should take a "namespace" as a parameter and be moved to a more generic "utils" package, as it's a useful command all around and doesn't have any logic specific to alerts.
-  `ExecInPod` should have a more specific name, maybe something like "ExecInAlertManagerPod", and a more descriptive comment to go along with that. Right now it's unclear to someone looking at the function name and comments that it will try on both AlertManager pods specifically.

cmd/alerts/utils/utils.go
```diff
@@ -15,14 +16,35 @@ const (
...
+	// If execution with primary pod fails, try with the secondary pod
+	cmdOutput, err = ExecWithPod(kubeconfig, clientset, SecondaryPod, cmd)
+	if err == nil {
+		return cmdOutput, nil // Successfully executed
+	}
 
+	// If execution with both pods fails, print error message
+	fmt.Println("Exec Failed. Please put silence manually")
```
AlexVulaj, line 41: I'd remove the println from this util function and leave output to calling functions. Returning the error should be enough for callers to craft their own output to the user.

itsmitul9:
```
fmt.Println("Execution Failed with alertmanager-main-0 and alertmanager-main-1. Please put silence manually")
	return "", er
```
Are we looking something like ^

cmd/alerts/utils/utils.go
itsmitul9:
```
cmdOutput, err = ExecWithPod(kubeconfig, clientset, PrimaryPod, cmd)
	if err == nil {
		return cmdOutput, nil
	}

	fmt.Printf("Execution with alertmanager-main-0 failed: %v\n", err)

	fmt.Println("Attempting with alertmanger-main-1")
	cmdOutput, err = ExecWithPod(kubeconfig, clientset, SecondaryPod, cmd)
	if err == nil {
		return cmdOutput, nil
	}

	fmt.Printf("Execution with alertmanager-main-1 failed: %v\n", err)
```

itsmitul9: Current execution returns 
```

Execution with alertmanager-main-0 failed: failed to stream with context: unable to upgrade connection: container not found ("alertmanager")
Attempting with alertmanger-main-1
Alert Information:
  AlertName:  AlertmanagerSilencesActiveSRE
  Severity:   warning
  State:      active
  Message:
```

cmd/alerts/utils/utils.go
AlexVulaj: I see now the error is printed out in `ExecWithPod`. I don't love this, as general good practice is to bubble errors up to the caller and let them handle it as they see fit (i.e. functions that perform business logic should only return an error, not print it).

Alas, I don't think that change should be the responsibility of this PR.

AlexVulaj: Negative - I'd remove the Println on line 41 entirely. Let the calling functions handle output to the user.

PR #585, opened 2024-06-18
OSD-22635: use elevation for oc execs

typeid:
As part of 
https://issues.redhat.com/browse/OSD-22635 `oc exec` and `oc debug` [have been gated to backplane cluster admin](https://github.com/openshift/managed-cluster-config/pull/2149). We now need to elevate of `oc exec`. 

SOPs have already been updated for the changes ([see here](https://github.com/openshift/ops-sop/pull/3200)). 

https://issues.redhat.com/browse/OSD-22635

### DIFF
--- a/cmd/alerts/silence/list_silence.go
+++ b/cmd/alerts/silence/list_silence.go
@@ -33,25 +33,42 @@ type Silence struct {
 	StartsAt  string `json:"startsAt"`
 }
 
+type listSilenceCmd struct {
+	clusterID string
+	reason    string
+}
+
 func NewCmdListSilence() *cobra.Command {
-	return &cobra.Command{
+	listSilenceCmd := &listSilenceCmd{}
+	cmd := &cobra.Command{
 		Use:               "list <cluster-id>",
 		Short:             "List all silences",
 		Long:              `print the list of silences`,
 		Args:              cobra.ExactArgs(1),
 		DisableAutoGenTag: true,
 		Run: func(cmd *cobra.Command, args []string) {
-			ListSilence(args[0])
+			listSilenceCmd.clusterID = args[0]
+			ListSilence(listSilenceCmd)
 		},
 	}
+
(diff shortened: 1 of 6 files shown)


cmd/cluster/detachstuckvolume.go
```diff
@@ -66,13 +71,17 @@ func (o *detachStuckVolumeOptions) detachVolume(clusterID string) error {
...
+		o.reason,
+		"Detach stuck volume in openshift-monitoring",
+	}
+	_, _, clientset, err := common.GetKubeConfigAndClient(o.clusterID, elevationReasons...)
 
 	if err != nil {
 		return fmt.Errorf("failed to retrieve Kubernetes configuration and client for cluster with ID %s: %w", o.clusterID, err)
 	}
 
-	err = getVolumeID(clientset, Namespace, "")
```
typeid, line 75: This was unused

cmd/alerts/silence/list_silence.go
```diff
@@ -33,25 +33,42 @@ type Silence struct {
...
 	}
+
+	cmd.Flags().StringVar(&listSilenceCmd.reason, "reason", "", "The reason for this command, which requires elevation, to be run (usualy an OHSS or PD ticket)")
+	_ = cmd.MarkFlagRequired("reason")
+
+	return cmd
 }
 
-func ListSilence(clusterID string) {
+func ListSilence(cmd *listSilenceCmd) {
```
AlexVulaj, line 61: What do you think of just having that function take two parameters - a "reason" and a "clusterId"? This decouples it entirely from the cobra library and makes it a bit easier to test.

typeid: Hmm, this is actually already decoupled. `listSilenceCmd` is just a struct and has nothing to do with cobra, maybe the name is causing confusion? But this is how we do it for the other commands as well (before this change).

PR #591, opened 2024-07-02
OSD-24351: permission denied command should also evaluate read-only events

typeid:
**What?**
Fixes the `osdctl cloudtrail permission-denied-events` command to also evaluate read-only events
Changes the default poll time to 5 minutes, as we are polling significantly more events and are not interested in too much history for this command.

**Why?**
 The idea is that we see any permission denied events, even ready only events being denied is interesting to us. The current functionality only shows write events for which permission was denied.


mjlshen (approved):
/lgtm

PR #592, opened 2024-07-09
Don't use deprecated 'securityGroupId' in verifier library

AlexVulaj:
In the network verifier library, `securityGroupId` has long been deprecated in favor of `securityGroupIds`. This PR moves us over to use the latter so we aren't using a deprecated field.


AlexVulaj:
I see some tests broke here - I'll take a look at those ASAP.

mjlshen (approved):
/lgtm
/hold

LGTM, but I understand @abyrne55 would like to test this locally first

abyrne55:
Thanks for the patience, all. Tested locally and confirmed new output formatting and other expected behaviors. 

/lgtm
/unhold

PR #594, opened 2024-07-10
Prompt when putting critical customers into LS

AlexVulaj:
When putting a cluster into LS that is owned by a critical customer, the following prompt will show first:
```
WARNING: This cluster is owned by a critical customer. Make sure that an SL has been sent and proactive case opened with the customer. Only continue if there has been no customer response for 24 hours.

See: https://source.redhat.com/groups/public/sre/wiki/defining_limited_support_process_for_osdrosa_for_critical_customers
Continue? (y/N):
```

Details in [OSD-23920](https://issues.redhat.com//browse/OSD-23920)

### DIFF
--- a/cmd/cluster/support/post.go
+++ b/cmd/cluster/support/post.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -33,6 +34,7 @@ const (
 	InternalServiceLogSeverity                           = "Warning"
 	InternalServiceLogServiceName                        = "SREManualAction"
 	InternalServiceLogSummary                            = "LimitedSupportEvidence"
+	managedCriticalCustomerLabel                         = "capability.organization.managed_critical_customer"
 )
 
 type Post struct {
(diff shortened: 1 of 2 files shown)


cmd/cluster/support/post.go
```diff
@@ -163,6 +165,37 @@ func (p *Post) Run(clusterID string) error {
...
+		Labels().
+		Label(managedCriticalCustomerLabel).
+		Get().Send()
+	if err != nil {
+		// if the label is missing, there's no need to show an error to the user
+		if labelResponse.Error().Status() != http.StatusNotFound {
+			return fmt.Errorf("failed to retrieve cluster labels: %w", err)
+		}
+	} else if labelResponse.Body().Value() == "true" {
+		fmt.Println("This cluster is owned by a critical customer.")
```
typeid, line 193: **Question:** Should we be more specific here in terms of what this means for the user?

typeid (approved):
LGTM, left one question but I'll leave it up to you if you want to address it or not (feel free to unhold).

cmd/cluster/support/post.go
AlexVulaj: Great suggestion. What about something like this?
```
WARNING: This cluster is owned by a critical customer. Make sure that an SL has been sent and proactive case opened with the customer. Only continue if there has been no customer response for 24 hours.

See: https://source.redhat.com/groups/public/sre/wiki/defining_limited_support_process_for_osdrosa_for_critical_customers

Continue? (y/N)
```

PR #596, opened 2024-07-11
Update filter logic for finding subnets and security groups due to 4.16 changes

AlexVulaj:
Resolves [OSD-24457](https://issues.redhat.com/browse/OSD-24457).

As of 4.16, some downstream component is causing the subnet tag `kubernetes.io/cluster/<infra_id>` to show as "shared" instead of "owned" now, even for fresh installs in clean AWS accounts. As far as the network verifier is concerned, we probably don't actually care about the value of this tag, just that it exists.

There was also a regression captured in [OCPBUGS-36902](https://issues.redhat.com/browse/OCPBUGS-36902) with the `kubernetes.io/cluster/<infra_id>` tag being removed from security groups. Since the `Name` tag already includes a unique cluster identifier, I've dropped this tag entirely from the filters to prevent issues with searching.

I verified on a cluster installed at 4.1.z that the Name tag was present with the value that we expect, so I wouldn't expect any issues with this change on older clusters.

### DIFF
--- a/cmd/network/verification.go
+++ b/cmd/network/verification.go
@@ -415,12 +415,12 @@ func (e *EgressVerification) getSubnetIds(ctx context.Context) ([]string, error)
 
 	// If this is a non-BYOVPC cluster, we can find the private subnets based on the cluster and internal-elb tag
 	if len(e.cluster.AWS().SubnetIDs()) == 0 {
-		e.log.Info(ctx, "searching for subnets by tags: kubernetes.io/cluster/%s=owned and %s=", e.cluster.InfraID(), nonByovpcPrivateSubnetTagKey)
+		e.log.Info(ctx, "searching for subnets by tags: kubernetes.io/cluster/%s and %s=", e.cluster.InfraID(), nonByovpcPrivateSubnetTagKey)
 		resp, err := e.awsClient.DescribeSubnets(ctx, &ec2.DescribeSubnetsInput{
 			Filters: []types.Filter{
 				{
-					Name:   aws.String(fmt.Sprintf("tag:kubernetes.io/cluster/%s", e.cluster.InfraID())),
-					Values: []string{"owned"},
+					Name:   aws.String("tag-key"),
+					Values: []string{fmt.Sprintf("kubernetes.io/cluster/%s", e.cluster.InfraID())},
 				},
 				{
 					Name:   aws.String("tag-key"),
@@ -433,7 +433,7 @@ func (e *EgressVerification) getSubnetIds(ctx context.Context) ([]string, error)
 		}
 
 		if len(resp.Subnets) == 0 {
-			return nil, fmt.Errorf("found 0 subnets with kubernetes.io/cluster/%s=owned and %s, consider the --subnet-id flag", e.cluster.InfraID(), e.cluster.InfraID())
+			return nil, fmt.Errorf("found 0 subnets with tags: kubernetes.io/cluster/%s and %s, consider the --subnet-id flag", e.cluster.InfraID(), nonByovpcPrivateSubnetTagKey)
 		}
 		if e.AllSubnets {
 			subnets := make([]string, len(resp.Subnets))
@@ -553,20 +553,14 @@ func (e *EgressVerification) getSecurityGroupId(ctx context.Context) (string, er
 				Name:   aws.String("tag:Name"),
 				Values: []string{fmt.Sprintf("%s-default-sg", e.cluster.ID())},
 			},
-			{
-				Name:   aws.String(fmt.Sprintf("tag:kubernetes.io/cluster/%s", e.cluster.ID())),
-				Values: []string{"owned"},
-			},
 		}
 	default:
 		filters = []types.Filter{
 			{
+				// Prior to 4.16: <infra_id>-master-sg
+				// 4.16+: <infra_id>-controlplane
 				Name:   aws.String("tag:Name"),
-				Values: []string{fmt.Sprintf("%s-master-sg", e.cluster.InfraID())},
-			},
-			{
-				Name:   aws.String(fmt.Sprintf("tag:kubernetes.io/cluster/%s", e.cluster.InfraID())),
-				Values: []string{"owned"},
+				Values: []string{fmt.Sprintf("%s-master-sg", e.cluster.InfraID()), fmt.Sprintf("%s-controlplane", e.cluster.InfraID())},
 			},
 		}
 	}


cmd/network/verification.go
```diff
@@ -415,12 +415,12 @@ func (e *EgressVerification) getSubnetIds(ctx context.Context) ([]string, error)
...
 	// If this is a non-BYOVPC cluster, we can find the private subnets based on the cluster and internal-elb tag
 	if len(e.cluster.AWS().SubnetIDs()) == 0 {
-		e.log.Info(ctx, "searching for subnets by tags: kubernetes.io/cluster/%s=owned and %s=", e.cluster.InfraID(), nonByovpcPrivateSubnetTagKey)
+		e.log.Info(ctx, "searching for subnets by tags: kubernetes.io/cluster/%s=[owned/shared] and %s=", e.cluster.InfraID(), nonByovpcPrivateSubnetTagKey)
 		resp, err := e.awsClient.DescribeSubnets(ctx, &ec2.DescribeSubnetsInput{
 			Filters: []types.Filter{
 				{
 					Name:   aws.String(fmt.Sprintf("tag:kubernetes.io/cluster/%s", e.cluster.InfraID())),
-					Values: []string{"owned"},
+					Values: []string{"owned", "shared"},
```
AlexVulaj, line 423: Thought - how do we feel about blowing away the "Value" here entirely? I don't think we particularly care what the value is, just that it exists.

AlexVulaj: After discussion with @fahlmant I've gone ahead and removed the value check here.

cmd/network/verification.go
```diff
@@ -553,20 +552,12 @@ func (e *EgressVerification) getSecurityGroupId(ctx context.Context) (string, er
...
 	default:
 		filters = []types.Filter{
 			{
 				Name:   aws.String("tag:Name"),
-				Values: []string{fmt.Sprintf("%s-master-sg", e.cluster.InfraID())},
-			},
-			{
-				Name:   aws.String(fmt.Sprintf("tag:kubernetes.io/cluster/%s", e.cluster.InfraID())),
-				Values: []string{"owned"},
+				Values: []string{fmt.Sprintf("%s-master-sg", e.cluster.InfraID()), fmt.Sprintf("%s-controlplane", e.cluster.InfraID())},
```
AlexVulaj, line 560: in 4.16m `<infra_id>-master-sg` => `<infra_id>-controlplane`. We need to check for both now.

AlexVulaj:
/hold
Still some discussion going on.

cmd/network/verification.go
```diff
@@ -415,12 +415,12 @@ func (e *EgressVerification) getSubnetIds(ctx context.Context) ([]string, error)
...
 	if len(e.cluster.AWS().SubnetIDs()) == 0 {
-		e.log.Info(ctx, "searching for subnets by tags: kubernetes.io/cluster/%s=owned and %s=", e.cluster.InfraID(), nonByovpcPrivateSubnetTagKey)
+		e.log.Info(ctx, "searching for subnets by tags: kubernetes.io/cluster/%s and %s=", e.cluster.InfraID(), nonByovpcPrivateSubnetTagKey)
 		resp, err := e.awsClient.DescribeSubnets(ctx, &ec2.DescribeSubnetsInput{
 			Filters: []types.Filter{
 				{
-					Name:   aws.String(fmt.Sprintf("tag:kubernetes.io/cluster/%s", e.cluster.InfraID())),
-					Values: []string{"owned"},
+					Name:   aws.String("tag-key"),
+					Values: []string{fmt.Sprintf("kubernetes.io/cluster/%s", e.cluster.InfraID())},
```
AlexVulaj, line 423: I'm _pretty_ sure that all of the values in a `Values` block here get `or`'d together, so I'm keeping these two blocks separate since we want both.

cmd/network/verification.go
```diff
@@ -553,20 +553,12 @@ func (e *EgressVerification) getSecurityGroupId(ctx context.Context) (string, er
...
 	default:
 		filters = []types.Filter{
 			{
 				Name:   aws.String("tag:Name"),
-				Values: []string{fmt.Sprintf("%s-master-sg", e.cluster.InfraID())},
-			},
-			{
-				Name:   aws.String(fmt.Sprintf("tag:kubernetes.io/cluster/%s", e.cluster.InfraID())),
-				Values: []string{"owned"},
+				Values: []string{fmt.Sprintf("%s-master-sg", e.cluster.InfraID()), fmt.Sprintf("%s-controlplane", e.cluster.InfraID())},
```
clcollins, line 561: Add a comment so we can deprecate code later as clusters age up?  We won't have the mental context for this in years++, and will end up carrying this forward without knowing why.

```suggestion
				// Prior to 4.16: <infra_id>-master-sg
				// 4.16+: <infra_id>-controlplane
				Values: []string{fmt.Sprintf("%s-master-sg", e.cluster.InfraID()), fmt.Sprintf("%s-controlplane", e.cluster.InfraID())},
```

cmd/network/verification.go
clcollins: I think you're right.  The CLI works that way too, and from the type definition for `Filter`: 

```
	// The filter values. Filter values are case-sensitive. If you specify multiple
	// values for a filter, the values are joined with an OR , and the request returns
	// all results that match any of the specified values.
	
```

So, confirmed, I guess :)

clcollins (changes requested):
Minor suggestion for some comments, but otherwise LGTM.  Feel free to either add, or not, the comments, and I'll LGTM officially either way.

cmd/network/verification.go
AlexVulaj: +1 to this, great suggestion

AlexVulaj: It may be hard to ever fully deprecate this, as clusters could technically be installed on a version way before 4.16 and then update beyond. In that case, the security group tags would still stick with their installation values.

_That said_, I still like this comment anyway 🙂

cmd/network/verification.go
```diff
@@ -433,7 +433,7 @@ func (e *EgressVerification) getSubnetIds(ctx context.Context) ([]string, error)
 		}
 
 		if len(resp.Subnets) == 0 {
-			return nil, fmt.Errorf("found 0 subnets with kubernetes.io/cluster/%s=owned and %s, consider the --subnet-id flag", e.cluster.InfraID(), e.cluster.InfraID())
+			return nil, fmt.Errorf("found 0 subnets with kubernetes.io/cluster/%s and %s, consider the --subnet-id flag", e.cluster.InfraID(), e.cluster.InfraID())
```
mjlshen, line 436:
```suggestion
			return nil, fmt.Errorf("found 0 subnets with tags: kubernetes.io/cluster/%s and %s, consider the --subnet-id flag", e.cluster.InfraID(), nonByovpcPrivateSubnetTagKey)
```
I'm not sure that this current error message makes sense

mjlshen (approved):
/lgtm

PR #597, opened 2024-07-15
[OSD-24377] Add new osdctl command to allow silencing alert(s) across a specific org

dakotalongRH:
Add a new command that allows SRE to silence all or a list of specific alerts across all clusters belonging to a specific organization.

Usage:
`Usage:
  osdctl alert silence org <org-id> [--all --duration --comment | --alertname --duration --comment] [flags]

Flags:
      --alertname strings   alertname (comma-separated)
  -a, --all                 add silences for all alert
  -c, --comment string      add comment about silence
  -d, --duration string     add duration for silence (default "15d")
  -h, --help                help for org
`

### DIFF
--- a/cmd/alerts/silence/add_silence.go
+++ b/cmd/alerts/silence/add_silence.go
@@ -69,12 +69,12 @@ func AddSilence(cmd *addSilenceCmd) {
 	if all {
 		err := AddAllSilence(clusterID, duration, comment, username, clustername, kubeconfig, clientset)
 		if err != nil {
-			fmt.Printf("Failed to add silence: %w", err)
+			fmt.Printf("Failed to add silence: %s", err)
 		}
 	} else if len(alertID) > 0 {
 		err := AddAlertNameSilence(alertID, duration, comment, username, kubeconfig, clientset)
 		if err != nil {
-			fmt.Printf("Failed to add silence: %w", err)
+			fmt.Printf("Failed to add silence: %s", err)
 		}
 	} else {
 		fmt.Println("No valid option specified. Use --all or --alertname.")
--- a/cmd/alerts/silence/add_silence.go
+++ b/cmd/alerts/silence/add_silence.go
@@ -69,12 +69,12 @@ func AddSilence(cmd *addSilenceCmd) {
 	if all {
 		err := AddAllSilence(clusterID, duration, comment, username, clustername, kubeconfig, clientset)
 		if err != nil {
-			fmt.Errorf("Failed to add silence: %w", err)
+			fmt.Printf("Failed to add silence: %w", err)
 		}
 	} else if len(alertID) > 0 {
 		err := AddAlertNameSilence(alertID, duration, comment, username, kubeconfig, clientset)
 		if err != nil {
-			fmt.Errorf("Failed to add silence: %w", err)
+			fmt.Printf("Failed to add silence: %w", err)
 		}
 	} else {
 		fmt.Println("No valid option specified. Use --all or --alertname.")
--- a/cmd/alerts/silence/add_silence.go
+++ b/cmd/alerts/silence/add_silence.go
@@ -75,7 +75,7 @@ func AddSilence(cmd *addSilenceCmd) {
 	}
 
 }
-func AddAllSilence(clusterID, duration, comment, username, clustername string, kubeconfig *rest.Config, clientset *kubernetes.Clientset) {
+func AddAllSilence(clusterID, duration, comment, username, clustername string, kubeconfig *rest.Config, clientset *kubernetes.Clientset) error {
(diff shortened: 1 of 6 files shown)


cmd/alerts/silence/silence_org.go
```diff
@@ -0,0 +1,96 @@
...
+			log.Fatal(err)
+		}
+
+		if all {
+			AddAllSilence(clusterID, duration, comment, username, clustername, kubeconfig, clientset)
+		} else if len(alertID) > 0 {
+			AddAlertNameSilence(alertID, duration, comment, username, kubeconfig, clientset)
+		} else {
+			fmt.Println("No valid option specified. Use --all or --alertname.")
+		}
```
tnierman, line 94: It looks like the functions `AddAllSilence` and `AddAlertNameSilence` will panic if they encounter an error, rather that bubble it up.

I think, for this implementation specifically, it'd be better to log and collect the failures rather than exiting immediately; that way we can still potentially silence the remaining clusters belonging to this organization if we hit a snag early on in the loop. 

Can we refactor those functions here so that they return their errors?

dakotalongRH: Agreed, updated.

cmd/alerts/silence/add_silence.go
```diff
@@ -75,7 +75,7 @@ func AddSilence(cmd *addSilenceCmd) {
 	}
 
 }
-func AddAllSilence(clusterID, duration, comment, username, clustername string, kubeconfig *rest.Config, clientset *kubernetes.Clientset) {
+func AddAllSilence(clusterID, duration, comment, username, clustername string, kubeconfig *rest.Config, clientset *kubernetes.Clientset) error {
```
AlexVulaj, line 78: 1. It looks like `clusterID` and `clustername` never get used by this function - can we blow them away?
2. Now that this function returns an error, we should make sure that any call sites are handling that error, such as line 70 in this file.

cmd/alerts/silence/add_silence.go
```diff
@@ -117,7 +119,7 @@ func fetchAllAlerts(kubeconfig *rest.Config, clientset *kubernetes.Clientset) []
 	return fetchedAlerts
 }
 
-func AddAlertNameSilence(alertID []string, duration, comment, username string, kubeconfig *rest.Config, clientset *kubernetes.Clientset) {
+func AddAlertNameSilence(alertID []string, duration, comment, username string, kubeconfig *rest.Config, clientset *kubernetes.Clientset) error {
```
AlexVulaj, line 122: Line 72 in this file has a call to this function without handling the error - let's wrap that

cmd/alerts/silence/silence_org.go
```diff
@@ -0,0 +1,105 @@
...
+
+	cmd.Flags().StringSliceVar(&AddOrgSilenceCmd.alertID, "alertname", []string{}, "alertname (comma-separated)")
+	cmd.Flags().StringVarP(&AddOrgSilenceCmd.comment, "comment", "c", "", "add comment about silence")
+	cmd.Flags().StringVarP(&AddOrgSilenceCmd.duration, "duration", "d", "15d", "add duration for silence") //default duration set to 15 days
+	cmd.Flags().BoolVarP(&AddOrgSilenceCmd.all, "all", "a", false, "add silences for all alert")
+
+	return cmd
+}
+
+// Add alert silences to organization's clusters
```
AlexVulaj, line 43: stylistic nitpick - Go function docs usually start with the name of the function
```suggestion
// AddOrgSilence Add alert silences to organization's clusters
```

cmd/alerts/silence/silence_org.go
```diff
@@ -0,0 +1,105 @@
...
+func AddOrgSilence(cmd *AddOrgSilenceCmd) {
+	alertID := cmd.alertID
+	comment := cmd.comment
+	duration := cmd.duration
+	all := cmd.all
+	organizationID := cmd.organization
+
+	//Retrieve v1.Subscriptions list
```
AlexVulaj, line 51: I'm personally against comments like this, line 59, line 65, line 78, etc. Your code is written cleanly enough that it's obvious what's going on here.

dustman9000:
Can we please require that a Jira card is tracked with this org silence? I see the we have -c (comment) as a required parameter, but I would take it a step further and also require -j (jira card). We need a process around this to track the Jira and eventually unsilence the org for this alert.

dakotalongRH:
> Can we please require that a Jira card is tracked with this org silence? I see the we have -c (comment) as a required parameter, but I would take it a step further and also require -j (jira card). We need a process around this to track the Jira and eventually unsilence the org for this alert.

Agreed @dustman9000 , made the comment flag required and added a message that an OHSS is required to do an org-wide silence

PR #600, opened 2024-07-23
Integrate verifier 1RC

AlexVulaj:
Updates verifier to use version v1.0.0-rc.1

Some other minor bug fixes as well that came up during testing `osdctl network verify-egress` without a `--cluster-id`

Resolves OSD-24731


cmd/network/verification.go
```diff
@@ -5,6 +5,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/openshift/osd-network-verifier/pkg/probes/curl"
+	"github.com/openshift/osd-network-verifier/pkg/probes/legacy"
```
mjlshen, line 9: nit: can we move these imports into the lower block

cmd/network/verification.go
```diff
@@ -366,6 +372,16 @@ func (e *EgressVerification) generateAWSValidateEgressInput(ctx context.Context,
 	// Forward the timeout
 	input.Timeout = e.EgressTimeout
 
+	switch strings.ToLower(e.Probe) {
+	case "curl":
+		input.Probe = curl.Probe{}
+	case "legacy":
+		input.Probe = legacy.Probe{}
+	default:
+		fmt.Printf("Unrecognized probe %s - defaulting to curl probe.", e.Probe)
```
mjlshen, line 381:
```suggestion
		e.log.Info(ctx, "unrecognized probe %s - defaulting to curl probe.", e.Probe)
```
This command logs these kinds of messages

cmd/network/verification_test.go
```diff
@@ -64,20 +64,56 @@ func Test_egressVerificationSetup(t *testing.T) {
...
 		},
 		{
 			name: "ClusterId optional",
 			e: &EgressVerification{
-				ClusterId:       "",
 				SubnetIds:       []string{"subnet-a", "subnet-b", "subnet-c"},
 				SecurityGroupId: "sg-b",
+				PlatformType:    "aws",
 			},
-			expectErr: false,
```
mjlshen, line 80: I think we should put this back. Yes, the behavior is the same with or without the line, but it's more obvious what the intent of the test is if `expectErr: false` is here.

mjlshen (changes requested):
Just nits, core changes make sense

mjlshen (approved):
/lgtm

PR #602, opened 2024-07-25
Add cpu arch flag to network verifier command

AlexVulaj:
Add a new flag for cpu architecture to the network verifier wrapper in osdctl.

Resolves [OSD-24768](https://issues.redhat.com//browse/OSD-24768)

### DIFF
--- a/cmd/network/verification.go
+++ b/cmd/network/verification.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/openshift/osd-network-verifier/pkg/data/cpu"
 	"log"
 	"os"
 	"strings"
@@ -77,6 +78,9 @@ type EgressVerification struct {
 	Version bool
 	// Probe The type of probe to use for the verifier
 	Probe string
+	// CpuArchName the architecture to use for the compute instance
+	CpuArchName string
+	cpuArch     cpu.Architecture
 }
 
 func NewCmdValidateEgress() *cobra.Command {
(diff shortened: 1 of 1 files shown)


cmd/network/verification.go
```diff
@@ -214,6 +219,10 @@ func (e *EgressVerification) setup(ctx context.Context) (*aws.Config, error) {
 	}
 	e.log = logger
 
+	if validArches := []string{"", "x86", "x86_64", "amd64", "arm", "arm64", "aarch64"}; !slices.Contains(validArches, e.CpuArchName) {
+		return nil, fmt.Errorf("%s is not a valid architecture - must use one of %s", e.CpuArchName, strings.Join(validArches, ","))
+	}
+
```
abyrne55, line 225: This check isn't really necessary as ` cpu.ArchitectureByName(e.CpuArchName)` will return an empty `CPUArchitecture{}` that will fail `IsValid()` and throw "unknown CPU architecture" (lines 319-322). I'd suggest removing this so that you don't have to import `slices` and so that there's only one source of truth when it comes to the arch's that the verifier supports

cmd/network/verification.go
```diff
@@ -133,6 +137,7 @@ func NewCmdValidateEgress() *cobra.Command {
 	validateEgressCmd.Flags().DurationVar(&e.EgressTimeout, "egress-timeout", 2*time.Second, "(optional) timeout for individual egress verification requests")
 	validateEgressCmd.Flags().BoolVar(&e.Version, "version", false, "When present, prints out the version of osd-network-verifier being used")
 	validateEgressCmd.Flags().StringVar(&e.Probe, "probe", "curl", "(optional) select the probe to be used for egress testing. Either 'curl' (default) or 'legacy'")
+	validateEgressCmd.Flags().StringVar(&e.CpuArchName, "cpu-arch", "x86", "(optional) compute instance CPU architecture. Either 'x86'/'x86_64'/'amd64' or 'arm'/'arm64'/'aarch64'")
```
abyrne55, line 140:
```suggestion
	validateEgressCmd.Flags().StringVar(&e.CpuArchName, "cpu-arch", "x86", "(optional) compute instance CPU architecture. E.g., 'x86' or 'arm'")
```
Better to simplify this IMO

abyrne55 (changes requested):
Nice work, @AlexVulaj. Haven't yet tested locally, so just left a few minor suggestions based on a readthru

cmd/network/verification.go
AlexVulaj: I want us to error out early on here. Otherwise we're going through backplane to grab the cluster credentials, setting up an AWS config, fetching info, etc. This wastes a lot of resources if we know right from the get go that we have an invalid flag value.

mjlshen: I would want network-verifier to expose an arch validation function (if it's too costly), it doesn't seem to make sense for osdctl to pre-validate this before passing it to network-verifier.

AlexVulaj: Idea incoming in a new commit.

abyrne55: Can do low-cost early arch evaluation by calling cpu.ArchitectureByName(e.CpuArchName).IsValid() early on

AlexVulaj: @abyrne55 +1, this is what I ended up going with.

cmd/network/verification.go
```diff
@@ -214,6 +219,11 @@ func (e *EgressVerification) setup(ctx context.Context) (*aws.Config, error) {
 	}
 	e.log = logger
 
+	e.cpuArch = cpu.ArchitectureByName(e.CpuArchName)
```
AlexVulaj, line 222: @abyrne55 @mjlshen PTAL - this was my solution to our other thread.

cmd/network/verification.go
```diff
@@ -307,6 +317,8 @@ func (e *EgressVerification) generateAWSValidateEgressInput(ctx context.Context,
 	e.log.Info(ctx, "using platform: %s", platform)
 	input.PlatformType = platform
 
+	input.CPUArchitecture = e.cpuArch
```
mjlshen, line 320: Is there a use-case for this feature at this time? All clusters are x86 right? HCP clusters in the future can be multiarch (mix of x86/arm), how would an SRE decide which architecture to use (does it matter)?

If there's some criteria, I would expect this to be auto-populated from OCM data about the cluster for SRE.

abyrne55 (approved):
Tested locally with `--cpu-arch=arm` and confirmed via AWS console that an ARM-based machine type (t4g.micro) was launched. Probe completed successfully. Great work @AlexVulaj 

/lgtm

cmd/network/verification.go
AlexVulaj: Eventually we'll probably determine it automatically from `--cluster-id` like the other fields, but this will allow us to start testing asap against AWS enviroments.

PR #604, opened 2024-08-08
Allow reading creds as env var for account mgmt commands

sam-nguyen7:
https://issues.redhat.com/browse/OSD-20603

Update osdctl account mgmt commands to use RH SSO credentials generated via https://github.com/app-sre/rh-aws-saml-login which spawns a new shell with the AWS_ACCOUNT_NAME environment variable.

### DIFF
--- a/cmd/account/mgmt/account-assign.go
+++ b/cmd/account/mgmt/account-assign.go
@@ -3,6 +3,7 @@ package mgmt
 import (
 	"fmt"
 	"math/rand"
+	"os"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/service/organizations"
@@ -80,9 +81,6 @@ func (o *accountAssignOptions) complete(cmd *cobra.Command, _ []string) error {
 	if o.username == "" {
 		return cmdutil.UsageErrorf(cmd, "LDAP username was not provided")
 	}
-	if o.payerAccount == "" {
-		return cmdutil.UsageErrorf(cmd, "Payer account was not provided")
-	}
 
 	o.output = o.GlobalOptions.Output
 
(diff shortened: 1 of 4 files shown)


sam-nguyen7:
/hold for testing
assign command fails due to: `error: operation error Organizations: CreateAccount, https response error StatusCode: 400, RequestID: bb0e1fda-4bed-411d-a2c4-1ad6d68594f5, AccessDeniedException: You don't have permissions to access this resource.`

MR was created to add permission: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/114730

cmd/account/mgmt/account-assign.go
```diff
@@ -97,10 +95,10 @@ func (o *accountAssignOptions) run() error {
 		rootID          string
 	)
 
-	if o.payerAccount == "osd-staging-1" {
+	if o.payerAccount == "osd-staging-1" || os.Getenv("AWS_ACCOUNT_NAME") == "osd-staging-1" {
```
iamkirkbater, line 98: Can we pull `osd-staging-1` and the osd-staging-2 string out into constants?

Only adding the comment once but this would apply to all of the files contained within this PR.

iamkirkbater (changes requested):
Can I ask why we're going with `osd-staging-1` and `osd-staging-2-prod`? Can we standardize on just being `osd-staging-#`? I specifically worry about `osd-staging-2-prod` because staging-2 is not prod and should not be used for any prod things.

Additionally can we pull `osd-staging-1` and `osd-staging-2` out into constants somewhere within osdctl that makes sense and reference them from that?

cmd/account/mgmt/account-assign.go
AlexVulaj: Piggy backing here - can we also pull "AWS_ACCOUNT_NAME" out as a constant?

cmd/account/mgmt/account-assign.go
```diff
@@ -97,10 +95,16 @@ func (o *accountAssignOptions) run() error {
 		rootID          string
 	)
 
-	if o.payerAccount == "osd-staging-1" {
+	const (
+		osdStaging1    = "osd-staging-1"
+		osdStaging2    = "osd-staging-2"
+		awsAccountName = "AWS_ACCOUNT_NAME"
+	)
```
iamkirkbater, line 102: One _smalllll_ little nit here - we're redeclaring the consts in each of these files. Could we define these somewhere like in `cmd.go` and just reference them from all of these commands?

cmd/account/mgmt/account-assign.go
```diff
@@ -97,10 +95,16 @@ func (o *accountAssignOptions) run() error {
 		rootID          string
 	)
 
-	if o.payerAccount == "osd-staging-1" {
+	const (
+		osdStaging1    = "osd-staging-1"
+		osdStaging2    = "osd-staging-2"
+		awsAccountName = "AWS_ACCOUNT_NAME"
```
iamkirkbater, line 101: Can we make this const name more descriptive? Something like `EnvKeyAWSAccountName`?

iamkirkbater (changes requested):
Two _small_ changes, one just a small nit about the usage of constants and the other just a better name for a var.

Thanks!

iamkirkbater:
/lgtm

Thanks for making the changes!

PR #606, opened 2024-08-13
OCM-10376 | Add Limited Support Reason Override support

mnecas:
Right now we have new override parameters in the LSR so we can keep the LSR visible to the SRE but it can be `overriden` when there is support exception. The cluster status shows only active LSR, without the override.


mnecas:
@AlexVulaj could you PTAL?

NautiluX (commented):
I think I'm missing some detail on the different options to get the number of reasons, can you explain - or even better make in more obvious in the code - why sometimes len() is the right thing and sometimes we us a count that's being returned count from the function?

cmd/cluster/support/status.go
```diff
@@ -58,23 +59,26 @@ func (o *statusOptions) complete(cmd *cobra.Command, args []string) error {
...
 	// No reasons found, cluster is fully supported
-	if len(clusterLimitedSupportReasons) == 0 {
+	if count == 0 {
 		fmt.Printf("Cluster is fully supported\n")
 		return nil
 	}
 
 	table := printer.NewTablePrinter(os.Stdout, 20, 1, 3, ' ')
-	table.AddRow([]string{"Reason ID", "Summary", "Details"})
+	table.AddRow([]string{"Reason ID", "Summary", "Override", "Details"})
```
mjlshen, line 74: As an SRE, the meaning of `Override` isn't immediately obvious to me. I have two ideas in mind that would mitigate my concern, either titling this column `Overridden (SUPPORTEX)` or explaining that override means that a specific limited support reason has been overridden in the help text around line 30.

mjlshen (commented):
Code LGTM

cmd/cluster/support/common.go
```diff
@@ -32,14 +32,14 @@ func getLimitedSupportReasons(clusterId string) ([]*cmv1.LimitedSupportReason, e
...
 
 	//getting the limited support reasons for the cluster
 	clusterLimitedSupportReasons, err := ctlutil.GetClusterLimitedSupportReasons(connection, cluster.ID())
 	if err != nil {
-		return nil, errors.New(fmt.Sprintf("Can't retrieve cluster limited support reasons: %v\n", err))
+		return nil, -1, errors.New(fmt.Sprintf("Can't retrieve cluster limited support reasons: %v\n", err))
 	}
 
-	return clusterLimitedSupportReasons, nil
+	return clusterLimitedSupportReasons, cluster.Status().LimitedSupportReasonCount(), nil
```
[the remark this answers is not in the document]
mnecas: with the override, we still return LSRs but they can be overridden
this is reflected in the count

cmd/cluster/support/delete.go
```diff
@@ -117,10 +117,10 @@ func (o *deleteOptions) run() error {
 	/*conditions to check the presence of --all & -i flags;
 	also, checking if there is one or more limited support resonID before deleting the same */
 	var limitedSupportReasonIds []string
-	limitedSupportReasons, err := getLimitedSupportReasons(o.clusterID)
+	limitedSupportReasons, _, err := getLimitedSupportReasons(o.clusterID)
```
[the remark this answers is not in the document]
mnecas: This is for deletion, you still can delete overridden LSRs but if you have all overridden LSRs then you want to show that the cluster status is supported.

cmd/cluster/support/status.go
mnecas: ah good note `Overridden (SUPPORTEX)` sounds good to me

cmd/cluster/support/common.go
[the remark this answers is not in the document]
mjlshen: fyi dupe of https://github.com/openshift/osdctl/pull/606#discussion_r1754311015 - I agree it's confusing, but `cluster.Status().LimitedSupportReasonCount()` doesn't count limited support reasons that have been overridden, so it can be different from `len(clusterLimitedSupportReasons)`, which doesn't take into account overrides.

cmd/cluster/support/delete.go
```diff
@@ -117,10 +117,10 @@ func (o *deleteOptions) run() error {
...
 
 	if len(limitedSupportReasons) == 0 {
-		return fmt.Errorf("Cluster is not in limited support. \n")
+		return fmt.Errorf("No limited support reasons found or all reasons are overridden, the cluster is fully supported \n")
```
NautiluX, line 123: Is that accurate here? I think here were using the _actual_ count of limited support reasons, so it will include overridden ones, right?

cmd/cluster/support/delete.go
```diff
@@ -120,7 +120,7 @@ func (o *deleteOptions) run() error {
 	limitedSupportReasons, err := getLimitedSupportReasons(o.clusterID)
 
 	if len(limitedSupportReasons) == 0 {
-		return fmt.Errorf("Cluster is not in limited support. \n")
+		return fmt.Errorf("No limited support reasons found or all reasons are overridden, the cluster is fully supported \n")
```
NautiluX, line 123:
```suggestion
		return fmt.Errorf("Cluster is not in limited support. \n")
```

NautiluX (commented):
/lgtm

PR #608, opened 2024-09-10
OSD-25183: Update Dynatrace related commands for osdctl

devppratik:
The aim of the PR is:

- Fetches HCP Namespace for HCP Cluster without Logging into the MC. It now fetches using OCM.
- Includes HCP Namespace for logs command when it's an HCP Cluster
- Adds Dynatrace Logs URL to cluster context command for HCP and MC
- Refactors some of the code and moves to OCM package

Also fixes [OSD-25323](https://issues.redhat.com//browse/OSD-25323) as it does not depend on cluster login anymore for getting HCP namespace


cmd/cluster/context.go
```diff
@@ -395,26 +397,28 @@ func (o *contextOptions) generateContextData() (*contextData, []error) {
...
 		var clusterID string = o.clusterID
 		defer wg.Done()
 		defer utils.StartDelayTracker(o.verbose, "Dynatrace URL").End()
 
-		clusterID, _, err := dynatrace.GetManagementCluster(ocmClient, o.cluster)
+		hcpCluster, err := dynatrace.FetchClusterDetails(clusterID)
 		if err != nil {
-			errors = append(errors, err)
-			data.DyntraceEnvURL = err.Error()
+			if strings.Contains(err.Error(), "not an HCP or MC Cluster") {
```
rendhalver, line 407: Is this an exact string match?
There should be a better way to do this that doesn't involve checking for a string that may change.

In the mean time we should make a note in `cmd/cluster/dynatrace/cluster.go` that we are using this string to check error conditions that we are passing out to the cli.

cmd/cluster/context.go
```diff
@@ -395,26 +397,28 @@ func (o *contextOptions) generateContextData() (*contextData, []error) {
...
+				data.DyntraceEnvURL = "Not an HCP/MC Cluster"
+			} else {
+				errors = append(errors, fmt.Errorf("failed to acquire cluster details %v", err))
```
rendhalver, line 410: Same as previous comment.

cmd/cluster/context.go
devppratik: > There should be a better way to do this that doesn't involve checking for a text string that may change.
Good point. I can update this to a different return value and make the changes there

iamkirkbater: I think an alternate approach would be to export a specific type of error and check for that.

So in `dynatrace/cluster.go` we would have the following:

```go
var ErrUnsupportedCluster = fmt.Errorf("not an HCP or MC Cluster")

...

if !isMC || err != nil {
			// if the cluster is not a HCP or MC, then return an error
			return HCPCluster{}, ErrUnsupportedCluster
		} else {
...
```

And here in this file we'd do a check like this:
```go
hcpCluster, err := dynatrace.FetchClusterDetails(clusterID)
		if err != nil {
			if err == dynatrace.ErrUnsupportedCluster {
			    ...
```

Some of the benefits to an approach like this is being able to _distinctly_ check which exact error it is without having to do string matching, so if at a later time we decide that we want to change the error message to something like:

```go
var ErrUnsupportedCluster = fmt.Errorf("Not an HCP, MC or SC Cluster")
```
We have no further code changes that need to be made.

I won't _hold_ this PR up over this point, as your method is not an invalid way to handle this, however I always tend to prefer safer checks over string matching :)

devppratik: Thanks, @iamkirkbater, For the suggestion. This definitely makes much more sense to pass the error. 
PR updated with the changes

PR #609, opened 2024-09-13
OSD-24259: osdctl network verify should prompt putting clusters into LS if DMS/PD is blocked

TheUndeadKing:
**JIRA**: 
[OSD-24259](https://issues.redhat.com/browse/OSD-24259)

**SUMMARY**:
If cluster can't access DMS (OR) PD while validating network using _verify-egress_. It should ask prompt to move cluster into LS instead of SL prompt

### DIFF
--- a/cmd/network/verification.go
+++ b/cmd/network/verification.go
@@ -178,17 +178,18 @@ func (e *EgressVerification) Run(ctx context.Context) {
 		out := onv.ValidateEgress(c, *inputs[i])
 		out.Summary(e.Debug)
 
-		// Move cluster into LS if in case cluster can't connect with PD(OR)DMS || service log if the failures other egress-url related
+		// Prompt putting the cluster into LS if egresses crucial for monitoring (PagerDuty/DMS) are blocked.
+		// Prompt sending a service log instead for other blocked egresses.
 		if !out.IsSuccessful() && len(out.GetEgressURLFailures()) > 0 {
 			postCmd := generateServiceLog(out, e.ClusterId)
 
-			bUrl := strings.Join(postCmd.TemplateParams, ",")
+			blockedUrl := strings.Join(postCmd.TemplateParams, ",")
 
-			if strings.Contains(bUrl, "deadmanssnitch") || strings.Contains(bUrl, "pagerduty") {
-				fmt.Println("This cluster traffic has block pagerdury (OR) DMS outgoing traffic")
+			if strings.Contains(blockedUrl, "deadmanssnitch") || strings.Contains(blockedUrl, "pagerduty") {
+				fmt.Println("PagerDuty and/or DMS outgoing traffic is blocked, resulting in a loss of observability. As a result, Red Hat can no longer guarantee SLAs and the cluster should be put in limited support")
 				err := putLimitedSupport(e.ClusterId)
 				if err != nil {
-					log.Fatal(err)
+					fmt.Println("failed to post limited support reason: %w", err)
 				}
 
 			} else if err := postCmd.Run(); err != nil {
--- a/cmd/network/verification.go
+++ b/cmd/network/verification.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/openshift/osd-network-verifier/pkg/data/cpu"
 	"log"
 	"os"
 	"strings"
@@ -15,9 +14,12 @@ import (
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/openshift-online/ocm-cli/pkg/dump"
+	sdk "github.com/openshift-online/ocm-sdk-go"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/openshift-online/ocm-sdk-go/logging"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	"github.com/openshift/osd-network-verifier/pkg/data/cpu"
 	"github.com/openshift/osd-network-verifier/pkg/helpers"
 	"github.com/openshift/osd-network-verifier/pkg/output"
 	"github.com/openshift/osd-network-verifier/pkg/probes/curl"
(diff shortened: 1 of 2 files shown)


cmd/network/verification.go
```diff
@@ -176,13 +178,22 @@ func (e *EgressVerification) Run(ctx context.Context) {
...
-		// Only suggest sending a service log if the failures are egress-url related
+		// Move cluster into LS if in case cluster can't connect with PD(OR)DMS || service log if the failures other egress-url related
 		if !out.IsSuccessful() && len(out.GetEgressURLFailures()) > 0 {
 			postCmd := generateServiceLog(out, e.ClusterId)
-			if err := postCmd.Run(); err != nil {
+
+			bUrl := strings.Join(postCmd.TemplateParams, ",")
+
+			if strings.Contains(bUrl, "deadmanssnitch") || strings.Contains(bUrl, "pagerduty") {
+				fmt.Println("This cluster traffic has block pagerdury (OR) DMS outgoing traffic")
```
typeid, line 188:
```suggestion
				fmt.Println("PagerDuty and/or DMS outgoing traffic is blocked, resulting in a loss of observability. As a result, Red Hat can no longer guarantee SLAs and the cluster should be put in limited support.")
```

cmd/network/verification.go
```diff
@@ -176,13 +178,22 @@ func (e *EgressVerification) Run(ctx context.Context) {
 		out := onv.ValidateEgress(c, *inputs[i])
 		out.Summary(e.Debug)
 
-		// Only suggest sending a service log if the failures are egress-url related
+		// Move cluster into LS if in case cluster can't connect with PD(OR)DMS || service log if the failures other egress-url related
```
typeid, line 181:
```suggestion
		// Prompt putting the cluster into LS if egresses crucial for monitoring (PagerDuty/DMS) are blocked. 
		// Prompt sending a service log instead for other blocked egresses.
```

cmd/network/verification.go
```diff
@@ -790,3 +801,60 @@ func printVersion() {
...
+			os.Exit(1)
+		}
+	}()
+
+	summary := "Your cluster requires you to take action. SRE has observed that there have been changes made to the network configuration which impacts normal working of the cluster, including lack of network egress to these internet-based resources which are required for the cluster operation and support. Please revert changes, and refer to documentation regarding firewall requirements for PrivateLink clusters: https://access.redhat.com/documentation/en-us/red_hat_openshift_service_on_aws/4/html/prepare_your_environment/rosa-sts-aws-prereqs#osd-aws-privatelink-firewall-prerequisites_rosa-sts-aws-prereqs"
+	limitedSupportBuilder := cmv1.NewLimitedSupportReason().Summary("Cluster is in Limited Support due to unsupported network configuration").Details(summary).DetectionType("manual")
+	limitedSupport, err := limitedSupportBuilder.Build()
+	if err != nil {
+		return fmt.Errorf("failed to build new limited support reason: %w", err)
+	}
```
typeid, line 824: **suggestion:** We can likely make this more descriptive and also list blocked egresses. I suggest running this through region leads / management for approval afterwards.

cmd/network/verification.go
```diff
@@ -176,13 +178,22 @@ func (e *EgressVerification) Run(ctx context.Context) {
...
 			postCmd := generateServiceLog(out, e.ClusterId)
-			if err := postCmd.Run(); err != nil {
+
+			bUrl := strings.Join(postCmd.TemplateParams, ",")
+
+			if strings.Contains(bUrl, "deadmanssnitch") || strings.Contains(bUrl, "pagerduty") {
+				fmt.Println("This cluster traffic has block pagerdury (OR) DMS outgoing traffic")
+				err := putLimitedSupport(e.ClusterId)
+				if err != nil {
+					log.Fatal(err)
```
typeid, line 191: **nitpick:** bubble up the error instead

cmd/network/verification.go
```diff
@@ -176,13 +178,22 @@ func (e *EgressVerification) Run(ctx context.Context) {
 		out := onv.ValidateEgress(c, *inputs[i])
 		out.Summary(e.Debug)
 
-		// Only suggest sending a service log if the failures are egress-url related
+		// Move cluster into LS if in case cluster can't connect with PD(OR)DMS || service log if the failures other egress-url related
 		if !out.IsSuccessful() && len(out.GetEgressURLFailures()) > 0 {
 			postCmd := generateServiceLog(out, e.ClusterId)
-			if err := postCmd.Run(); err != nil {
+
+			bUrl := strings.Join(postCmd.TemplateParams, ",")
```
typeid, line 185: **nitpick:** we could give this parameter a proper name and re-use it in the SL sent / LS sent.

cmd/network/verification.go
```diff
@@ -176,13 +178,22 @@ func (e *EgressVerification) Run(ctx context.Context) {
...
+
+			if strings.Contains(bUrl, "deadmanssnitch") || strings.Contains(bUrl, "pagerduty") {
+				fmt.Println("This cluster traffic has block pagerdury (OR) DMS outgoing traffic")
+				err := putLimitedSupport(e.ClusterId)
```
typeid, line 189: **Question:** do we not have a LS command similarly to `postCmd.Run()` for service logs that we can re-use to avoid code duplication? We would specifically need this to avoid putting "critical customers" into limited support.

cmd/network/verification.go
```diff
@@ -790,3 +801,60 @@ func printVersion() {
 	}
 	log.Println(fmt.Sprintf("Using osd-network-verifier version %v", version))
 }
+
+// Put Cluster to LS if it can't access pd (OR) DMS
+func putLimitedSupport(ClusterId string) error {
```
typeid, line 806: **chore:** As mentioned above, I suggest re-using the existing LS post command to avoid code duplication.

typeid (changes requested):
Good stuff @TheUndeadKing - this will help us with consistency in the way we use limited support!

cmd/network/verification.go
TheUndeadKing: The postCmd is the o/p of `generateServiceLog` func. it doesn't have LS func

TheUndeadKing: Added func in putLimitedSupport to check the critical cu label is present or not

typeid: I meant that we can follow the same path for posting limited support as we're already following for sending service logs. We're not re-writing the code to send service logs, instead we're re-using what is in the `osdctl servicelog post` command. We can do the same for limited support. [This](https://github.com/openshift/osdctl/blob/master/cmd/cluster/support/post.go) code already posts limited support, we can make it re-usable instead of duplicating code :)

TheUndeadKing: I'd exported SendLimitedSupportPostRequest & PrintLimitedSupportReason  func in post.go, In order to re-use here

TheUndeadKing: reused post's struct to set cluster into LS. Removed all re-used code

cmd/network/verification.go
TheUndeadKing: reused post's struct to set cluster into LS. Removed all re-used code

TheUndeadKing:
Created Template PR: https://github.com/openshift/managed-notifications/pull/339

cmd/network/verification.go
```diff
@@ -790,3 +800,14 @@ func printVersion() {
...
+// Generate LimitedSupportTemplate
+func generateLimitedSupportTemplate(out *output.Output) lsupport.Post {
+	failures := out.GetEgressURLFailures()
+	if len(failures) > 0 {
+		return lsupport.Post{
+			Template: LimitedSupportTemplate,
+		}
+	}
+	return lsupport.Post{}
+}
```
typeid, line 813: Do we even need this function? I'm not even sure we would want to return `lsupport.Post{}`, ever. 
We also already check for egress count before calling this. 

Can we just do the following in the above logic, instead of having `generateLimitedSupportTemplate`? 
```
pCmd := lsupport.Post{Template: LimitedSupportTemplate}
if err := pCmd.Run(e.ClusterId); err != nil {
   fmt.Printf("failed to post limited support reason: %v", err)
}
```

cmd/network/verification.go
```diff
@@ -175,14 +177,22 @@ func (e *EgressVerification) Run(ctx context.Context) {
...
 		if !out.IsSuccessful() && len(out.GetEgressURLFailures()) > 0 {
 			postCmd := generateServiceLog(out, e.ClusterId)
-			if err := postCmd.Run(); err != nil {
+			blockedUrl := strings.Join(postCmd.TemplateParams, ",")
+			if strings.Contains(blockedUrl, "deadmanssnitch") || strings.Contains(blockedUrl, "pagerduty") {
+				fmt.Println("PagerDuty and/or DMS outgoing traffic is blocked, resulting in a loss of observability. As a result, Red Hat can no longer guarantee SLAs and the cluster should be put in limited support")
+				pCmd := generateLimitedSupportTemplate(out)
+				if err := pCmd.Run(e.ClusterId); err != nil {
+					fmt.Printf("failed to post limited support reason: %v", err)
+
```
typeid, line 190: **nit:** remove empty line

cmd/network/verification.go
TheUndeadKing: Applied the above changes.

cmd/network/verification.go
TheUndeadKing: Removed empty line

typeid (approved):
/lgtm
/approve

PR #615, opened 2024-10-10
Bump osd-network-verifier to v1.1.2

abyrne55:
This PR addresses [OSD-25867](https://issues.redhat.com/browse/OSD-25867), which calls for the latest version (v1.1.2) of osd-network-verifier to be integrated into `osdctl network verify-egress`. This new version came with breaking changes to the API: specifically, we now use a `Platform` struct type instead of a string to specify the cloud platform under test. This latest version of the verifier also includes experimental support for HCP zero egress clusters; enable this with `--platform=aws-hcp-zeroegress`.


mjlshen (approved):
/lgtm

PR #620, opened 2024-10-26
Add 'list' and 'describe' sre_operators to osdctl cluster commands

MateSaary:


clcollins:
@MateSaary - could you add a description to this PR providing some human-friendly info and the relevant Jira card if there is one, please?

MateSaary:
Hi @clcollins apologies for the oversight, added!

cmd/cluster/sre_operators/describeCmd.go
```diff
@@ -0,0 +1,331 @@
...
+	sreOperatorsDescribeExample = `
+		# Describe SRE operators
+		$ osdctl cluster sre-operators describe <operator-name>
+	`
+	sreOperatorsLongDescription = `
+  Helps obtain various health information about a specified SRE operator within a cluster,
+  including CSV, Subscription, OperatorGroup, Deployment, and Pod health statuses.
+
+  A GitLab access token is required to fetch the latest version of the operator,
+  and can be set in the '.config/osdctl' file as 'gitlab_access: <TOKEN>'.
```
devppratik, line 55: Since this is a new config added to osdctl, can we also add this to the setup command to ensure it is setup properly?

cmd/cluster/sre_operators/describeCmd.go
```diff
@@ -0,0 +1,331 @@
...
+
+	if err := p.Flush(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (ctx *sreOperatorsDescribeOptions) DescribeOperator(cmd *cobra.Command, operatorName string) ([]sreOperatorDetails, error) {
+	listOfOperators := []string{
```
devppratik, line 122: Is it possible if we can get this list dynamically from repositories or somewhere?

cmd/cluster/sre_operators/describeCmd.go
```diff
@@ -0,0 +1,331 @@
...
+		"deadmanssnitch-operator",
+		"openshift-deployment-validation-operator",
+		"gcp-project-operator",
+		"openshift-velero",
+		"pagerduty-operator",
+		"openshift-route-monitor-operator",
+		"openshift-observability-operator",
+	}
+
+	listOfOperatorNames := []string{
```
devppratik, line 146: Is it possible if we can get this list dynamically from repositories or somewhere?

cmd/cluster/sre_operators/listCmd.go
```diff
@@ -0,0 +1,475 @@
...
+  Lists the current version, channel, and status of SRE operators running in the current 
+  cluster context, and by default fetches the latest version from the operators' repositories.
+  
+  A GitLab access token is required to fetch the latest version of the operators,
+  and can be set in the '.config/osdctl' file as 'gitlab_access: <TOKEN>'.
+
+  The command creates a Kubernetes client to access the current cluster context, and GitLab/GitHub
+  clients to fetch the latest versions of each operator from its respective repository.
+	`
+	repositoryBranch = "production"
```
devppratik, line 77: This would make the command specific to prod only. We can try to fetch the environment from OCM directly or add a note that this command would only be relevant to Prod and should actually fail in Stg/Int

cmd/cluster/sre_operators/describeCmd.go
```diff
@@ -0,0 +1,331 @@
...
+	genericclioptions.IOStreams
+	kubeCli client.Client
+}
+
+type sreOperatorDetails struct {
+	Name      string
+	Current   string
+	Expected  string
+	CsvStatus string
+	Channel   string
```
devppratik, line 32: This is the same as the `type sreOperator struct` 
Can we use the Other Struct as a part of this one?

devppratik: Also, looking at the code, there are a lot of structs used. Can we maybe move them out to a separate file and add relevant comments? 
I am fine with having the current implementation as well

cmd/cluster/sre_operators/listCmd.go
```diff
@@ -0,0 +1,475 @@
...
+}
+
+type operatorResult struct {
+	Operator sreOperator
+	Error    error
+}
+
+func (ctx *sreOperatorsListOptions) ListOperators(cmd *cobra.Command) ([]sreOperator, error) {
+
+	listOfOperators := []string{
```
devppratik, line 167: This seems to be repeated in both `describe` and `list` files. Can we move the definition to either of the files or a separate file entirely

cmd/cluster/sre_operators/describeCmd.go
```diff
@@ -0,0 +1,331 @@
...
+		Group:   "operators.coreos.com",
+		Version: "v1",
+		Kind:    "OperatorGroup",
+	})
+	opGroupList := &unstructured.UnstructuredList{}
+	opGroupList.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "operators.coreos.com",
+		Version: "v1",
+		Kind:    "OperatorGroupList",
+	})
```
devppratik, line 205: I see these are repeated as well across the files

devppratik (changes requested):
I see quite a lot of code which is common for both of the sub commands. Please ensure to refactor those to a single file (I think list would be better) or move them out entirely to a different file to keep things tidy
Feel free to reach out if any help is required or anything to discuss

cmd/cluster/sre_operators/describeCmd.go
MateSaary: Hi Pratik thank you very much for the suggestion, added.

cmd/cluster/sre_operators/describeCmd.go
MateSaary: As this command is specific to SRE operators there isn't anywhere consistent containing those, and would also likely add to the runtime of the command.

cmd/cluster/sre_operators/listCmd.go
MateSaary: Unfortunately not all of the operators' repositories obey the stg/int/prod branching format, so it wouldn't be possible to correctly fetch versions based on this. The command still works in stg/int, with only a select few operators displaying different versions, although this is made clear by the 'channel' column of the command (i.e. on stg an operator using staging channel will be shown as such)

cmd/cluster/sre_operators/describeCmd.go
MateSaary: Thank you for the suggestion, changed.

cmd/cluster/sre_operators/listCmd.go
MateSaary: Refactored to be delcared only once in listCmd file, thanks for the suggestion

MateSaary:
@devppratik thank you so much for all the suggestions and review, I refactored the code slightly taking the suggestions into account, but keeping the majority of the code still separate for readability and for some differences between much of the common code.

cmd/cluster/sre_operators/listCmd.go
```diff
@@ -0,0 +1,477 @@
...
+	RestoreColor = "\033[0m"
+	Red          = "\033[31m"
+	Bold         = "\u001b[1m"
+	Gap          = "         "
+)
+
+var csv = &unstructured.Unstructured{}
+var csvList = &unstructured.UnstructuredList{}
+var sub = &unstructured.Unstructured{}
+var subList = &unstructured.UnstructuredList{}
```
joshbranham, line 132: I know you are using `DeepCopy` but we really shouldn't be using global variables, especially when goroutines are involved. Pass these around as dependencies or define them inline of your functions.

cmd/cluster/sre_operators/listCmd.go
```diff
@@ -0,0 +1,477 @@
...
+	
+	# List a specific SRE operator
+	$ osdctl cluster sre-operators list --operator='OPERATOR_NAME'
+	`
+	sreOperatorsListDescription = `
+	Lists the current version, channel, and status of SRE operators running in the current 
+	cluster context, and by default fetches the latest version from the operators' repositories.
+	
+	A GitLab access token is required to fetch the latest version of the operators, and can be 
+	set in the '.config/osdctl' file as 'gitlab_access: <TOKEN>', or using the 'osdctl setup' command.
```
joshbranham, line 65: If we are providing a command to configure the credential programmatically, we shouldn't encourage people to do it manually.

joshbranham (commented):
Left some misc code related comments, additionally this should have _some_ tests.

Can you speak to the difference in this command versus just using the data already provided by this [dashboard](https://grafana.app-sre.devshift.net/d/7OSPFbtId/sre-operators?orgId=1)?

zmird-r:
> Can you speak to the difference in this command versus just using the data already provided by this [dashboard](https://grafana.app-sre.devshift.net/d/7OSPFbtId/sre-operators?orgId=1)?

I can answer this: as you already mentioned, that is a dashboard with an overview of the whole fleet. This is a command that provides a view of a single cluster listing the operators installed on it. This command also provides the latest version of operators, which is a piece of information that the dashboard does not and can not provide so doing the same with metrics is not possible.

cmd/cluster/sre_operators/listCmd.go
MateSaary: Thank you for the suggestion, reverted

cmd/cluster/sre_operators/listCmd.go
MateSaary: I have changed the message, thank you for the suggestion!

joshbranham:
> > Can you speak to the difference in this command versus just using the data already provided by this [dashboard](https://grafana.app-sre.devshift.net/d/7OSPFbtId/sre-operators?orgId=1)?
> 
> I can answer this: as you already mentioned, that is a dashboard with an overview of the whole fleet. This is a command that provides a view of a single cluster listing the operators installed on it. This command also provides the latest version of operators, which is a piece of information that the dashboard does not and can not provide so doing the same with metrics is not possible.

I see, well we could add a cluster variable to that dashboard to see outdated per cluster. Have we thought about exposing a way in the operator CRD to include if it's outdated or not?

zmird-r:
> I see, well we could add a cluster variable to that dashboard to see outdated per cluster. Have we thought about exposing a way in the operator CRD to include if it's outdated or not?

I don't know how would you do that. That piece of information is generated dynamically in the operator bundle repository and needs to be matched with the commit set in the app-interface config file of the operator, there's no easily queryable service exposing that. Even if you could get the information on the cluster you wouldn't be able to rely on it since there could be issues with the cluster preventing it from retrieving it.

I also want to add that a Graphana dashboard is not a substitute to this command, while it would be interesting to explore the possibility that could be a topic for a spike or a thread on slack.

joshbranham:
> you wouldn't be able to rely on it since there could be issues with the cluster preventing it from retrieving it.

That is true for this command too.

Makes sense, was just wanting to understand if we could do this more "upstream" but sounds like at this time no.

PR #621, opened 2024-10-29
[OSD-19887]add replace pull secret for HCP clusters

bmeng:
Implement the owner transfer for HCP clusters.

SOP: https://github.com/openshift/ops-sop/blob/master/hypershift/knowledge_base/howto/replace-pull-secret.md


cmd/cluster/transferowner.go
```diff
@@ -83,6 +89,68 @@ func generateServiceLog(clusterId string, template string) servicelog.PostCmdOpt
...
+				if err != nil {
+					return err
+				}
+				secret := &corev1.Secret{}
+				err = json.Unmarshal(jsonData, secret)
+				if err != nil {
+					return err
+				}
+				if secret.Name == secretName {
+					secret.Data[".dockerconfigjson"] = pullsecret
```
devppratik, line 132: The pull secret name also needs an update to ensure that the new secret is populated in nodes. We can maybe generate a random suffix at the end
We also need to be aware of clusters that have different domain name

pkg/utils/ocm.go
```diff
@@ -432,28 +432,56 @@ func GetHiveCluster(clusterId string) (*cmv1.Cluster, error) {
 	return resp.Items().Get(0), nil
 }
 
-// GetManagementCluster returns the OCM Cluster object for a provided clusterId
-func GetManagementCluster(clusterId string) (*cmv1.Cluster, error) {
+// GetMgmtServiceCluster returns the Mgmt Cluster and Service Cluster object for a provided HCP clusterId
+func GetMgmtServiceCluster(clusterId string) (*cmv1.Cluster, *cmv1.Cluster, error) {
```
devppratik, line 436: nit pick: Wondering if `GetManagementAndServiceCluster` would be a better name. It is a bit long but is clear enough

pkg/utils/ocm.go
```diff
@@ -432,28 +432,56 @@ func GetHiveCluster(clusterId string) (*cmv1.Cluster, error) {
...
-		return nil, err
+		return nil, nil, err
+	}
+
+	if err == nil && hypershiftResp != nil {
+		mgmtClusterName = hypershiftResp.Body().ManagementCluster()
+	}
+
+	if mgmtClusterName == "" {
+		return nil, nil, fmt.Errorf("failed to lookup service cluster for cluster %s", clusterId)
```
devppratik, line 459:
```suggestion
		return nil, nil, fmt.Errorf("failed to lookup management cluster for cluster %s", clusterId)
```
Should it not be management cluster here?

pkg/utils/ocm.go
```diff
@@ -488,6 +516,22 @@ func IsManagementCluster(clusterID string) (isMC bool, err error) {
 	return false, nil
 }
 
+func IsHostedCluster(clusterID string) (bool, error) {
```
devppratik, line 519: If we are adding this function, wondering if would make sense to update other places where such checks is performed? 
Other wise it would be a lot of divergent stuff going around

cmd/cluster/transferowner.go
```diff
@@ -305,17 +373,41 @@ func (o *transferOwnerOptions) run() error {
...
+			return fmt.Errorf("cannot check if the given cluster is HCP or not")
+		}
+		if !hcp {
+			return fmt.Errorf("the given cluster is not HCP cluster")
+		}
+		mgmtCluster, svcCluster, err = utils.GetMgmtServiceCluster(o.clusterID)
+		if err != nil {
+			return err
+		}
+		masterCluster = svcCluster
```
devppratik, line 391: The masterCluster naming could be a bit confusing. Wondering if the variable can be renamed into something more fitting both for SC and Hive

cmd/cluster/transferowner.go
```diff
@@ -83,6 +89,68 @@ func generateServiceLog(clusterId string, template string) servicelog.PostCmdOpt
...
+
+	err = kubeCli.Update(context.TODO(), manifestWork, &client.UpdateOptions{})
+	if err != nil {
+		return fmt.Errorf("cannot update the pull-secret within manifestwork: %w", err)
+	}
+
+	// The secret will be synced to the management cluster and guest cluster in a few seconds, wait here
+	time.Sleep(time.Second * 60)
+
+	return nil
```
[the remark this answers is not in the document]
devppratik: Also the reason I add this is because, it could take more than few minutes to actually propagate down all the changes

pkg/utils/ocm.go
samanthajayasinghe: or is it good to add another func to get service cluster by clusterID, 
otherwise, we may call 2 API calls when it's only needed MC cluster

cmd/network/verification.go
```diff
@@ -650,7 +651,7 @@ func (e *EgressVerification) getCaBundleFromManagementCluster(ctx context.Contex
 		return "", err
 	}
 
-	mc, err := utils.GetManagementCluster(e.cluster.ID())
+	mc, _, err := utils.GetMgmtServiceCluster(e.cluster.ID())
```
samanthajayasinghe, line 654: `GetMgmtServiceCluster` internally call 2 API to get MC and SC, but we don't use SC here , is it good to have two func to get MC and SC ?

cmd/cluster/transferowner.go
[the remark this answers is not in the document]
bmeng: There is another function later `verifyClusterPullSecret()` right after we update the manifestwork which is to compare the new pullSecret and the existing secret on the target cluster.
If we do not wait, the pull secret will still be updated, but the verify step will fail mostly.

cmd/cluster/transferowner.go
bmeng: I am not quite understand about this one, I think the secret name is fixed as the `${clustername}-pull` and the secret data key name is fixed as `.dockerconfigjson`
So we just do updating the existing secret key with the new value.

cmd/cluster/transferowner.go
bmeng: yeah, I do not think it is best name, but I am not sure what could be better. Please help suggest if you have any idea.

cmd/network/verification.go
bmeng: ok, I think I can split them into two

cmd/cluster/transferowner.go
```diff
@@ -56,6 +61,7 @@ func newCmdTransferOwner(streams genericclioptions.IOStreams, globalOpts *global
 		},
 	}
 	// can we get cluster-id from some context maybe?
+	transferOwnerCmd.Flags().BoolVar(&ops.hypershift, "hcp", false, "Specify if this is for HCP owner transfer")
```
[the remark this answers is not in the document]
bmeng: Yeah, should be possible, we can call the new `IsHostedCluster` at the beginning and set the bool value.

cmd/cluster/transferowner.go
devppratik: Please refer to Note in step 5 of https://github.com/openshift/ops-sop/blob/master/hypershift/knowledge_base/howto/replace-pull-secret.md#replaceupdate-current-pull-secret

bmeng: ok, thanks, seems there are several more steps to replace the pull secret, I will update the core logic to follow the same

PR #622, opened 2024-10-29
Migrate `gomock` to `go.uber.org/mock/mockgen`

joshbranham:
This migrates our usage of the `gomock` module, which has been [deprecated](https://github.com/golang/mock?tab=readme-ov-file) to the new fork that the community has standardized on https://github.com/uber-go/mock

First commit bumps the package and configures the Makefile, second commit re-generates the mocks and updates package imports across files, third commit ensures `pagerduty` is configured to be generated in the future as well.


typeid:
Good stuff, thanks @joshbranham

PR #623, opened 2024-11-04
OSD-26326: Replace GetHCPNamespace input field to support External ID

TheUndeadKing:
### Summary:
**Issue 1:**
When we pass hcp external Id to fetch logs from dt it throws following error. 

`error: failed to acquire cluster details error retreiving HCP Namespace for given cluster
`

Upon investigation, i found _FetchClusterDetails_  func gets the _clusterKey_ input from the cmdline. The following value passed directly to _ocmutils.GetHCPNamespace_ func. I'd replace to pass the `cluster.ID()` to resolve this panic issue & accept the external id as well.

**Issue 2:**
The dt url generated is in classic format. Which result issue in adding the fliters in the web ui, when we open the url generated from osdctl cmd. 

i'd made changes to generate link in app logs format

### Card [OSD-26326](https://issues.redhat.com/browse/OSD-26326)

### DIFF
--- a/cmd/cluster/context.go
+++ b/cmd/cluster/context.go
@@ -1,7 +1,6 @@
 package cluster
 
 import (
-	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -418,7 +417,10 @@ func (o *contextOptions) generateContextData() (*contextData, []error) {
 			}
 			queryTxt := query.Build()
 			data.DyntraceEnvURL = hcpCluster.DynatraceURL
-			data.DyntraceLogsURL = dynatrace.GetLinkToWebConsole(hcpCluster.DynatraceURL, 10, base64.StdEncoding.EncodeToString([]byte(queryTxt)))
+			data.DyntraceLogsURL, err = dynatrace.GetLinkToWebConsole(hcpCluster.DynatraceURL, 10, queryTxt)
+			if err != nil {
+				errors = append(errors, fmt.Errorf("failed to get url: %v", err))
+			}
 		}
 	}
 
--- a/cmd/cluster/dynatrace/logsCmd.go
+++ b/cmd/cluster/dynatrace/logsCmd.go
@@ -91,7 +91,7 @@ func GetLinkToWebConsole(dtURL string, since int, finalQuery string) (string, er
 	}
 	mStr, err := json.Marshal(SearchQuery)
 	if err != nil {
-		return fmt.Sprintf("failed to create JSON for sharable URL"), err
+		return "", fmt.Errorf("failed to create JSON for sharable URL: %v", err)
 	}
 	return fmt.Sprintf("%sui/apps/dynatrace.logs/?gtf=-%dh&gf=all&sortDirection=desc&advancedQueryMode=true&isDefaultQuery=false&visualizationType=table#%s\n\n", dtURL, since, url.PathEscape(string(mStr))), nil
 }
--- a/cmd/cluster/dynatrace/logsCmd.go
+++ b/cmd/cluster/dynatrace/logsCmd.go
@@ -1,8 +1,9 @@
 package dynatrace
 
 import (
-	"encoding/base64"
+	"encoding/json"
 	"fmt"
+	"net/url"
 
 	"github.com/spf13/cobra"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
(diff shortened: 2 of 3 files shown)


cmd/cluster/dynatrace/logsCmd.go
```diff
@@ -76,7 +77,42 @@ func main(clusterID string) error {
 	}
 
 	fmt.Println(query.Build())
-	fmt.Println("\nLink to Web Console - \n", GetLinkToWebConsole(hcpCluster.DynatraceURL, since, base64.StdEncoding.EncodeToString([]byte(query.finalQuery))))
+	SearchQuery := map[string]interface{}{
```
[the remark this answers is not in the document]
devppratik: Or maybe directly into GetLinkToWebConsole function

cmd/cluster/dynatrace/logsCmd.go
```diff
@@ -76,7 +77,42 @@ func main(clusterID string) error {
...
+				},
+				"showDqlEditor": true,
+			},
+		},
+	}
+	mStr, err := json.Marshal(SearchQuery)
+	if err != nil {
+		return fmt.Errorf("failed to Marshal json %v", err)
+	}
+	fmt.Println("\nLink to Web Console - \n", GetLinkToWebConsole(hcpCluster.DynatraceURL, since, url.PathEscape(string(mStr))))
```
devppratik, line 115: The since variable does not seem to be used in new Logs app when you generate the query and rather is added above in query. Can this be refactored to ensure we donot use it

cmd/cluster/dynatrace/logsCmd.go
```diff
@@ -76,7 +77,42 @@ func main(clusterID string) error {
...
+						"direction": "desc",
+					},
+				},
+				"showDqlEditor": true,
+			},
+		},
+	}
+	mStr, err := json.Marshal(SearchQuery)
+	if err != nil {
+		return fmt.Errorf("failed to Marshal json %v", err)
```
devppratik, line 113: The error message needs to improve to atleast let know user where it actually failed. Maybe something like `failed to create JSON for sharable URL`

cmd/cluster/dynatrace/logsCmd.go
TheUndeadKing: We didn't change the entire URL format. We'd replace base64 encoded with json encoded. Thus the since value is adding to _GetLinkToWebConsole_ func

cmd/cluster/dynatrace/logsCmd.go
```diff
@@ -76,7 +77,42 @@ func main(clusterID string) error {
 	}
 
 	fmt.Println(query.Build())
-	fmt.Println("\nLink to Web Console - \n", GetLinkToWebConsole(hcpCluster.DynatraceURL, since, base64.StdEncoding.EncodeToString([]byte(query.finalQuery))))
+	SearchQuery := map[string]interface{}{
+		"version": 0,
```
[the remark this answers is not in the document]
TheUndeadKing: Can't remove that version part. If we remove it. The DQL query won't get reflected in search bar. i.e, version obj seems to be mandatory

cmd/cluster/dynatrace/logsCmd.go
devppratik: This is the return value for the function - https://github.com/openshift/osdctl/blob/2ebcf8bdf736343d069e6ee043ced5c25d7eb96f/cmd/cluster/dynatrace/logsCmd.go#L60

If you check the current implementation, the since value even if passed in the URL will not be reflect the actual time frame for logs. This is done in the following change of your PR
```
"timeframe": map[string]interface{}{"from": fmt.Sprintf("now()-%vh", since), "to": "now()"},```

This means the value is unused even if we pass it to the function

File: cmd/cluster/dynatrace/logsCmd.go [hunk 5]
```diff
@@ -76,7 +113,14 @@ func main(clusterID string) error {
 	}
 
 	fmt.Println(query.Build())
-	fmt.Println("\nLink to Web Console - \n", GetLinkToWebConsole(hcpCluster.DynatraceURL, since, base64.StdEncoding.EncodeToString([]byte(query.finalQuery))))
+
+	url := GetLinkToWebConsole(hcpCluster.DynatraceURL, since, query.finalQuery)
+
+	if strings.Contains(url, "failed to create JSON for sharable URL") {
+		return fmt.Errorf("failed to get url")
+	}
```
devppratik, line 121: We should properly return errors and bubble them up, instead of just returning a single value

cmd/cluster/context.go
```diff
@@ -418,7 +417,7 @@ func (o *contextOptions) generateContextData() (*contextData, []error) {
 			}
 			queryTxt := query.Build()
 			data.DyntraceEnvURL = hcpCluster.DynatraceURL
-			data.DyntraceLogsURL = dynatrace.GetLinkToWebConsole(hcpCluster.DynatraceURL, 10, base64.StdEncoding.EncodeToString([]byte(queryTxt)))
+			data.DyntraceLogsURL = dynatrace.GetLinkToWebConsole(hcpCluster.DynatraceURL, 10, queryTxt)
```
devppratik, line 420: Can we make the "since" variable either 
1. An optional parameter for the function entirely because we donot need that here to be passed as 10 hrs. We can go with default value OR
2. Move the `10` to a separate variable and then pass it here to give clarity what the paramter us

cmd/cluster/dynatrace/logsCmd.go
TheUndeadKing: Added return error for _GetLinkToWebConsole_ func to resolve above review

cmd/cluster/context.go
TheUndeadKing: The 10 is defaulf value. I've not edited it.
The since variable is present in _logsCmd's_ GetLinkToWebConsole func

cmd/cluster/dynatrace/logsCmd.go
```diff
@@ -56,8 +57,43 @@ func NewCmdLogs() *cobra.Command {
...
+					},
+				},
+				"showDqlEditor": true,
+			},
+		},
+	}
+	mStr, err := json.Marshal(SearchQuery)
+	if err != nil {
+		return fmt.Sprintf("failed to create JSON for sharable URL"), err
+	}
```
devppratik, line 95:
```suggestion
	if err != nil {
		return "", fmt.Errorf("failed to create JSON for sharable URL", err)
	}
```

devppratik (changes requested):

cmd/cluster/dynatrace/logsCmd.go
TheUndeadKing: I'd applied the suggested changes

PR #624, opened 2024-11-06
OSD-23467: Facilitate usage of osdctl dynatrace logs

MateSaary:
[OSD-23467](https://issues.redhat.com//browse/OSD-23467): Facilitate usage of osdctl dynatrace logs

* Add clusterID fetching from current cluster context as default for dynatrace logs command, can still use `--cluster-id` flag to provide clusterID manually.

* Amend `--po`, `--container`, `--namespace`, flags to better align with `oc logs` command flags.

* Add examples to command to help with use.

### DIFF
--- a/cmd/cluster/dynatrace/logsCmd.go
+++ b/cmd/cluster/dynatrace/logsCmd.go
@@ -4,6 +4,7 @@ import (
 	"encoding/base64"
 	"fmt"
 
+	k8s "github.com/openshift/osdctl/pkg/k8s"
 	"github.com/spf13/cobra"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 )
@@ -14,6 +15,7 @@ var (
 	since         int
 	contains      string
 	sortOrder     string
+	clusterID     string
 	namespaceList []string
 	nodeList      []string
 	podList       []string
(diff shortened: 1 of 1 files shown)


cmd/cluster/dynatrace/logsCmd.go
```diff
@@ -14,44 +17,75 @@ var (
...
+				}
+			}
+			err = main(clusterID)
 			if err != nil {
 				cmdutil.CheckErr(err)
 			}
 		},
 	}
 
+	logsCmd.Flags().StringVar(&clusterID, "cluster-id", "", "Cluster ID (default to current cluster context)")
```
devppratik, line 77:
```suggestion
	logsCmd.Flags().StringVar(&clusterID, "cluster", "", "Name or ID of the cluster (defaults to current cluster context)")
```

Looking at other CLI tools like ROSA/OCM, we use `cluster` so would like to keep it similar
```
ocm list idps --cluster=mycluster
```
OR
```
rosa describe cluster --cluster=mycluster

Flags:
  -c, --cluster string             Name or ID of the cluster.
```

cmd/cluster/dynatrace/logsCmd.go
```diff
@@ -144,3 +178,21 @@ func GetQuery(hcpCluster HCPCluster) (query DTQuery, error error) {
...
+		return "", err
+	}
+	re := regexp.MustCompile("/?backplane/cluster/([a-zA-Z0-9]+)/?")
+	matches := re.FindStringSubmatch(parsedURL.Path)
+	if len(matches) < 2 {
+		return "", fmt.Errorf("couldn't find cluster-id from the backplane cluster url")
+	}
+	clusterID := matches[1]
+	return clusterID, nil
+}
```
devppratik, line 198: This is a re-implementation of https://github.com/openshift/osdctl/blob/master/pkg/k8s/client.go#L166
I think we can use the existing function instead of adding a new one which does the same

devppratik (changes requested):

MateSaary:
@devppratik Thank you for the suggestions, all changed accordingly now.

cmd/cluster/dynatrace/logsCmd.go
```diff
@@ -14,44 +15,75 @@ var (
...
+  This command also prints the Dynatrace URL and the corresponding DQL in the output.
+
+`
+
+	logsCmdExample = `
+  # Get the logs of HCP cluster in current cluster context.
+  $ osdctl cluster dynatrace logs
+
+  # Get the logs of a specific HCP cluster
+  $ osdctl cluster dynatrace logs -c <cluster-id>
```
devppratik, line 39: I think the `-c` flag is used for container. This can cause confusion for the usage where -c is valid on different CLI but can refer to either container (oc) or cluster(rosa/ocm)
I would recommend not to include `-c` at all for any flags to avoid any confusion. While testing, I thought the `-c` flag to be cluster ID initially 😓

devppratik:
Thanks for the work @MateSaary 
This is really awesome

PR #630, opened 2024-11-14
OSD-25053: Add basic GCP support for network egress verification

joshbranham:
This adds support for `osdctl network verify-egress` for GCP clusters. With this first pass this only works with BYOVPC clusters as in that scenario OCM has the network information needed. With non-BYOVPC, we need to implement further logic to lookup subnets, vpc and zones. For now, in this scenario, you can provide that data via CLI arguments.

 This works as follows:

- We reach out to OCM to get cluster information
- When GCP, we then attempt to pull all the data from the cluster model
- If we run into issues, inform the user to pass the data manually
- Perform the network verification just like AWS

**Note** Right now, OCM does not return the GCP project ID, although it should, and a bug [exists](https://issues.redhat.com/browse/OCM-12562) so this can go out in the meantime, but once they merge that we can omit `--gcp-project-id` 

I tested this on each type of GCP cluster (BYOVPC and non), and an existing AWS cluster. Also added some basic tests for GCP. Additionally, it is worth noting the existing `verification_test.go` tests are centered around the existing use case which was AWS only, so I left those as is (with a few adjustments for some refactoring.)

I will note that there is more refactoring to be done here to decouple the verification struct building from the CLI's struct, but this is a good first pass imo.

## Sample Run
```
 ./osdctl network verify-egress --cluster-id 2f1dafmj9clha0cilltvq862rgknhi3v -S --gcp-project-id jbranham
2024/11/14 13:23:16 Preparing to check 2 subnet(s) with network verifier.
2024/11/14 13:23:16 running network verifier for subnet  jb-test-gcp-zk789-worker-subnet, security group []
Created instance with ID: verifier-6460
Applying labels
Successfully applied labels 
ComputeService Instance: verifier-6460 RUNNING
Gathering and parsing console log output...
Summary:
printing out failures:
 - egressURL error: https://events.amazonaws.com:443 (Could not resolve host: events.amazonaws.com)

printing out exceptions preventing the verifier from running the specific test:
printing out errors faced during the execution:
INFO[0076] The following clusters match the given parameters: 
jb-test-reuse-vpc   2f1dafmj9clha0cilltvq862rgknhi3v   ready               4.17.3              gcp                 us-east1

INFO[0077] The following template will be sent:         
{
  "severity":"Major",
  "service_name":"SREManualAction",
  "summary":"Action required: Network misconfiguration",
  "description":"Your cluster requires you to take action. SRE has observed that there have been changes made to the network configuration which impacts normal working of the cluster, including lack of network egress to these internet-based resources which are required for the cluster operation and support: https://events.amazonaws.com:443 (Could not resolve host: events.amazonaws.com). Please revert changes, and refer to documentation regarding firewall requirements for PrivateLink clusters: https://access.redhat.com/documentation/en-us/red_hat_openshift_service_on_aws/4/html/prepare_your_environment/rosa-sts-aws-prereqs#osd-aws-privatelink-firewall-prerequisites_rosa-sts-aws-prereqs#.",
  "internal_only":false,
  "event_stream_id":"",
  "doc_references": [
    "https://access.redhat.com/documentation/en-us/red_hat_openshift_service_on_aws/4/html/prepare_your_environment/rosa-sts-aws-prereqs#osd-aws-privatelink-firewall-prerequisites_rosa-sts-aws-prereqs."
  ]
}
Continue? (y/N): N
...
```

### DIFF
--- a/cmd/network/aws.go
+++ b/cmd/network/aws.go
@@ -1,7 +1,6 @@
 package network
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -10,13 +9,9 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/openshift/osd-network-verifier/pkg/data/cloud"
-	"github.com/openshift/osd-network-verifier/pkg/data/cpu"
-	"github.com/openshift/osd-network-verifier/pkg/probes/curl"
-	"github.com/openshift/osd-network-verifier/pkg/probes/legacy"
-	onv "github.com/openshift/osd-network-verifier/pkg/verifier"
+	"github.com/openshift/osd-network-verifier/pkg/verifier"
 	"github.com/openshift/osdctl/pkg/osdCloud"
 	"github.com/openshift/osdctl/pkg/utils"
-	"os"
 	"strings"
 )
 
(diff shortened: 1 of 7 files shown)


cmd/network/aws.go
```diff
@@ -90,42 +80,20 @@ func (e *EgressVerification) setupForAws(ctx context.Context) (*aws.Config, erro
...
-	if e.cluster != nil {
-		if e.cluster.Product().ID() != "rosa" &&
-			e.cluster.Product().ID() != "osd" &&
-			e.cluster.Product().ID() != "osdtrial" {
-			return nil, fmt.Errorf("only supports rosa, osd, and osdtrial, got %s", e.cluster.Product().ID())
-		}
-	}
-
-	input, err := defaultValidateEgressInput(ctx, e.cluster, region)
+func (e *EgressVerification) generateAWSValidateEgressInput(ctx context.Context, platform cloud.Platform) ([]*verifier.ValidateEgressInput, error) {
```
joshbranham, line 86: A lot of the changes here are just moving common bits of the setup that doesn't differ between clouds into shared functions

PR #631, opened 2024-11-15
[WIP][OSD-19887] Keep old registries that are not part of the new pull-secret.

bergmannf:
Probably might want a way to select what to keep - this is mostly to keep the AWS registries as those are not stored in OCM.


bmeng:
@bergmannf please help check if this is good to cover the requirements https://github.com/openshift/osdctl/pull/633

PR #635, opened 2024-11-20
Block running the network verifier for HCP clusters

joshbranham:
We currently do not support running egress verification for HCP clusters. However, this was allowing users to run it anyways

This change blocks the ability to run it, which creates unnecessary noise when troubleshooting.

Example:
```
> ./osdctl network verify-egress --cluster-id d5548186-344b-4811-9078-e1dcbf8b034f -S
2024/11/20 10:58:51 SRE has insufficient AWS permissions to run network verifier on hosted control plane clusters. Exiting
```


fahlmant (approved):
/lgtm
/approve

PR #641, opened 2024-12-06
OSD-27029: Guard against improper usage of the --subnet-id flag

joshbranham:
The `--subnet-id` flag is a `StringArrayVar` which allows you to pass the flag multiple times to construct a slice of values. However, if you pass one value like `--subnet-id foo,bar,baz`, it manifests as strange failures later in the verification process.

This just adds a guard to that footgun and updates the examples to reflect.

```
> ./osdctl network verify-egress -S --subnet-id foo,bar,baz
2024/12/06 09:01:26 network verification failed to validate input: multiple subnets passed to a single --subnet-id flag, you must pass the flag per subnet, eg --subnet-id foo --subnet-id bar
```

### DIFF
--- a/cmd/network/aws.go
+++ b/cmd/network/aws.go
@@ -217,7 +217,6 @@ func (e *EgressVerification) isSubnetPublic(ctx context.Context, subnetID string
 	var routeTable string
 
 	// Try and find a Route Table associated with the given subnet
-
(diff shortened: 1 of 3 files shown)


clcollins (commented):
/lgtm

PR #643, opened 2024-12-13
Default to cpms on controlplane resize

RaphaelBut:
CPMS should be the default case now when this command is used, as clusters < 4.12 are EOL. 
We keep the old way for edge cases, issues with CPMS or any leftover EOL cluster.

Solves https://issues.redhat.com/browse/OSD-27239


joshbranham (commented):
/lgtm

RaphaelBut:
Yes, that would be great :+1:  :)

PR #646, opened 2024-12-20
Require a reason for using 'osdctl cluster ssh key'

joshbranham:
I used this command and received a compliance ticket to fill out. This just plumbs one through via `osdctl cluster ssh key --elevation-reason $reason`. I made this flag required as well so you don't do the harder thing by default.

I am going to update all the SOPs as well that don't reference this command to do so once this is released. Will link that PR here.


typeid (approved):
/lgtm
/approve

Thanks Josh!

PR #653, opened 2025-01-23
Update pullsecret checks to include auth tokens and hive checks

nephomaniac:
This is related to OSD-28215, and OSD-21669. 
This attempts to update the existing pull secret validations to include:
- backplane config/connection validation (prevent working against the wrong cluster)
- Add OCM account email validation for all checked auths
- Add OCM auth token validation for all checked auths
- OCM registry_credentials checks against cluster openshift-config/pull-secret. 
- OCM access_token checks against cluster openshift-config/pull-secret.
- Allow access_token checks to run without impersonation when the tool detects the cluster is owned by the current OCM account/user. 
- Logger, and tabled validation results

./osdctl -S cluster validate-pull-secret -h

	Attempts to validate if a cluster's pull-secret auth values are in sync with the account's email,
	registry_credential, and access token data stored in OCM.
	This requires the caller to be logged into the cluster to be validated.
	If this is being executed against a cluster which is not owned by the current OCM account,
	Region Lead permissions are required to view and validate the OCM AccessToken.

Usage:
  osdctl cluster validate-pull-secret [CLUSTER_ID] [flags]

Examples:

	# Compare OCM Access-Token, OCM Registry-Credentials, and OCM Account Email against cluster's pull secret
	osdctl cluster validate-pull-secret ${CLUSTER_ID} --reason "OSD-XYZ"

	# Exclude Access-Token, and Registry-Credential checks...
	osdctl cluster validate-pull-secret ${CLUSTER_ID} --reason "OSD-XYZ" --skip-access-token --skip-registry-creds

Flags:
  -h, --help                  help for validate-pull-secret
  -l, --log-level string      debug, info, warn, error. (default=info) (default "info")
      --reason string         Mandatory reason for this command to be run (usually includes an OHSS or PD ticket)
      --skip-access-token     Exclude OCM AccessToken checks against cluster secret
      --skip-registry-creds   Exclude OCM Registry Credentials checks against cluster secret

Example run...
```
  ./osdctl -S cluster validate-pull-secret --reason "testing pull secret validations" f88b3fe3-9b02-4a0a-aa22-ABC123ABC123

----------          ----                               ---------        ------      ----  ------
OCM_SOURCE          AUTH                               NAMESPACE        SECRET      ATTR  RESULT
----------          ----                               ---------        ------      ----  ------
account.Email       cloud.openshift.com                openshift-config pull-secret email PASS
account.Email       Redhat_registry.connect.redhat.com openshift-config pull-secret email PASS
registry_credential Redhat_registry.connect.redhat.com openshift-config pull-secret token PASS
account.Email       Redhat_registry.redhat.io          openshift-config pull-secret email PASS
registry_credential Redhat_registry.redhat.io          openshift-config pull-secret token PASS
account.Email       Quay_quay.io                       openshift-config pull-secret email PASS
registry_credential Quay_quay.io                       openshift-config pull-secret token PASS
access_token        cloud.openshift.com                openshift-config pull-secret token PASS
access_token        cloud.openshift.com                openshift-config pull-secret email PASS
access_token        quay.io                            openshift-config pull-secret token PASS
access_token        quay.io                            openshift-config pull-secret email PASS
access_token        registry.connect.redhat.com        openshift-config pull-secret token PASS
access_token        registry.connect.redhat.com        openshift-config pull-secret email PASS
access_token        registry.redhat.io                 openshift-config pull-secret token PASS
access_token        registry.redhat.io                 openshift-config pull-secret email PASS
```


cmd/cluster/validatepullsecret.go
```diff
@@ -48,153 +133,855 @@ func newValidatePullSecretOptions() *validatePullSecretOptions {
...
 	if err != nil {
 		return err
 	}
-	if done {
-		return nil
+	// account email to be for auth comparison
+	emailOCM := o.account.Email()
+	o.log.Info(o.ctx, "Found email for cluster's OCM account: %s\n", emailOCM)
+
+	if !o.hiveOnly {
```
zmird-r, line 244: The logic for flags is pretty unintuitive, is there a way to better structure the code so that it's more clear what operations are performed based on the flag combinations?

cmd/cluster/validatepullsecret.go
```diff
@@ -48,153 +133,855 @@ func newValidatePullSecretOptions() *validatePullSecretOptions {
...
+		// get the pull secret in cluster
+		err = getPullSecretElevated(o.clusterID, o.reason, pullSecret)
```
zmird-r, line 246: In the current validate-pull-secret-email command we have an integration with managed-scripts to retrieve the email without admin elevation, is there a reason we should discard that and return to generate Compliance Alerts?

cmd/cluster/validatepullsecret.go
```diff
@@ -48,153 +133,855 @@ func newValidatePullSecretOptions() *validatePullSecretOptions {
...
+		}
+	}
+
+	// If user chose to use the OCM AccessToken in validations...
+	if o.useAccessToken {
+		userName := o.account.Username()
+		if len(userName) <= 0 {
+			err = fmt.Errorf("found empty 'username' for account:'%s', needed for accessToken", o.account.HREF())
+		} else {
+			accessToken, err = o.getAccessTokenFromOCM(userName)
```
zmird-r, line 289: This check requires Region Lead permission to pull the secret from OCM, I don't see this stated anywhere in a message to the cli user or in the help command. Ideally it should require user confirmation if the flag is active e.g.:
"Retrieving credentials from OCM require Region Lead permissions, continue? [y,N]

cmd/cluster/validatepullsecret.go
```diff
@@ -48,153 +133,855 @@ func newValidatePullSecretOptions() *validatePullSecretOptions {
...
+		if err != nil {
+			fmt.Printf("Error validating AccessToken:%s'.\nWould you like to continue with validations? ", err)
+			if !utils.ConfirmPrompt() {
+				return err
+			}
+		}
+	}
+
+	/* Checks against hive to confirm hive is sync'd with OCM */
+	if o.checkHive {
```
zmird-r, line 315: I don't see why we need to check the credentials stored in Hive.
We don't have a formal process to update the pull secret for managed openshift, once the credentials are stored in Hive after cluster creation, AFIK there's no other process that changes them and would cause them to get out of sync (except owner transfer but those are manual processes with their own tooling).

Updates to PS are not supported in managed openshift and need to be explicitly required via support ticket, so I don't see why complicate the code with this check for a situation we can already detect from the PS on the cluster. It seems redundant.

zmird-r: This also generates an additional Compliance Ticket

zmird-r:
Wouldn't it be easier to rename the command we currently have to: validate-pull-secret-email and create a new command named validate-pull-secret-credentials? The two checks even if related have different steps, it would simplify the code.

joshbranham:
Filed https://issues.redhat.com/browse/OSD-28215 for better tracking of the extended validation

cmd/cluster/validatepullsecret.go
nephomaniac: It appears this is the same behavior as the existing osdctl command, I believe it always runs as Backplane Cluster Admin when fetching the pull secret on the target cluster?

cmd/cluster/validatepullsecret.go
nephomaniac: For test purposes, the command 'may' not always require RL perms. If the cluster is owned by the current OCM users executing this request, the command will not 'impersonate'. 
If the command fails for perms issues this error is provided:
https://github.com/openshift/osdctl/blob/c6d6b0c49dedb250b2c01b0eb999dc5bd8ea967c/cmd/cluster/validatepullsecret.go#L847
and then the user is prompted if they'd like to continue:
https://github.com/openshift/osdctl/blob/c6d6b0c49dedb250b2c01b0eb999dc5bd8ea967c/cmd/cluster/validatepullsecret.go#L293

We could also add a prompt during the prerun()  when the cli args indicate the user will want to fetch the AccessToken?

cmd/cluster/validatepullsecret.go
nephomaniac: This part was added to help a user validate the hive data + syncsets. The hive checks do not run by default, and can be run independent of the target cluster checks. This would only need to be checked if there the checks on the target cluster fail. In that case the comp ticket is likely justified? 
This is currently waiting for feedback on the actual update secret process and tooling.  If this is not a touch point then validating this would be mis-leading.

cmd/cluster/validatepullsecret.go
zmird-r: We did have an integration with managed-scripts in the past https://github.com/openshift/osdctl/pull/488 but it was removed due to a bug in the implementation, I was wrongly under the impression that it was still there. I think we can revisit this idea with a new implementation, I opened a ticket for it https://issues.redhat.com/browse/OSD-28292

cmd/cluster/validatepullsecret.go
zmird-r: We could be more upfront and add it to the description and during command runtime. The message could also be less ambiguous:

> AccessToken ops may require 'region lead' permissions to execute.

This leaves doubts about when it is required, the user needs to find this information elsewhere.

I suggest something like:

"Fetching the AccessToken from OCM requires Region Lead permissions unless you're the cluster owner. Would you like to continue?"

cmd/cluster/validatepullsecret.go
zmird-r: The Hive team confirmed here https://redhat-internal.slack.com/archives/CE3ETN3J8/p1739453387930049 that the secret is not used in any capacity after provisioning. 

This validation is indeed misleading, it raises doubt on 
- The role of that secret
- Follow-up action if it is found not up to date: should it be updated and how?
- When this validation should be run and as part of which investigation

I think we should remove this check but if you have more doubts, could you please open a thread on the sre channel so we can gather more feedback from the team?

cmd/cluster/validatepullsecret.go
nephomaniac: Flags have been updated, removing most of the previous flags and replacing with 2 flags providing the option to skip either or both OCM 'registry credential' and/or 'access token' comparisons.

cmd/cluster/validatepullsecret.go
nephomaniac: Hive checks have been removed so should now behave the same as the existing/previous script in regards to compliance alerts.

cmd/cluster/validatepullsecret.go
nephomaniac: Information about region lead permissions has been added to the cmd help text. The logged error message to the user was moved and formatted to make it stand out with a prompt for them to proceed or exit.

cmd/cluster/validatepullsecret.go
nephomaniac: Hive checks have been removed.  Worth noting we still use Hive to update the pull secret when transferring cluster ownership, but tracking the (random) secret name used for this op is likely beyond the scope of this validation util.

nephomaniac:
> Wouldn't it be easier to rename the command we currently have to: validate-pull-secret-email and create a new command named validate-pull-secret-credentials? The two checks even if related have different steps, it would simplify the code.

Happy to separate this out into a 2nd cmd/util if this is a better approach?

nephomaniac:
Closing this pull request per new branch addressing OSD-28215 to replace this PR. See https://github.com/openshift/osdctl/pull/702

PR #658, opened 2025-02-03
SDCICD-1418 add cmd option to promote operator saas through canary hives

ritmun:
Adds canary promotion option to osd saas promotions. 

```	
./osdctl promote saas --appInterfaceDir <appiface repo path>  --serviceName <operator svc name>  --gitHash  <desired git> --osd  
 ```

- if saas file is set up for canary targets, canary promotion will be performed. Otherwise all production hives will be promoted per usual. 
- log message has additional messages showing which targets are updated ie:
```
Service: saas-osd-example-operator will be promoted to d589fe229b49f1160560203be069931d0462b937
updating target:  osd-example-operator-prod-canary
updating target:  osd-example-operatorother-99088-prod-canary
commitMessage: Promote saas-osd-example-operator to d589fe229b49f1160560203be069931d0462b937
```
- Followup planned for saas files to mark some production hives as canary (example [SFO](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/132092))
  

Example:

- input saas file with 2 canary hives and one post-canary hive:  https://go.dev/play/p/H5ArwZ2exQD.go?download=true 
- promoted canary from `96acff2470fdf0d8f514463f00ef751bb2833fd0` to `d589fe229b49f1160560203be069931d0462b937`
- output saas file: https://go.dev/play/p/YWfnBNIcbs3.go?download=true

### CODE REVIEW

File: cmd/promote/saas/saas.go [hunk 1]
```diff
@@ -72,6 +73,7 @@ func NewCmdSaas() *cobra.Command {
 	saasCmd.Flags().StringVarP(&ops.namespaceRef, "namespaceRef", "n", "", "SaaS target namespace reference name")
 	saasCmd.Flags().BoolVarP(&ops.osd, "osd", "", false, "OSD service/operator getting promoted")
 	saasCmd.Flags().BoolVarP(&ops.hcp, "hcp", "", false, "HCP service/operator getting promoted")
+	saasCmd.Flags().StringVarP(&ops.canary, "canary", "", "auto", "Select to promote service to canary production hives (yes/no/auto). By default, auto will deploy to canary only if target names containing -prod-canary- are set up in saas file. Otherwise will deploy to all production targets.")
```
iamkirkbater, line 76: Is there a specific reason to use `yes|no|auto` here as a `StringVarP`, versus a `BoolVarP` with a default to `true`?

I believe that a `BoolVarP` with default to true would give us the same experience, and be a more consistent user experience, so it would just end up being a `--canary` flag versus `--canary no`

Actually - as I started thinking through this in my head after typing I started thinking about how this would work 🤔 and I don't think `BoolVarP` would make sense either, as if the default behavior would be no-flag and that would target canary but if we want to NOT target canary we'd end up _having_ to pass the `--canary` flag...

---

What if instead of calling this `canary` as the flag name we call this `target` or `sector`? Then we could modify the accepted targets in the future as we extend this command.

Examples:

Would default to canary:
```
osdctl promote saas  --serviceName myService --osd
``` 

Would explicitly target canary:
```
osdctl promote saas  --serviceName myService --osd --target canary
```

Would also be the same set of flags for hcp sector targeting in the future:
```
osdctl promote saas  --serviceName myService --hcp --target sector2
```

Could be further extended to have "all" as a target to bring everything to the same target (in this case `prod-canary-*` and `hivep-*`:
```
osdctl promote saas  --serviceName myService --osd --target all
``` 

Additionally let's say we have 3 git shas in various environments:
```
aaabbb - current prod version
bbbccc - previously tested but new canary was promoted before this rolled out to the rest of prod
cccddd - current canary version
```

We would just name our current `hivep-*` as "main" or "prod" and target that specifically:
```
osdctl promote saas  --serviceName myService --osd --target main --gitHash bbbccc
```

It would require a bit of rework here to support this specifically, but I personally think it would be worth the effort to make this flag a bit more robust and extendable for the future.

cmd/promote/saas/utils.go
```diff
@@ -73,6 +73,7 @@ func servicePromotion(appInterface git.AppInterface, serviceName, gitHash string
 	}
 	fmt.Printf("Service: %s will be promoted to %s\n", serviceName, promotionGitHash)
 
+	// fixme unassigned err
```
joshbranham, line 76: Nit: check this

joshbranham (commented):
/lgtm

PR #664, opened 2025-02-13
SDCICD-1418 improve resourceTemplates looping for promotion, remove dependency on target name if canary not set

ritmun:
[SDCICD-1418](https://issues.redhat.com//browse/SDCICD-1418) improve resourceTemplates looping for promotion, remove dependency on target name if canary not set

1. instead of looking for a specific yaml path in resource template, loop through all of them. This is important when a saas file does not have the standard operator yaml path eg MCC, MCVW
2. If saas file does not have canary targets set, continue to do a full replace of old sha with new.

### DIFF
--- a/cmd/promote/git/app_interface.go
+++ b/cmd/promote/git/app_interface.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 
 	kyaml "sigs.k8s.io/kustomize/kyaml/yaml"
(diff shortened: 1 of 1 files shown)


cmd/promote/git/app_interface.go
```diff
@@ -46,24 +46,29 @@ func replaceTargetSha(fileContent string, targetSuffix string, promotionGitHash
...
+	for i := range len(rts.Content()) {
+		targets, err := kyaml.Lookup("resourceTemplates", strconv.Itoa(i), "targets").Filter(node)
+		if err != nil {
+			return "", fmt.Errorf("error querying saas YAML: %v", err), false
 		}
-		return nil
-	})
-
+		err = targets.VisitElements(func(element *kyaml.RNode) error {
+			name, _ := element.GetString("name")
```
clcollins, line 59: Naive question (and I realize this isn't your code and existed in the previous version) - is there a good reason to ignore the error here? 

It seems like it was deliberate because the regexp.MatchString also ignores the error, which would report of the pattern was invalid.  So, maybe there's not _always_ a name?  :thinking:

joshbranham (commented):
/lgtm

PR #669, opened 2025-02-19
OSD-19395 - Adding rebuild-ovn subcommand

mcipamo:
https://issues.redhat.com/browse/OSD-19395

---
Adding subcommand to rebuild OVN for 4.14+ clusters
This is the same process described here: https://access.redhat.com/articles/6963671#rebuild_ovn_on_ocp_4_8_4_13

### DIFF
--- a/cmd/network/rebuild-ovn.go
+++ b/cmd/network/rebuild-ovn.go
@@ -1,0 +1,253 @@
+package network
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/openshift/osdctl/pkg/k8s"
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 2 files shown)


cmd/network/rebuild-ovn.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	for _, node := range nodes {
+		log.Printf("Processing node: %s", node)
+		if err := executeCommandOnNode(kubeClient, node, commands); err != nil {
+			log.Printf("Error executing commands on node %s: %v", node, err)
+		}
+
+		// Add a sleep time before deleting the ovnkube-node pod
+		time.Sleep(5 * time.Second) // Wait for 5 seconds
```
tnierman, line 82: Do we need to sleep, or can we wait for a condition to be set on the pod?

joshbranham (commented):
This feels like something that shouldn't be one-off automated by SRE. Is there an upstream bug for OCP that tracks the scenarios where we need to rebuild?

Also, how often does this actually happen? I noticed the Jira is from 2023.

Lastly, the steps for 4.14+ are here https://access.redhat.com/articles/6963671#rebuild_ovn_on_ocp_4_14_later and looks like we might have automated for < 4.14 which feels not relevant to the fleet SRE manages.

devppratik:
Closing due to inactivity

PR #671, opened 2025-02-24
OSD-24997: Show if a user is banned when using osdctl cluster context

Mhodesty:
This PR adds output to osdctl cluster context to show if a user is banned. The goal is to cut down on SRE toil time by showing this information at the beginning of the troubleshooting process. Output looks like this

>> User Ban Details
User is banned
Ban code = pending
Ban description = pending

### DIFF
--- a/cmd/cluster/context.go
+++ b/cmd/cluster/context.go
@@ -95,6 +95,11 @@ type contextData struct {
 
 	// OCM Cluster description
 	Description string
+
+	// User Banned Information
+	UserBanned     bool
+	BanCode        string
+	BanDescription string
 }
 
 // newCmdContext implements the context command to show the current context of a cluster
@@ -241,6 +246,11 @@ func (o *contextOptions) printLongOutput(data *contextData) {
 
 	// Print Dynatrace URL
 	printDynatraceResources(data)
+
+	// Print User Banned Details
+	if data.UserBanned {
+		printUserBannedStatus(data)
+	}
 }
 
 func (o *contextOptions) printShortOutput(data *contextData) {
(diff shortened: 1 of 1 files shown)


cmd/cluster/context.go
```diff
@@ -734,6 +759,21 @@ func printDynatraceResources(data *contextData) {
 	}
 }
 
+func printUserBannedStatus(data *contextData) {
+	var name string = "User Ban Details"
+	fmt.Println("\n" + delimiter + name)
```
joshbranham, line 764: Are we always using `fmt.Println` already above? If so this is adding an extra `\n` between the last printed text and this.

cmd/cluster/context.go
```diff
@@ -378,6 +386,22 @@ func (o *contextOptions) generateContextData() (*contextData, []error) {
 		}
 	}
 
+	GetBannedUser := func() {
+		defer wg.Done()
```
joshbranham, line 390: This is never used, was going to suggest removing it but I can open a follow up to clean all the waitgroup usage up.

joshbranham (commented):
Can you post some terminal output of this running with the new output, mostly because this command has zero unit tests and would be good to see it work if we aren't going to add test coverage

cmd/cluster/context.go
Mhodesty: Hi JB. 
https://redhat-internal.slack.com/archives/CCX9DB894/p1739809796051329
Here's a thread with the output on a cluster where a user is banned.  See Jim's comment from 12:01 Saturday. I've also pasted the output as it shows in the command in the code block below.
This adds an extra \n between "User Ban Details" and "User is banned" or "User is not banned" increasing readability.  

```
>> User Ban Details
User is banned
Ban code = pending
Ban description = pending
```

joshbranham: I more meant the full command output of `osdctl cluster context $CLUSTER_ID` to see if there is an extra newline

Mhodesty: Sure. Here's the output of the osdctl command run on a cluster. I redacted the cluster identifying information but the output looks clean to me.

```
>> PagerDuty Alerts
Service: [Redacted]
None

>> External resources
CCX dashboard             [Redacted]
OHSS Cards                  [Redacted]
PagerDuty Service        [Redacted]
Splunk Audit Logs         [Redacted]

>> Dynatrace Details
Not an HCP or MC Cluster

>> User Ban Details
User is not banned
```

joshbranham (commented):
/lgtm

PR #672, opened 2025-02-24
OSD-24628 - added external id to osdctl org clusters

anispate:
[OSD-24628](https://issues.redhat.com//browse/OSD-24628) - osdctl org clusters is returning the external ids


cmd/org/clusters.go
```diff
@@ -138,6 +138,7 @@ func printClusters(items []*accountsv1.Subscription) {
 		for _, item := range items {
 			subscription := map[string]string{
 				"cluster_id":   item.ClusterID(),
+				"external_id":  item.ExternalClusterID(), //external ID
```
joshbranham, line 141: The function name is well defined and tells us what it is, therefore we don't need this comment. Same applies to line 156

PR #674, opened 2025-03-03
Update README Documentation as part of every release - OSD-19018

ratnam915:
As a part of story :

 [OSD-19018](https://issues.redhat.com//browse/OSD-19018),the requirement was to Update README Documentation as a part of every release.

Below are the changes as part of this PR :

    Added a new package dockgen.go, this generates documentation based on all the commands available in osdctl
    The documentation created will be independent of the README file and will be available on the ROOT folder as osdctl_commands.md.
    ./docs folder added with .md documents for all the osdctl commands.
    Since we want the documentation to update automatically as part of every release, a make file entry for creating documents added as part of make release command.


Makefile
```diff
@@ -39,6 +45,7 @@ build:
 	goreleaser build --clean --snapshot --single-target=${SINGLE_TARGET}
 
 release:
+	make generate-docs
 	goreleaser release --clean
```
joshbranham, line 49: So this is going to generate the docs every time a team lead runs a release locally, which will

- Generate them if they are missing, creating a dirty checkout
- Possibly cause `goreleaser` to fail to do the release since the checkout isn't clean

I think we want this to be something that is enforced for all PRs. The flow could look something like:

- I make a branch, change a command to add a new flag, the docs need to be updated
- I push my changes and a prow test runs `make generate-docs` and checks if the checkout is clean (or this is added as a new make target
- The test fails, and outputs that you need to update the docs

Bonus points if we add a git hook (pre-commit or just git hooks, no strong preference) and steps to the README on enabling them so users can be alerted about the docs needing to be generated before pushing.

ratnam915: Hey @joshbranham:

Thank you for your feedback, the suggestions you've posted do like they would be a cleaner solution to this.

I'll give it another go and re-submit the PR.

Thanks

PR #675, opened 2025-03-04
OSD-24355 : Updated mc list command to respect output flag

ratnam915:
Bug Fix : OSD - 24355 : Below are the changes as a part of the PR:

1. Initially the command was not respecting the output flag and was giving the output as table format irrespective of the flag.
2. Enabled the command with the global "output" flag.
3. The command now supports text,json,yaml formats with the output flag

### DIFF
--- a/cmd/mc/list.go
+++ b/cmd/mc/list.go
@@ -1,6 +1,7 @@
 package mc
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"os"
@@ -9,14 +10,26 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws/arn"
 	"github.com/openshift/osdctl/pkg/utils"
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
 )
 
 type list struct {
+	// Add a reference to the output format
+	outputFormat string
+}
+
+// Define a struct for output formatting
+type managementClusterOutput struct {
+	Name      string `json:"name" yaml:"name"`
+	ID        string `json:"id" yaml:"id"`
+	Sector    string `json:"sector" yaml:"sector"`
+	Region    string `json:"region" yaml:"region"`
+	AccountID string `json:"account_id" yaml:"account_id"`
+	Status    string `json:"status" yaml:"status"`
 }
 
 func newCmdList() *cobra.Command {
 	l := &list{}
-
 	listCmd := &cobra.Command{
 		Use:     "list",
 		Short:   "List ROSA HCP Management Clusters",
@@ -24,6 +37,8 @@ func newCmdList() *cobra.Command {
 		Example: "osdctl mc list",
 		Args:    cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			// Get the global output flag value
+			l.outputFormat = cmd.Flag("output").Value.String()
 			return l.Run()
 		},
 	}
@@ -43,8 +58,9 @@ func (l *list) Run() error {
 		return fmt.Errorf("failed to list management clusters: %v", err)
 	}
 
-	w := tabwriter.NewWriter(os.Stdout, 1, 1, 1, ' ', 0)
-	fmt.Fprintln(w, "NAME\tID\tSECTOR\tREGION\tACCOUNT_ID\tSTATUS")
+	// Prepare a slice to hold the structured output
+	output := []managementClusterOutput{}
+
 	for _, mc := range managementClusters.Items().Slice() {
 		cluster, err := ocm.ClustersMgmt().V1().Clusters().Cluster(mc.ClusterManagementReference().ClusterId()).Get().Send()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/mc/list.go
```diff
@@ -58,18 +74,66 @@ func (l *list) Run() error {
...
+		yamlOutput, err := yaml.Marshal(output)
+		if err != nil {
+			return fmt.Errorf("failed to format YAML output: %v", err)
+		}
+		fmt.Println(string(yamlOutput))
+	case "text":
+		// Plain text output format, one cluster per section with labeled fields
+		for i, item := range output {
+			if i > 0 {
+				fmt.Println() // Add blank line between clusters
```
btoll, line 111: Can this be addressed another way, such as adding a double \n\n to the last write?  If there is only one item, it probably isn't bad if it's still outputting an additional empty line before returning the prompt.

cmd/mc/list.go
```diff
@@ -58,18 +74,66 @@ func (l *list) Run() error {
...
+			}
+			fmt.Printf("Management Cluster #%d:\n", i+1)
+			fmt.Printf("  Name:       %s\n", item.Name)
+			fmt.Printf("  ID:         %s\n", item.ID)
+			fmt.Printf("  Sector:     %s\n", item.Sector)
+			fmt.Printf("  Region:     %s\n", item.Region)
+			fmt.Printf("  Account ID: %s\n", item.AccountID)
+			fmt.Printf("  Status:     %s\n", item.Status)
```
btoll, line 119: Why not use a writer here?

cmd/mc/list.go
```diff
@@ -9,21 +10,35 @@ import (
...
 
 func newCmdList() *cobra.Command {
 	l := &list{}
-
 	listCmd := &cobra.Command{
 		Use:     "list",
 		Short:   "List ROSA HCP Management Clusters",
 		Long:    "List ROSA HCP Management Clusters",
 		Example: "osdctl mc list",
 		Args:    cobra.NoArgs,
```
btoll, line 38: Should this be changed since this is accepting a parameter?  Also, how will the user know that they now have the option of choosing an output format?

cmd/mc/list.go
```diff
@@ -58,18 +74,66 @@ func (l *list) Run() error {
 			supportRoleARN, err := arn.Parse(supportRole)
 			if err != nil {
 				log.Printf("failed to convert %s to an ARN: %v", supportRole, err)
+			} else {
```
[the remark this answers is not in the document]
ratnam915: Hi Christopher, 

The else block is now removed.

cmd/mc/list.go
```diff
@@ -9,21 +10,35 @@ import (
...
 )
 
 type list struct {
+	// Add a reference to the output format
+	outputFormat string
+}
+
+// Define a struct for output formatting
+type managementClusterOutput struct {
+	Name      string `json:"name" yaml:"name"`
```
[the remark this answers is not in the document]
ratnam915: Made the changes to keep the fields the same.

cmd/mc/list.go
```diff
@@ -9,21 +10,35 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws/arn"
 	"github.com/openshift/osdctl/pkg/utils"
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
```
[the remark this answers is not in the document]
ratnam915: Upgraded to v3 as suggested

cmd/mc/list.go
```diff
@@ -9,21 +10,35 @@ import (
...
-
 	listCmd := &cobra.Command{
 		Use:     "list",
 		Short:   "List ROSA HCP Management Clusters",
 		Long:    "List ROSA HCP Management Clusters",
 		Example: "osdctl mc list",
 		Args:    cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			// Get the global output flag value
+			l.outputFormat = cmd.Flag("output").Value.String()
```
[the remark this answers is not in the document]
ratnam915: Hey Ben,

Currently the output flag takes four accepted values ' ' , yaml,json,text.

In this particular scenario i think it is ok for the command to display the default value since originally the command was displaying table format no matter what the value was in the output flag parameter.

cmd/mc/list.go
```diff
@@ -58,18 +74,66 @@ func (l *list) Run() error {
 			supportRoleARN, err := arn.Parse(supportRole)
 			if err != nil {
 				log.Printf("failed to convert %s to an ARN: %v", supportRole, err)
```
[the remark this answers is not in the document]
ratnam915: Hey Ben,

This is more of a check for logging and monitoring puposes, if one of the ARN's is not in the correct format it moves on to list the other management clusters.

cmd/mc/list.go
ratnam915: This has been addressed now.

cmd/mc/list.go
ratnam915: Code changed to use writer instead

cmd/mc/list.go
ratnam915: Hey Ben,

Further description has been added, also since '  ' is acceptable as a valid argument for the output flag it will still display in a table format

cmd/mc/list.go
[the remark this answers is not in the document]
ratnam915: Hey Ben,

This check is mostly for logging and monitoring purposes, hence the command will still contnue parsing and listing out the other management clusters

cmd/mc/list.go
```diff
@@ -59,17 +72,66 @@ func (l *list) Run() error {
...
 			}
+
 			awsAccountID = supportRoleARN.AccountID
```
btoll, line 76: Do you know what the value of `supportRoleARN` is, though, if `err` is not `nil` above?  This could then be trying to do a lookup on a null pointer.

cmd/mc/list.go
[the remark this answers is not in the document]
btoll: Sorry, I think I could have been clearer.  What I mean is that shouldn't the fields be non-exportable, too?  Something like this:
```
type managementClusterOutput struct {
	name      string `json:"name" yaml:"name"`
	id        string `json:"id" yaml:"id"`
	sector    string `json:"sector" yaml:"sector"`
	region    string `json:"region" yaml:"region"`
	accountID string `json:"account_id" yaml:"account_id"`
	status    string `json:"status" yaml:"status"`
}
```

cmd/mc/list.go
btoll: I was thinking something along the lines of:
```func newCmdList() *cobra.Command {
	l := &list{}

	listCmd := &cobra.Command{
		Use:     "list",
		Short:   "List ROSA HCP Management Clusters.",
		Long:    "List ROSA HCP Management Clusters.",
		Example: "osdctl mc list --output json",
		Args:    cobra.NoArgs,
		RunE: func(cmd *cobra.Command, args []string) error {
            return l.Run()
		},
	}

    flagSet := listCmd.Flags()
    flagSet.StringVar(
        &l.outputFormat,
        "output",
        "",
        "Output format. Supported output formats include: text, json, yaml",
    )

	return listCmd
}
```
This is untested.

btoll: Yes, but those global flags don't match what is needed for this command.  It doesn't have to be exactly as that snippet above, but it should reflect what the command expects.

cmd/mc/list.go
ratnam915: The above statement would be triggered if arn.Parse(supportRole) fails, this would fail in the below scenarios:

1. Invalid ARN Format: If the supportRole string does not conform to the Amazon Resource Name (ARN) format.
2. Incomplete ARN: If the supportRole string is missing critical components of an ARN.

The supportRole value would likely be an empty string or a malformed ARN when this occurs.

cmd/mc/list.go
btoll: It's not a duplicate flag.  If it makes sense to update the global flag to add `text` b/c every command supports it then fine, but then there is still an `env` format in the global flags that this command doesn't appear to support.

My understanding of `cobra` is that it needs to be "told" about parameters/flags/values that a command supports so it can then present it in its help menu (and possibly tab completion?).  I also would think that it would be a fairly common pattern to define something globally and then override at a more granular level, as this code snippet is doing.

Again, it doesn't have to be that code snippet, but I think it's important that we register the flag values with `cobra`.

We can let the upstream team determine what they want, I just wanted to flag it as a possible issue (no pun intended :) ).

cmd/mc/list.go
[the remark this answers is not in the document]
btoll: Yes, you're right.

cmd/mc/list.go
```diff
@@ -59,17 +80,66 @@ func (l *list) Run() error {
...
+			fmt.Fprintf(w, " Account ID:\t%s\n", item.AccountID)
+			fmt.Fprintf(w, " Status:\t%s\n", item.Status)
+
+			if i < len(output)-1 {
+				fmt.Fprintln(w, "")
+			}
+			w.Flush()
+		}
+	default:
+
```
joshbranham, line 128:
```suggestion
```

cmd/mc/list.go
```diff
@@ -59,17 +80,66 @@ func (l *list) Run() error {
...
+			fmt.Fprintf(w, " Region:\t%s\n", item.Region)
+			fmt.Fprintf(w, " Account ID:\t%s\n", item.AccountID)
+			fmt.Fprintf(w, " Status:\t%s\n", item.Status)
+
+			if i < len(output)-1 {
+				fmt.Fprintln(w, "")
+			}
+			w.Flush()
+		}
+	default:
```
joshbranham, line 127: So is default intended to be `table`? If so, I would update the flag to have a default value of `table`, then add it to the switch here, and then make the `default` case in the switch throw an error saying something like "unsupported output option, must be one of: x,y,z"

joshbranham (commented):
Couple last comments

cmd/mc/list.go
ratnam915: Hey @joshbranham,

The initial requirement for this change was that the osdctl mc list command was not repecting the --output flag, i.e whatever value we insert it was displaying the output in a table format, hence we have not kept an unsupported format section, if it is not explicitly mentioned as json,yaml or text it will default to table like it always had.

However if you'd want the change to add table explicitly i can go ahead and make the changes

joshbranham: Yes thinking of the user experience, magically defaulting to table without having that be an option in the flag is confusing.

ratnam915: Hey @joshbranham : I agree, i've made the change and pushed the code, please let me know your thoughts on the same

joshbranham (commented):
Thanks for making the changes! /lgtm

PR #677, opened 2025-03-05
implemented test cases for network package

anandpulakala:
added test cases for network package

JIRA Ticket:
https://issues.redhat.com/browse/OSD-28425


cmd/network/cmd_test.go
```diff
@@ -0,0 +1,71 @@
...
+			subCmd := cmd.Commands()[0]
+			gomega.Expect(subCmd.Use).To(gomega.Equal("packet-capture"))
+		})
+
+		ginkgo.It("should execute the network command without errors", func() {
+			cmd.SetArgs([]string{})
+			err := cmd.Execute()
+			gomega.Expect(err).To(gomega.BeNil())
+		})
+	})
```
joshbranham, line 47: Can you share a ticket for this work? I am curious the motivation for this work (more testing is good!).

However, I would question the benefit of testing things like cobra command setup. You are essentially testing that it compiles, which is already validated. I would also prefer to see basic testing done without using ginkgo, but I realize a lot of our projects have already pulled it in for some cases.

cmd/network/verification_test.go
```diff
@@ -3,19 +3,26 @@ package network
...
 	"fmt"
+	"github.com/openshift-online/ocm-sdk-go/logging"
+	"testing"
+	"time"
+
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+
+	"reflect"
```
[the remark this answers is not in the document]
btoll: Yeah, using tooling like `goimports` will help with that.

cmd/network/verification_test.go
```diff
@@ -39,6 +46,279 @@ func newTestCluster(t *testing.T, cb *cmv1.ClusterBuilder) *cmv1.Cluster {
...
+			output:    &output.Output{}, // Empty output for no failures
+			clusterId: "test-cluster",
+			want:      servicelog.PostCmdOptions{},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			got := generateServiceLog(tc.output, tc.clusterId)
+			if !reflect.DeepEqual(got, tc.want) {
```
btoll, line 252: I was recently looking into comparing objects and the `reflect` package and came across this that you may find interesting:
https://github.com/google/go-cmp

cmd/network/cmd_test.go
anandpulakala: I referred to backplane-cli while writing test cases for this.

For cmd.go, I used the following references:

[ocm-backplane ]  (https://github.com/openshift/backplane-cli/tree/main/cmd/ocm-backplane)

[root_test.go] (https://github.com/openshift/backplane-cli/blob/main/cmd/ocm-backplane/root_test.go)
[root_suite_test.go] (https://github.com/openshift/backplane-cli/blob/main/cmd/ocm-backplane/root_suite_test.go)
Related Jira ticket: [OSD-28408](https://issues.redhat.com/browse/OSD-28408).

btoll: I agree with @joshbranham, we don't want to be testing the prerogatives of the `cobra` package, which I believe these tests are doing.  We need to trust that the `cobra` team has already done their due diligence, leaving us free to test what we've done.

cmd/network/verification_test.go
btoll: In my view, we should be using `go-cmp` instead of `reflect.DeepEqual`.  There's good information in that link I provided, and the package is by the Go team.

cmd/network/verification_test.go
```diff
@@ -39,6 +46,279 @@ func newTestCluster(t *testing.T, cb *cmv1.ClusterBuilder) *cmv1.Cluster {
...
+	}{
+		{
+			name: "no cluster",
+			ev: &EgressVerification{
+				cluster: nil,
+			},
+			want:    "",
+			wantErr: false,
+		},
+		// Add more test cases for different scenarios
```
btoll, line 275: Is this something that can be done now?  :)

cmd/network/cmd_test.go
joshbranham: I will review this PR but I still would ask you remove testing for basic cobra behavior here. As far as "meeting 50% coverage" I personally will not hold your PR up for that, assuming we have added tests for our business logic 👍

cmd/network/verification_test.go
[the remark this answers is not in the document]
anandpulakala: @christophermancini , @btoll , 
incorporated the changes mentioned above.

cmd/network/packet-capture_test.go
```diff
@@ -1,11 +1,36 @@
...
+	"fmt"
+	"os"
+	"strings"
 	"testing"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
```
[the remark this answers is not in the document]
anandpulakala: @christophermancini ,
incorporated the changes mentioned above.

cmd/network/verification_test.go
anandpulakala: @christophermancini , @btoll 

On going through the code, I found multiple references to the reflect package, one of them being the below file

https://github.com/openshift/osdctl/blob/e049bffecfb689c003fb4750ee58653afd4b8b57/cmd/account/mgmt/account-list_test.go

Hence, the decision was taken to use the reflect package.

Also, using go-cmp package will result in dependencies being installed which we would want to ideally avoid.

cmd/network/verification_test.go
anandpulakala: As requested, the extra comment has been removed.

btoll:
I think there are problems with the `Test_egressVerificationGetPlatform` test, specifically the following, which I believe is giving a false positive:
```
        {                                                                                                                                                                                       
            name: "invalid",                                                                                                                                                                    
            e: &EgressVerification{                                                                                                                                                             
                cluster: newTestCluster(t, cmv1.NewCluster().CloudProvider(cmv1.NewCloudProvider().ID("foo"))),                                                                                 
                log:     newTestLogger(t),                                                                                                                                                      
            },        
            expectErr: true,                                                                                                                                                                          
        }, 
```

cmd/network/verification_test.go
```diff
@@ -39,6 +48,277 @@ func newTestCluster(t *testing.T, cb *cmv1.ClusterBuilder) *cmv1.Cluster {
 	return cluster
 }
 
+func TestNewCmdValidateEgress(t *testing.T) {
+	cmd := NewCmdValidateEgress()
+	assert.NotNil(t, cmd)
+	assert.Equal(t, "verify-egress", cmd.Use)
+}
```
btoll, line 55: This can be removed, it's just testing cobra.

cmd/network/verification_test.go
```diff
@@ -39,6 +48,277 @@ func newTestCluster(t *testing.T, cb *cmv1.ClusterBuilder) *cmv1.Cluster {
...
+			if tt.wantError {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+		})
+	}
+}
+
+func TestEgressVerification_GetPlatform(t *testing.T) {
```
btoll, line 98: I think it would be valuable to separate these into two distinct tests, one that tests `.platformName` and one that tests `.cluster`.

cmd/network/verification_test.go
btoll: Also, since you brought in the `assert` library, why not do:
```
    for _, tc := range testCases {                                                                                                                                                              
        t.Run(tc.name, func(t *testing.T) {                                                                                                                                                     
            got := generateServiceLog(tc.output, tc.clusterId)                                                                                                                                  
            assert.Equal(t, tc.want.Template, got.Template)                                                                                                                                     
            assert.Equal(t, tc.want.ClusterId, got.ClusterId)                                                                                                                                   
            assert.Equal(t, tc.want.TemplateParams, got.TemplateParams)                                                                                                                         
        })                                                                                                                                                                                      
    } 
```

cmd/network/verification_test.go
btoll: I'm looking again at this, and it's in scope to create more unit tests for this particular method.  For instance, the `getCaBundleFromManagementCluster` and `getCaBundleFromHive` could possibly have their paths tested.

The logic is somewhat confusing, too.  Passing it an empty string won't return an `error`, so why have a `wantErr` field in the struct?

cmd/network/verification_test.go
```diff
@@ -400,3 +678,76 @@ func TestGetCaBundleFromSyncSet(t *testing.T) {
...
+				log:             newTestLogger(t),
+			},
+			expectErr: false,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ctx := context.Background()
+			err := test.e.fetchCluster(ctx)
```
btoll, line 747: This isn't really testing much, because when a `ClusterId` isn't provided it's just short-circuiting the `fetchCluster` method and returning `nil`.  There's never a situation where an `error` would be returned.

cmd/network/packet-capture_test.go
```diff
@@ -31,3 +56,902 @@ func TestPacketCaptureCmdComplete(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ops := &packetCaptureOptions{
+				reason:  tt.reason,
+				kubeCli: &k8s.LazyClient{},
+			}
+			err := ops.complete(nil, nil)
```
btoll, line 95: This isn't really testing much.  When a reason is given, it should test the impersonated 'userName' and 'elevationReasons`.

Also, the `wantErr` pattern isn't useful here.  The `complete` method will never return a legitimate `error`.

cmd/network/packet-capture_test.go
```diff
@@ -31,3 +56,902 @@ func TestPacketCaptureCmdComplete(t *testing.T) {
...
+
+	assert.Equal(t, ops.name, pod.Name)
+	assert.Equal(t, ops.namespace, pod.Namespace)
+	assert.Equal(t, map[string]string{"app": ops.name}, pod.Labels)
+	assert.Equal(t, true, pod.Spec.HostNetwork)
+	assert.Equal(t, 1, len(pod.Spec.InitContainers))
+	assert.Equal(t, 1, len(pod.Spec.Containers))
+}
+
+func TestNewPacketCaptureOptionsOne(t *testing.T) {
```
btoll, line 147: This looks like a duplicate of `TestNewPacketCaptureOptions`.

cmd/network/packet-capture_test.go
```diff
@@ -31,3 +56,902 @@ func TestPacketCaptureCmdComplete(t *testing.T) {
...
+
+	assert.Equal(t, ops.name, ds.Name)
+	assert.Equal(t, ops.namespace, ds.Namespace)
+	assert.Equal(t, map[string]string{"app": ops.name}, ds.Spec.Selector.MatchLabels)
+	assert.Equal(t, true, ds.Spec.Template.Spec.HostNetwork)
+	assert.Equal(t, 1, len(ds.Spec.Template.Spec.InitContainers))
+	assert.Equal(t, 1, len(ds.Spec.Template.Spec.Containers))
+}
+
+func TestDesiredPacketCapturePodExist(t *testing.T) {
```
btoll, line 126: Is this significantly different from `TestDesiredPacketCapturePod`?

cmd/network/packet-capture_test.go
```diff
@@ -31,3 +56,902 @@ func TestPacketCaptureCmdComplete(t *testing.T) {
...
+	lazyClient := &k8s.LazyClient{}
+
+	ops := newPacketCaptureOptions(streams, lazyClient)
+
+	assert.NotNil(t, ops)
+	assert.Equal(t, streams, ops.IOStreams)
+	assert.Equal(t, lazyClient, ops.kubeCli)
+}
+
+func TestDesiredPacketCaptureDaemonSetOne(t *testing.T) {
```
btoll, line 158: How is this different from `TestDesiredPacketCaptureDaemonSet`?  Should they be grouped?

cmd/network/cmd_test.go
anandpulakala: Hi @joshbranham,
As per your comment, I have removed the cmd_test.go file as it only has test case for cobra command.
Below is the commit id for your reference:
0c3d955c531c7728bd8d5f1a8ef2c0f0d43d3ea8

pkg/k8s/client.go
```diff
@@ -170,3 +170,15 @@ func GetCurrentCluster() (string, error) {
...
+	return &LazyClient{
+		client: fc,
+	}
+}
+
+func LazyClientMock(c client.Client) *LazyClient {
+	return &LazyClient{
+		client: c,
+	}
+}
```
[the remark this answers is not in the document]
anandpulakala: client is private to k8s.LazyClient, that's why I have put LazyClientMock() in K8s.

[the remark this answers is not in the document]
btoll: In that case, we can use an interface which the mock would implement.  We can then remove these additions from that package.

cmd/network/verification_test.go
anandpulakala: changed as requested

cmd/network/verification_test.go
anandpulakala: Hi @btoll , 
   I have removed all unnecessary cases and only added case for nill error, 
If we go inside fetchCluster() it requires utils.createconnection mocking and as of now we are skipping it because  to mock utils we needs update its implementation from tightly coupled struct to interface based.

cmd/network/verification_test.go
anandpulakala: removed as suggested

cmd/network/verification_test.go
anandpulakala: seperated to two distinct tests as suggested

cmd/network/packet-capture_test.go
anandpulakala: I have removed the duplicate TestNewPacketCaptureOptions as suggested.

cmd/network/packet-capture_test.go
```diff
@@ -31,3 +56,850 @@ func TestPacketCaptureCmdComplete(t *testing.T) {
...
+
+			opts := &packetCaptureOptions{
+				kubeCli: mockClient.ToLazyClient(),
+			}
+			err := deletePacketCaptureDaemonSet(opts, tt.ds)
+			if tt.wantError {
+				assert.Error(t, err)
+				expectedErr := fmt.Errorf("failed to delete daemonset %s/%s: %v", tt.ds.Namespace, tt.ds.Name, tt.mockError)
+				assert.Equal(t, expectedErr.Error(), err.Error())
+			} else {
```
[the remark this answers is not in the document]
btoll: Because of the pattern that they're using, it makes sense to have an `else` clause.  I think we need to be careful of saying "don't use else" across the board, everything depends on the context.

pkg/k8s/client.go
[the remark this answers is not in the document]
anandpulakala: we can't mock that in packetcaptureOptions k8s.LazyClient is struct, to use mock we have to change it to interface. If  we change K8s.LazyClient to interface then we have to make changes at all the places where lazyClient is being called. 
if we don't want to change in client.go, then shall we skip this Test case ?

cmd/network/verification_test.go
anandpulakala: I have removed the extra cases, now it is testing only for negative case, positive cases cannot be tested as they are using utils.createConnection(), for this we need to mock utils package.

cmd/network/packet-capture_test.go
anandpulakala: I have removed wantErr field, but we cannot compare userName & elevationReasons as they are private to LazyClient.

cmd/network/packet-capture_test.go
anandpulakala: removed as it is not required

cmd/network/packet-capture_test.go
```diff
@@ -1,11 +1,36 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/openshift/osdctl/pkg/k8s"
```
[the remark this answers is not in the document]
anandpulakala: reordered with github.com

cmd/network/packet-capture_test.go
```diff
@@ -1,11 +1,36 @@
...
 )
 
+func TestHelperProcess(t *testing.T) {
```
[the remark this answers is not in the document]
anandpulakala: removed as it is not needed

cmd/network/verification_test.go
```diff
@@ -3,19 +3,28 @@ package network
 import (
 	"context"
 	"fmt"
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/runtime"
+
```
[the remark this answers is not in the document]
anandpulakala: removed extra line

cmd/network/packet-capture_test.go
```diff
@@ -31,3 +56,850 @@ func TestPacketCaptureCmdComplete(t *testing.T) {
...
+}
+
+func TestComplete(t *testing.T) {
+	tests := []struct {
+		name    string
+		reason  string
+		wantErr bool
+	}{
+		{
+			name:    "without reason",
```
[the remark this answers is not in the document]
anandpulakala: changed to snake_case as suggested

cmd/network/verification_test.go
```diff
@@ -3,19 +3,28 @@ package network
...
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+
```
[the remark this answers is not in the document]
anandpulakala: removed extra line

cmd/network/verification_test.go
```diff
@@ -3,19 +3,28 @@ package network
...
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/openshift-online/ocm-sdk-go/logging"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
+
 	"github.com/openshift/osd-network-verifier/pkg/data/cloud"
+	"github.com/openshift/osd-network-verifier/pkg/data/cpu"
 	"github.com/openshift/osd-network-verifier/pkg/output"
+	"github.com/openshift/osd-network-verifier/pkg/probes/curl"
 	onv "github.com/openshift/osd-network-verifier/pkg/verifier"
+
```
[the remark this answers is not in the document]
anandpulakala: removed extra line

cmd/network/verification_test.go
```diff
@@ -3,19 +3,28 @@ package network
...
+	"k8s.io/apimachinery/pkg/runtime"
+
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/openshift-online/ocm-sdk-go/logging"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
+
```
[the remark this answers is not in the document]
anandpulakala: removed extra line

rafael-azevedo (approved):
looks good if you can squash the commits next time 

/lgtm 

PR #678, opened 2025-03-05
Osd 22527: osdctl "jiratoken" flag in "cluster context" command is not used

Mhodesty:
Modified the GetJiraIssuesForCluster function to take an additional string "jiratoken." Prior to this PR the function didn't acknowledge a jiratoken passed in through the CLI. I also had to modify different calls for that function throughout the code to add an empty string to satisfy the requirements of the modified function.


cmd/jira/quick-task.go
```diff
@@ -54,7 +55,7 @@ osdctl jira quick-task "Update command to take new flag" --add-to-sprint
 		}
 		boardId := viper.GetInt(BoardIdLabel)
 
-		jiraClient, err := utils.GetJiraClient()
+		jiraClient, err := utils.GetJiraClient("")
```
joshbranham, line 58: When calling this without a token, isn't this always going to fail any lookups? I don't think we should support that?

Mhodesty: In pkg/utils/jira.go on lines 20-29 there is logic to handle if the jiratoken is an empty string. The GetJiraClient function follows the same logic it currently uses if the jiratoken passed in is an empty string. This allows us to handle all cases we currently support and also allow for a token to defined on command line.

pkg/utils/jira.go
```diff
@@ -16,29 +16,26 @@ const (
...
-	}
-
-	if os.Getenv("JIRA_API_TOKEN") != "" {
-		jiratoken = os.Getenv("JIRA_API_TOKEN")
-	}
-
+func GetJiraClient(jiratoken string) (*jira.Client, error) {
 	if jiratoken == "" {
-		return nil, fmt.Errorf("JIRA token is not defined.")
+		if !viper.IsSet(JiraTokenConfigKey) {
```
joshbranham, line 21: Shouldn't this return the viper config _if_ it's set, not `!set`?

cmd/jira/quick-task.go
joshbranham: Gotcha, I think it would be more concise to not allow passing a secret as a command line flag when that value is also stored in the config file, but I can understand if you don't want to go that route as part of this ticket.

clcollins: I agree it's more concise too, but I think Viper allows you to override config values in order of importance, and that ends up being useful for testing, at least.   CLI -> ENV -> Config (I _think_ that's the order).

I get where you're coming from too, though

joshbranham: Ah, yeah, in my head I would just do `MY_ENV=foo ./osdctl thing` and that would be my override testing versus adding the flag just for testing, but it doesn't matter too much.

joshbranham (commented):
/lgtm

PR #680, opened 2025-03-06
Update Documentation as a part of every release

ratnam915:
Originally executed as a part of the OSD-19018 story, currently being tracked as part of the https://issues.redhat.com/browse/OSD-28456

Changes made as part of the comments on the PR : https://github.com/openshift/osdctl/pull/674

1. Modified the script to have a state file that would hold the state of the changes made to the cmd directory, if any the script will create the necessary documentation.
2. Created a pre-push git hook that would look at changes to the osdctl_commands.md, only if the file as well as the directory is updated will the push go through, else it will prompt the developer to make sure the README file  and the folder is up to date.
3. Skipping the PROW change since the git hook will take care of document generation.

@joshbranham : Requesting you to have a look and post your comments on the same.


README.md
```diff
@@ -35,6 +35,12 @@ make build
 
 Then you can find the `osdctl` binary file in the `./dist` subdirectory matching your architecture.
 
+### Update documentation after every release##
```
joshbranham, line 38:
```suggestion
### Update documentation after every release
```

scripts/install-hooks.sh
```diff
@@ -0,0 +1,58 @@
...
+        echo "    git add osdctl_commands.md"
+        echo "    git commit -m \"Update documentation\""
+        exit 1
+    else
+        echo "Documentation is up to date."
+    fi
+else
+    echo "ERROR: Documentation file osdctl_commands.md was not generated."
+    exit 1
+fi
```
joshbranham, line 47: So I think you can greatly simplify the logic here by removing the state file altogether, and just generating the docs as part of the pre-commit hook. It should be pretty fast given it is Go generating markdown from the command structure. Then any diff will be caught and no need to check pre/post.

scripts/install-hooks.sh
```diff
@@ -0,0 +1,58 @@
...
+
+echo "Running pre-push hook to check documentation..."
+
+if [[ -f "osdctl_commands.md" ]]; then
+    PRE_MD5=$(md5sum "osdctl_commands.md" | cut -d ' ' -f 1)
+else
+    PRE_MD5=""
+fi
+
+echo "Generating documentation..."
```
joshbranham, line 24: I would not generate the documentation as part of installing the hooks, just use this to install the hooks.

scripts/pre-push
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
joshbranham, line 1: Can we also add a script that can be executed by `prow` to ensure docs are generated and there is no diff? We will need to configure that [here](https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/osdctl/openshift-osdctl-master-presubmits.yaml) after.

scripts/verify_docs.sh
```diff
@@ -0,0 +1,18 @@
...
+
+set -e
+
+echo "Verifying documentation..."
+
+cp -f osdctl_commands.md /tmp/original_docs.md 2>/dev/null || touch /tmp/original_docs.md
+
+make generate-docs
+
+if ! diff -q /tmp/original_docs.md osdctl_commands.md &>/dev/null; then
```
joshbranham, line 11: This would be a much simpler script if you just `make generate docs` and then check if their are git changes to `docs/.md`?

ratnam915: Hey Josh,

I've made the changes to make the whole workflow, please review and let me know your thoughts

.docgen_state
```diff
@@ -0,0 +1 @@
+92717fa0f5f5559302fdcd532144ee30ff2e7567e37cd1053c62f9a2479fea4c
```
joshbranham, line 1: You are still using a checksum here to compare things to gate wether docs have changed. I still feel strongly that we should just generate the docs every time and then use the git checkout status to understand if there is a diff.

ratnam915: Hey @joshbranham , 

You are right, sorry for the oversight, i have changed the script to primarily focus on generating the documenation.

Git checkout will take care of the difference

scripts/verify_docs.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -e
+
+echo "Verifying documentation..."
+
+make generate-docs
+
+if git diff --name-only | grep -q "\.md$"; then
```
iamkirkbater, line 8: Should we validate that this only looks within the `docs` folder in case there are any changes to a `.md` file anywhere else? Not sure if that would happen, but it would be a weird edge-case if you wanted to put some helpful `.md` file in a subfolder and then all of a sudden the merge jobs are blocked because a potential diff there.

osdctl_commands.md
iamkirkbater, line 1: I can't figure out how to word this without coming off as passive-aggressive or confrontational but I want to be clear that I'm curious: what is the point of this file here if we're already creating everything within the docs directory already? Happy to keep it if there's a good justification for it, but I'm wondering if this is just a duplicate of everything that we already have from the generate-docs tool?

pkg/docgen/docgen.go
iamkirkbater, line 1: If I understand correctly, this .go file is just to generate the `osdctl_commands.md` file? Or does this actually do more?

osdctl_commands.md
iamkirkbater: Maybe searchability would be easier in a single file? 🤔 

I wonder if we just move this file into the `docs` folder as something like `all_osdctl_commands.md` instead of keeping it in the root and that would be enough here?

iamkirkbater (changes requested):

osdctl_commands.md
ratnam915: Hey @iamkirkbater :

This script was initially created as a requirement from the Jira story that the README.md document was not being updated after every release and we wanted to move away from depending on engineers to manually update the README.md considering the number of changes being done to the repository.

The initial thought process was to append to the exisiting README.md document but we quickly moved away from that and went for the approach that we had an alternate file where all the information was stored, this file now acts as a single place where all the commands and the usage are hosed.

As you mentioned the docs folder is where these .md files are created and the reason this is in the root folder is that it is essentially an alternate to the README.md document.

pkg/docgen/docgen.go
ratnam915: @iamkirkbater : Yes this file is to create the markdown document primarily.

scripts/verify_docs.sh
ratnam915: @iamkirkbater : The thought process behind this was the make generate-docs would be the only script that would be generating .md documents, i think the edge case does sound valid though, do you want me to change the script to check only in the docs/ folder?

osdctl_commands.md
iamkirkbater: Fair enough. Do you have any preference one way or another to move this to the docs folder versus keeping this in the root folder? Maybe we could even move this to the docs and have this be the `README.md` in the docs folder or something similar to that? 

I don't have a preference one way or the other, so I'll leave the decision to you in this case.

ratnam915: i agree with the recommendation, i've made changes for the file to be the README.md in the docs/ directory.

scripts/verify_docs.sh
ratnam915: Sure Kirk, changes have been made accordingly

iamkirkbater (approved):
/lgtm

PR #684, opened 2025-03-07
[WIP] OSD-28645 - always  choose routetable to examine

tnierman:
https://issues.redhat.com/browse/OSD-28645 

---

At present, `osdctl cluster cpd` does not always choose a routeTable when examining its cluster's subnets.

BYOVPC clusters need to have their routetables checked to ensure proper connectivity, but the `routeTable` [variable we initialize](https://github.com/openshift/osdctl/blob/master/cmd/cluster/cpd.go#L138) to hold the name of each subnets' routetable only gets [set](https://github.com/openshift/osdctl/blob/master/cmd/cluster/cpd.go#L169) if the cluster does _not_ have explicit RT associations. If it does have an explicit association, then it gets [called](https://github.com/openshift/osdctl/blob/master/cmd/cluster/cpd.go#L177) as an empty value, leading to errors like 

```
error: operation error EC2: DescribeRouteTables, https response error StatusCode: 400, RequestID: 52f0a696-ba79-437e-8a9f-5cb5e6a7ad99, api error InvalidRouteTableID.NotFound: The routeTable ID '' does not exist
```

Using the existing `FindRouteTableForSubnet()` function ensures that a routeTable is always picked for examination, while also reducing the redundancy of the code.


joshbranham (commented):
/lgtm

PR #688, opened 2025-03-09
Added unit  test case for current and customer file under org package

amitupadh:
**Description**
This PR adds unit test cases for the current and customers commands under the org package, enhancing test coverage and improving maintainability by replacing monkey patching with dependency injection.

**Code Re factoring**
    Updated functions to accept ocmClient as a parameter, allowing it to be mocked.
    Modified current.go and customer.go to support the new approach.
 
**Coverage Details**
    Increased unit test coverage for current and customers commands.

**Verification**

**Jira Ticket** 
https://issues.redhat.com/browse/OSD-28397

### DIFF
--- a/cmd/org/current_test.go
+++ b/cmd/org/current_test.go
@@ -1,60 +1,61 @@
 package org
 
 import (
-	"encoding/json"
-	"net/http"
-	"net/http/httptest"
 	"testing"
 
-	sdk "github.com/openshift-online/ocm-sdk-go"
+	"github.com/openshift/osdctl/pkg/utils"
 )
 
-func Test_run(t *testing.T) {
-	t.Run("Success Test", func(t *testing.T) {
-		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			w.Header().Set("Content-Type", "application/json")
-			w.WriteHeader(http.StatusOK)
+var userData = `
+{
+    "created_at": "2024-11-27T07:32:59.368563Z",
+    "email": "kilgore.trout@redhat.com",
+    "first_name": "Kilgore",
+    "href": "/api/accounts_mgmt/v1/accounts/xyz",
+    "id": "foobar",
+    "kind": "Account",
+    "last_name": "Trout",
+    "organization": {
+        "created_at": "2019-02-15T20:26:12.542449Z",
+        "ebs_account_id": "1111111",
+        "external_id": "2222222",
+        "href": "/api/accounts_mgmt/v1/organizations/xyz",
+        "id": "abc.xyz",
+        "kind": "Organization",
+        "name": "Kurt Vonnegut Appreciation Society",
+        "updated_at": "2025-03-10T06:16:08.047253Z"
+    },
+    "rhit_web_user_id": "57712380",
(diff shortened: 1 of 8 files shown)


btoll:
> added unit test case for current and customer file under org package

Can you add a link to the Jira ticket(s) please?

cmd/org/current_test.go
abyrne55, line 1: Seems like this file is mostly testing the OCM SDK; it's also not clear where getCustomers() is being called

cmd/org/customers_test.go
abyrne55, line 1: Seems like this file is mostly testing the OCM SDK. Could we add some output validation for getCustomers?

amitupadh: As per the code snippet below:

`cmdutil.CheckErr(getCustomers(cmd, ocmClient))`

I could modify `getCustomers` to return `customerList, nil`, but `getCustomers` is invoked within cmdutil.CheckErr(), which only accepts an error return type. 
Since `cmdutil.CheckErr() `is part of an external package and used in multiple places, modifying its behaviour is not an option.

Due to this, I cannot change the return type of `getCustomers`. However, the function does execute `printCustomers(customerList) ` before returning, which prints the customer list.

I have updated the test results in the corresponding story. Please consider this as the test execution result.

Story: [OSD-28397](https://issues.redhat.com/browse/OSD-28397)
Subtask: [OSD-28499](https://issues.redhat.com/browse/OSD-28499)

amitupadh: Thanks for the feedback! 
I've updated `getCustomers` to return data and an error, moving printing to the command function for better testability and separation of concerns. 
Let me know if any further changes are needed!

amitupadh: As per your suggestion, I have implemented the new approach in current.go and customer.go. Everything is working as expected. However, I noticed that the overall code coverage has decreased—it's currently below 40%, whereas the previous implementation had coverage above 50%.

To provide better visibility, I have pushed both the new implementation and the previous approach:
Old Approach (Coverage > 50%)

    customer.go
    customer_test.go
    current.go
    current_test.go

New Approach (Coverage < 40%)

    customer_new.go
    customer_new_test.go
    current_new.go
    current_new_test.go

I would love to hear your thoughts on how we can improve test coverage while maintaining the benefits of the new structure or to follow old approach.
 Let me know if you have any suggestions!

btoll: @amitupadh Thanks for being open to the new approach and implementing it!

As for your concerns, I personally feel that it's best that we're testing the correct things rather than meeting a goal that may or may not be realistic.  The important thing is that we're increasing the amount of test coverage from 0%.

Would you mind closing this and creating a new draft PR with just the new approach?  I understand your intentions here, but it's made it more difficult to read, at least for me.

amitupadh: Hi @btoll  
Thanks for your feedback! I’ve created a new PR with just the new approach as requested: [#692](https://github.com/openshift/osdctl/pull/692). Let me know if you have any suggestions or if any further adjustments are needed.
New PR : [#692](https://github.com/openshift/osdctl/pull/692)
Closing this PR.

btoll: Thanks!

PR #689, opened 2025-03-11
Remove unused --cluster-id flag from context command

joshbranham:
I have noticed this command has a `--cluster-id` flag, but requires the cluster as a positional argument.

Additionally, the builtin cobra validation means that it would never hit the check that was on line 138 so I removed it.

Also made some minor cleanup

Sample run with changes:
```
 ./dist/osdctl_darwin_arm64_v8.0/osdctl cluster context 2d4c5kqu9adgrbcsl78dkc6h5qajfecv
WARN: The current version (v0.39.0-next) is different than the latest released version (v0.42.0). It is recommended that you update to the latest released version to ensure that no known bugs or issues are hit.
Continue? (y/N): y
Encountered Errors during data collection. Displayed data may be incomplete:
        error while getting support exceptions: error connecting to jira: JIRA token is not defined.
        error while getting the open jira tickets: error connecting to jira: JIRA token is not defined.
===================================================
ap-use1-t-dp-01 -- 2d4c5kqu9adgrbcsl78dkc6h5qajfecv
===================================================
ID:                             2d4c5kqu9adgrbcsl78dkc6h5qajfecv
....
```

### DIFF
--- a/cmd/cluster/context.go
+++ b/cmd/cluster/context.go
@@ -28,7 +28,6 @@ import (
 	"github.com/openshift/osdctl/pkg/utils"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


cmd/cluster/context.go
```diff
@@ -110,14 +109,17 @@ func newCmdContext() *cobra.Command {
...
+			if err != nil {
+				return err
+			}
+
+			return ops.run()
 		},
 	}
 
 	contextCmd.Flags().StringVarP(&ops.output, "output", "o", "long", "Valid formats are ['long', 'short', 'json']. Output is set to 'long' by default")
-	contextCmd.Flags().StringVarP(&ops.clusterID, "cluster-id", "C", "", "Cluster ID")
```
joshbranham, line 120: In the longer term I want to standardize across all commands either positional arg for cluster or `--cluster-id` flag, not both.

typeid (approved):
/lgtm
/approve

PR #690, opened 2025-03-11
Osd 24440: Elevate oc cp command

Mhodesty:
This PR elevates the oc cp command and allows us to actually copy network traces off of a cluster. The tricky part of this PR was the network interfaces. The previous logic looked for the ovnkube-master daemonset in the openshift-ovn-kubernetes namespace. If it was found, it would set the network type to the appropriate type based on the fact that the cluster was an OVN cluster. This did not work on multiple clusters in my testing, so I re-wrote the logic to use the network config and based on the output of that value set the network interface type.

More details in this slack thread: https://redhat-internal.slack.com/archives/C01M1NRKUF4/p1741368811179239?thread_ts=1741283963.209249&cid=C01M1NRKUF4


cmd/network/packet-capture.go
```diff
@@ -517,15 +517,16 @@ func waitForPacketCapturePod(o *packetCaptureOptions, capturePod *corev1.Pod) er
...
 	}
 
-	if ds.Status.DesiredNumberScheduled > 0 {
+	if networkConfig.Spec.NetworkType == "OVNKubernetes" {
 		o.captureInterface = "genev_sys_6081"
 		return nil
+	} else {
+		o.captureInterface = "vxlan_sys_4789"
+		return nil
 	}
```
joshbranham, line 531: Does this mean that SDN uses `vxlan_sys_4789` and then OVN obviously uses `genev_sys_6081`? If so, let's make this a case block and then have a `default:` throw and error for an unknown/unsupported network type.

cmd/network/packet-capture.go
```diff
@@ -517,15 +517,19 @@ func waitForPacketCapturePod(o *packetCaptureOptions, capturePod *corev1.Pod) er
...
-	if ds.Status.DesiredNumberScheduled > 0 {
+	switch networkConfig.Spec.NetworkType {
+	case "OVNKubernetes":
 		o.captureInterface = "genev_sys_6081"
 		return nil
+	case "OpenShiftSDN":
+		o.captureInterface = "vxlan_sys_4789"
+		return nil
+	default:
+		return fmt.Errorf("failed to determine network type or unknown network type")
```
joshbranham, line 533: One last `nit`: let's put the value in this string so it's quick to spot if it comes up :)

joshbranham (commented):
Thanks for making those tweaks

/lgtm

PR #691, opened 2025-03-11
Remove legacy controlplane resize command

RaphaelBut:
Always use cpms as we are no longer supporting clusters without it.

Updated the SOP accordingly, see https://github.com/openshift/ops-sop/pull/3475 

I tested a resize in stage with this change :)


joshbranham (commented):
Thanks for cleaning this up, good to see us remove legacy behavior.

/lgtm

PR #692, opened 2025-03-13
Added unit test case for current and customer file under org package

amitupadh:
Description
This PR adds unit test cases for the current and customers commands under the org package, enhancing test coverage and improving maintainability by replacing monkey patching with dependency injection.

Code Re factoring
Updated functions to accept ocmClient as a parameter, allowing it to be mocked.
Modified current.go and customer.go to support the new approach.

Coverage Details
Increased unit test coverage for current and customers commands.

Verification

Jira Ticket
https://issues.redhat.com/browse/OSD-28397


cmd/org/customers.go
```diff
@@ -1,127 +1,97 @@
...
+		if err != nil {
+			cmdutil.CheckErr(fmt.Errorf("failed to fetch customers: %v", err))
+		}
+		customers, err := parseCustomers(customersResp.Bytes())
+		if err != nil {
+			cmdutil.CheckErr(err)
+		}
+
+		printCustomers(customers)
+	},
```
btoll, line 57: Looks like you brought this logic over from `current.go`.   Do we need to do this here (i.e., get the org)?

amitupadh: @btoll 
Yes, printCustomers(customers) is needed here, similar to how get the org is used in current.go.
It was also present in the initial implementation before the refactoring, so I kept it to maintain the original functionality.

btoll: Yes, that part is fine, but I'm speaking to where you're using `getOrgRequest` and `getCurrentOrg`.  Also, it looks like you're now querying a different REST API.

amitupadh: @btoll 
I acknowledge that the current implementation queries a different API than intended. I have corrected this discrepancy to ensure alignment with our objectives.

Regarding the proposed approach, after careful consideration, I believe that extensive refactoring may not be feasible without compromising existing functionalities. The initial code structure allows for limited modifications to enhance testability. Consequently, I suggest reverting to the previous methodology, which, despite its constraints, facilitates effective testing within our current framework.

I appreciate your understanding and am open to further discussions to explore alternative solutions that balance both functionality and testability.

cmd/org/common_test.go
```diff
@@ -1,9 +1,118 @@
...
+		Client(fakeClientID, fakeClientSecret).
+		Build()
+	if err != nil {
+		t.Fatalf("Failed to build connection: %v", err)
+	}
+	defer conn.Close()
+
+	// Test the function
+	req := conn.Get().Path(testEndpoint)
+	_, err = sendRequest(req)
```
btoll, line 71: Looks like this is testing the OCM layer.

cmd/org/common_test.go
```diff
@@ -1,9 +1,118 @@
...
+	fakeStatus := "Active"
+	fakeManagedOnly := true
+	fakePage := 1
+	fakeSize := 20
+
+	// Create a fake OCM client
+	ocmClient := &sdk.Connection{}
+
+	// Call the function under test
+	request := createGetSubscriptionsRequest(ocmClient, fakeOrgID, fakeStatus, fakeManagedOnly, fakePage, fakeSize)
```
btoll, line 90: It doesn't look like this function is being called from anywhere.  Do we want to test this?  Also, do we want to be testing functions that aren't part of the public API?

cmd/org/common_test.go
```diff
@@ -40,3 +149,90 @@ func TestCheckOrgId(t *testing.T) {
...
+	assert.Regexp(t, regexp.MustCompile(`^Name:\s+Example Org$`), lines[1])
+	assert.Regexp(t, regexp.MustCompile(`^External ID:\s+ext123$`), lines[2])
+	assert.Regexp(t, regexp.MustCompile(`^Created:\s+2023-01-01T00:00:00Z$`), lines[4])
+	assert.Regexp(t, regexp.MustCompile(`^Updated:\s+2023-01-02T00:00:00Z$`), lines[5])
+}
+
+func Test_getSubscriptions(t *testing.T) {
+	t.Run("TestGetSubscriptions", func(t *testing.T) {
+		// Create a mock OCM client
+		got, err := getSubscriptions("org2", "active", true, 5, 10)
```
btoll, line 226: It doesn't look like this function is being called from anywhere. Do we want to test this? Also, do we want to be testing functions that aren't part of the public API?

cmd/org/common_test.go
```diff
@@ -1,9 +1,118 @@
...
+	var (
+		tokenPath        = "/oauth2/token"
+		testEndpoint     = "/test-endpoint"
+		fakeClientID     = "fake-id"
+		fakeClientSecret = "fake-secret"
+	)
+
+	// Test token response
+	tokenResponse := `{
+	"access_token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJ0ZXN0LXVzZXIiLCJleHAiOjI1MjQ2MDgwMDB9.signature-placeholder",
```
btoll, line 32: This will probably be flagged by tooling looking for secrets.

cmd/org/customers_test.go
```diff
@@ -0,0 +1,114 @@
...
+			URL(server.URL).
+			TokenURL(server.URL+tokenPath).
+			Insecure(true).
+			Client(clientID, clientSecret).
+			Build()
+		if err != nil {
+			t.Fatalf("Failed to build connection: %v", err)
+		}
+
+		customers, err := getCustomers(nil, conn)
```
btoll, line 71: Can we just test the returned customers without spinning up a test server?

btoll (commented):
The comments can be removed, as the code is already descriptive enough.

A couple of things I'm wondering:
- Is testing the `common.go` file out of scope for this ticket?  Is it even necessary?
- Do we want to be testing the print functions?  Would it be better to test that the data is being returned properly and not the formatting of it?

cmd/org/customers_test.go
amitupadh: @btoll 

Mocking the OCM SDK turned out to be tricky due to unexported fields and lack of public builders for responses, which caused compile issues when trying to create mock data. 
Given these limitations, using a fake server seemed like the most reliable and maintainable option to simulate realistic API behavior without depending on SDK internals. 
Let me know if you'd like me to explore another approach!

Also, I mistakenly included changes to common_test.go in this PR—I have reverted those.
I'll share different PR for those changes

cmd/org/common_test.go
amitupadh: reverted this file changes

cmd/org/common_test.go
amitupadh: reverted this file changes

cmd/org/common_test.go
amitupadh: reverted this file changes

cmd/org/common_test.go
amitupadh: reverted this file changes

cmd/org/common_test.go
```diff
@@ -39,4 +39,4 @@ func TestCheckOrgId(t *testing.T) {
 			}
 		}
 	}
-}
+}
```
btoll, line 42: Can you just remove this file?

amitupadh: Removed

cmd/org/customers.go
```diff
@@ -19,7 +19,21 @@ var (
 		Args:          cobra.ArbitraryArgs,
 		SilenceErrors: true,
 		Run: func(cmd *cobra.Command, args []string) {
```
btoll, line 21: `cmd` and `args` aren't used in the function body, they can be `_`.  I think that helps readability.

cmd/org/customers.go
```diff
@@ -19,7 +19,21 @@ var (
...
+			if err != nil {
+				cmdutil.CheckErr(err)
+			}
+			defer func() {
+				if err := ocmClient.Close(); err != nil {
+					fmt.Printf("Cannot close the ocmClient (possible memory leak): %q", err)
+				}
+			}()
+
+			customers, err := getCustomers(cmd, ocmClient)
```
btoll, line 32: You don't need to pass `cmd` to `getCustomers`.

cmd/org/customers.go
```diff
@@ -51,21 +65,10 @@ func init() {
 	AddOutputFlag(flags)
 }
 
-func getCustomers(cmd *cobra.Command) error {
+func getCustomers(cmd *cobra.Command, ocmClient *sdk.Connection) ([]Customer, error) {
```
btoll, line 68: This shouldn't accept `cmd`, it's not needed.

cmd/org/customers_test.go
btoll: Yeah, I looked at this and it may be more trouble than it's worth right now to refactor it to make it more testable.

cmd/org/customers.go
amitupadh: Fixed  as per suggestion

cmd/org/customers.go
amitupadh: Fixed  as per suggestion

cmd/org/customers.go
amitupadh: Fixed  as per suggestion

cmd/org/customers_test.go
```diff
@@ -0,0 +1,110 @@
...
+	clientID      = "fake-id"
+	clientSecret  = "fake-secret"
+	tokenPath     = "/fake-path/token"
+	testCustomers = []Customer{
+		{ID: "cust-1", OrganizationID: "org-1", SKU: "sku-1"},
+		{ID: "cust-2", OrganizationID: "org-2", SKU: "sku-2"},
+	}
+)
+
+func Test_getCustomers(t *testing.T) {
```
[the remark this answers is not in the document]
amitupadh: @christophermancini  Updated as per suggestion

cmd/org/current_test.go
```diff
@@ -0,0 +1,81 @@
...
+        "id": "abc.xyz",
+        "kind": "Organization",
+        "name": "Kurt Vonnegut Appreciation Society",
+        "updated_at": "2025-03-10T06:16:08.047253Z"
+    },
+    "rhit_web_user_id": "57712380",
+    "updated_at": "2025-02-21T21:05:50.544761Z",
+    "username": "kilgore.trout"
+}
+`
```
joshbranham, line 34: This is a bit of a nit but can you not store this in a global variable? Maybe return it as a private func in this test file?

joshbranham (commented):
Left one comment. Additionally can you squash your commits next time?

cmd/org/current_test.go
amitupadh: Thanks for the suggestion, @joshbranham ! That makes sense. 
I'll update it to return from a private function to keep the scope limited within the test file.

btoll (approved):
Thanks for your work on this!

PR #694, opened 2025-03-18
OSD-22389: To unify osdctl cluster identifier UX

ratnam915:
As per https://issues.redhat.com/browse/OSD-22389, created the below PR:

Some commands in osdctl were using cluster-id as an argument where others were using cluster-id as a flag, changed the code to make sure all commands are consistently using cluster-id as a flag.

Below are the commands,sub commands that were identified and changed:

    • alert	
        ◦ list
        ◦ silence add
        ◦ silence clear
        ◦ silence list
    • cluster
        ◦ break-glass
        ◦ cleanup
        ◦ check-banned-user
        ◦ context
        ◦ cpd
        ◦ detach-stuck-volume
        ◦ etcd-health-check
        ◦ etcd-member-replace
        ◦ health
        ◦ hypershift-info
        ◦ logging-check
        ◦ orgId
        ◦ ssh key
        ◦ support delete
        ◦ support post
        ◦ support status
        ◦ validate-pull-secret
    • dynatrace
        ◦ gather-logs
        ◦ logs
        ◦ url
    • hcp
        ◦ must-gather
    • servicelog	
        ◦ list
        ◦ post


cmd/alerts/silence/add_silence.go
```diff
@@ -26,22 +26,29 @@ type addSilenceCmd struct {
...
-		Args:              cobra.ExactArgs(1),
+		Args:              cobra.NoArgs,
 		DisableAutoGenTag: true,
 		Run: func(cmd *cobra.Command, args []string) {
-			addSilenceCmd.clusterID = args[0]
+			if addSilenceCmd.clusterID == "" {
+				fmt.Println("Error: --cluster-id flag is required")
+				_ = cmd.Help()
+				return
+			}
```
btoll, line 39: The way the code is structured is a bit confusing.  Instead of checking this field, can't cobra be leveraged to check for that `--cluster-id` is set since it's marked as required?

cmd/alerts/silence/list_silence.go
```diff
@@ -77,13 +81,13 @@ func printSilence(silence utils.Silence) {
...
-		fmt.Printf("Ends At:	%s\n", end)
-		fmt.Printf("Comment:	%s\n", comment)
-		fmt.Printf("AlertName:	%s\n", matcher.Value)
+		fmt.Printf("SilenceID: %s\n", id)
+		fmt.Printf("Status: %s\n", status.State)
+		fmt.Printf("Created By: %s\n", created)
+		fmt.Printf("Starts At: %s\n", starts)
+		fmt.Printf("Ends At: %s\n", end)
+		fmt.Printf("Comment: %s\n", comment)
+		fmt.Printf("AlertName: %s\n", matcher.Value)
```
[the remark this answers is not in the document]
ratnam915: @btoll  : Would you recommend using tabwriter in this case?

cmd/cluster/checkbanneduser.go
```diff
@@ -10,15 +11,21 @@ import (
...
+		Args:              cobra.NoArgs,
 		DisableAutoGenTag: true,
 		Run: func(cmd *cobra.Command, args []string) {
-			cmdutil.CheckErr(CheckBannedUser(args[0]))
+			cmdutil.CheckErr(CheckBannedUser(clusterID))
 		},
 	}
+
+	cmd.Flags().StringVarP(&clusterID, "cluster-id", "c", "", "Provide internal ID of the cluster")
+	cmd.MarkFlagRequired("cluster-id")
```
btoll, line 26: We should be consistent about how we're handling the return value.  Ideally, we're handling the possible error like we do in other areas of the code instead of ignoring it.

cmd/alerts/silence/add_silence.go
ratnam915: @btoll :

We can still use the global flag everywhere, but it would become redundant, the reason for using the individual flag was that the initial setup was taking arguments and a lot of functions within the command were dependent on these, hence since we were changing this to a flag all the dependencies within each file had to be fixed.

Also using the global flag would not solve our dependencies with these individual functions.

cmd/cluster/checkbanneduser.go
ratnam915: Sure, i'll make the changes so that the return value and error handling is consistent across all the file changes.

cmd/alerts/list_alerts.go
```diff
@@ -21,16 +21,17 @@ type alertCmd struct {
...
 	alertCmd := &alertCmd{}
 	newCmd := &cobra.Command{
-		Use:               "list <cluster-id> --level [warning, critical, firing, pending, all]",
-		Short:             "List all alerts or based on severity",
-		Long:              `Checks the alerts for the cluster and print the list based on severity`,
-		Args:              cobra.ExactArgs(1),
+		Use:   "list --cluster-id <cluster-id> --level [warning, critical, firing, pending, all]",
+		Short: "List all alerts or based on severity",
+		Long:  `Checks the alerts for the cluster and print the list based on severity`,
+		// Remove the Args: cobra.ExactArgs(1) line
```
joshbranham, line 27: We don't need this comment

cmd/alerts/silence/add_silence.go
```diff
@@ -26,22 +26,25 @@ type addSilenceCmd struct {
 func NewCmdAddSilence() *cobra.Command {
 	addSilenceCmd := &addSilenceCmd{}
 	cmd := &cobra.Command{
-		Use:               "add <cluster-id> [--all --duration --comment | --alertname --duration --comment]",
+		Use:               "add [--cluster-id <cluster-identifier>] [--all --duration --comment | --alertname --duration --comment]",
```
joshbranham, line 29: If `--cluster-id` is required I think it should be not in brackets right?
```suggestion
		Use:               "add --cluster-id <cluster-identifier> [--all --duration --comment | --alertname --duration --comment]",
```

cmd/cluster/etcd_health.go
```diff
@@ -51,38 +51,47 @@ func (capture *LogCapture) Write(p []byte) (n int, err error) {
...
 	}
 
+	cmd.Flags().StringVar(&opts.clusterID, "cluster-id", "", "Provide the internal Cluster ID or name to perform health check on")
 	cmd.Flags().StringVar(&opts.reason, "reason", "", "Specify a reason for privilege escalation")
-	err := cmd.MarkFlagRequired("reason")
+
+	err := cmd.MarkFlagRequired("cluster-id")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "WARNING: failed to mark flag '--cluster-id' as required: %v", err)
+	}
```
joshbranham, line 77: You aren't checking this error elsewhere. I usually _don't_ check this, it only fails if you haven't created a flag that matches. Either way, let's be consistent.

cmd/cluster/hypershift_info.go
```diff
@@ -134,8 +133,11 @@ func newInfoOptions(streams genericclioptions.IOStreams) *infoOptions {
 
 func (i *infoOptions) complete(cmd *cobra.Command) error {
 	var errMsg string
+	if i.clusterID == "" {
+		errMsg = "missing required flag --cluster-id. "
+	}
```
joshbranham, line 138: This is marked required by Cobra, therefore this should not be needed.

cmd/cluster/org_id.go
```diff
@@ -17,23 +19,30 @@ type OrgIdOutput struct {
...
-			if err := o.Run(args[0]); err != nil {
-				return fmt.Errorf("error fetching OCM org ID for cluster %v", args[0])
+			if err := o.Run(); err != nil {
+				return fmt.Errorf("error fetching OCM org ID for cluster %v", o.clusterID)
 			}
 			return nil
 		},
 	}
+
+	// Add cluster-id flag
```
joshbranham, line 34: This comment is not needed.

cmd/cluster/org_id.go
```diff
@@ -17,23 +19,30 @@ type OrgIdOutput struct {
...
+	orgIdCmd.Flags().StringVarP(&o.clusterID, "cluster-id", "c", "", "The internal ID of the cluster to check (required)")
+	if err := orgIdCmd.MarkFlagRequired("cluster-id"); err != nil {
```
joshbranham, line 36: Another example of checking the error whereas other places we are not. Consistency is preferred :)

joshbranham (commented):
Left some comments around lack of consistency for:

- Checking errors when `--cluster-id` is required
- Double checking the value is set outside of the Cobra func, which shouldn't be possible
- Ensuring `Use` and any other text generating config includes real use cases that showcase `--cluster-id` being required

cmd/cluster/org_id.go
ratnam915: Hey @joshbranham,

As suggested all the extra comments as well as the extra error checks have been removed to maintain consistency across the files

PR #695, opened 2025-03-19
OSD-28400: Added test cases for clustersync.go

drakshakhan:
This PR adds test cases for the Cluster Sync file, improving the test coverage from 0% to 41.4%. The following changes were made:

Test Case Addition
Added unit tests to thoroughly test the Cluster Sync functionality.

Test Frameworks Used
GoMock to mock dependencies and isolate unit tests.

Coverage Details
The code coverage for the Cluster Sync file has been increased to 54.2%. This includes tests for various scenarios, such as edge cases and failure paths.

Test Execution

Verification

Jira Story - https://issues.redhat.com/browse/OSD-28400


cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,276 @@
...
+	"testing"
+	"time"
+
+	"os"
+
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
```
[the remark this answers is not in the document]
drakshakhan: sorted the imports by grouping the standard library imports first, followed by the third-party packages, and then the local packages.

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,276 @@
...
+			ErrorMessage:    "Failed to sync selectorsyncset1\n\nFailed to sync syncset1\n\n",
+		},
+	}
+
+	testListFailingClusterSyncs(t, nil, expectedResults)
+}
+
+func TestSortBy(t *testing.T) {
+
+	tests := []struct {
```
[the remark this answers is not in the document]
drakshakhan: Thanks. I find table-driven tests to be a great way to keep the tests concise and easy to maintain. Appreciate the feedback.

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,276 @@
...
+		sortField   string
```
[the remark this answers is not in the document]
drakshakhan: Thanks for the suggestion. I've added a name field for each test case as recommended. This should make it much easier to identify and troubleshoot specific test failures. Appreciate the feedback.

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,276 @@
...
+		sortOrder   string
+		expected    []failingClusterSync
+		expectedErr string
+	}{
+		// Test sorting by name in ascending order
+		{
```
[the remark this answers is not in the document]
drakshakhan: implemented the changes as per the suggestion.

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,276 @@
...
+			},
+		},
+		// Test invalid sort field
+		{
+			sortField: "invalid", sortOrder: "asc", expectedErr: "Specify one of the following fields as a sort argument: name, timestamp, failingsyncsets.",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.sortField+"-"+tt.sortOrder, func(t *testing.T) {
```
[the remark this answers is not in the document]
drakshakhan: Thanks for the suggestion. I’ve updated the test to use tt.name for the test case name.

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,276 @@
...
+
+			failingClusterSyncList := []failingClusterSync{
+				{Name: "zeta", Timestamp: "2023-01-01T00:00:00Z", FailingSyncSets: "syncset3"},
+				{Name: "alpha", Timestamp: "2022-01-01T00:00:00Z", FailingSyncSets: "syncset1"},
+				{Name: "beta", Timestamp: "2023-02-01T00:00:00Z", FailingSyncSets: "syncset2"},
+			}
+
+			err := options.sortBy(failingClusterSyncList)
+
+			if tt.expectedErr != "" {
```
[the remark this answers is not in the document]
drakshakhan: removed the extra newline.

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,270 @@
...
+			case *hivev1.ClusterDeploymentList:
+				*v = cdList
+			case *v1alpha1.ClusterSyncList:
+				*v = csList
+			}
+			return returnErr
+		}).Times(2) // Expect List to be called twice for ClusterDeployment and ClusterSync
+}
+
+func TestNewCmdClusterSyncFailures(t *testing.T) {
```
btoll, line 89: We don't need to be testing the `cobra` API.

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,270 @@
...
+	cmd.Flags().Set("output", "json")
+
+	err := cmd.Execute()
+	assert.NoError(t, err, "Command should execute without error")
+
+	err = cmd.Flags().Set("invalid-flag", "true")
+	assert.Error(t, err, "Setting invalid flag should return an error")
+}
+
+func testListFailingClusterSyncs(t *testing.T, returnErr error, expectedResults []failingClusterSync) {
```
btoll, line 135: Can't this be in the body of `TestListFailingClusterSyncs`?

cmd/hive/clustersync_test.go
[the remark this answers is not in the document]
btoll: Would you mind formatting these objects like this?
```
        // Test sorting by name in ascending order                                                                                                              
        {                                                                                                                                                       
            name: "sort_by_name_ascending_order",                                                                                                               
            sortField: "name",                                                                                                                                  
            sortOrder: "asc",                                                                                                                                   
            expected: []failingClusterSync{                                                                                                                     
                {Name: "alpha", Timestamp: "2022-01-01T00:00:00Z", FailingSyncSets: "syncset1"},                                                                
                {Name: "beta", Timestamp: "2023-02-01T00:00:00Z", FailingSyncSets: "syncset2"},                                                                 
                {Name: "zeta", Timestamp: "2023-01-01T00:00:00Z", FailingSyncSets: "syncset3"},                                                                 
            },                                                                                                                                                  
        },
```
I find that much easier to read.

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+	setupMockClient(mockClient, returnErr)
+
+	options := &clusterSyncFailuresOptions{
+		kubeCli: mockClient,
+	}
+
+	result, err := options.listFailingClusterSyncs()
+
+	if returnErr != nil {
```
btoll, line 149: This will never be `true` for these tests, right?

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,270 @@
...
+		{
+			Name:            "example-clustersync",
+			Namespace:       "uhc-production-1234",
+			Timestamp:       time.Now().Format(time.RFC3339),
+			LimitedSupport:  false,
+			Hibernating:     true,
+			FailingSyncSets: "selectorsyncset1 syncset1 ",
+			ErrorMessage:    "Failed to sync selectorsyncset1\n\nFailed to sync syncset1\n\n",
+		},
+	}
```
btoll, line 179: Can you either add more failing `ClusterSync`s to test here or remove the `for` loop?

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,270 @@
...
+			}
+
+			failingClusterSyncList := []failingClusterSync{
+				{Name: "zeta", Timestamp: "2023-01-01T00:00:00Z", FailingSyncSets: "syncset3"},
+				{Name: "alpha", Timestamp: "2022-01-01T00:00:00Z", FailingSyncSets: "syncset1"},
+				{Name: "beta", Timestamp: "2023-02-01T00:00:00Z", FailingSyncSets: "syncset2"},
+			}
+
+			err := options.sortBy(failingClusterSyncList)
+			if tt.expectedErr != "" {
```
btoll, line 261: Instead of testing for `expectedErr`, why not remove that from the inline `struct` and just test the return `err` value from `sortBy`?

cmd/hive/clustersync_test.go
drakshakhan: I understand your point, and I agree testing the Cobra API might seem redundant. However, since it uses several internal functions, I wrote the test cases for the API itself to avoid writing separate tests for each internal function. This keeps things simpler and more manageable while still covering all the necessary functionality.

cmd/hive/clustersync_test.go
drakshakhan: added the test case as requested.

cmd/hive/clustersync_test.go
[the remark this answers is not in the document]
drakshakhan: formatted the objects as per the suggestion.

cmd/hive/clustersync_test.go
drakshakhan: Thanks for the suggestion. I opted to keep the function separate to make the code more modular and easier to read and maintain. I felt that keeping the logic in a separate function helps in simplifying the test and making it more manageable.

cmd/hive/clustersync_test.go
drakshakhan: I’ve made the changes and removed the expectedErr from the inline struct. Now, the test directly checks the returned err value from sortBy. This should make the test more focused and simpler.

cmd/hive/clustersync_test.go
btoll: I still think this should be removed.  If there are other functions that particular to the package, can they be tested with testing the cobra API.

.gitignore
```diff
@@ -26,3 +26,4 @@ osdctl
 # goReleaser 
 dist/
 bin/
+vendor/
```
btoll, line 29: Should we be ignoring `vendor`?

drakshakhan: Apologies, that was a mistake on my end. I’ll remove the changes .

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,396 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	"github.com/openshift/hive/apis/hiveinternal/v1alpha1"
+	mockk8s "github.com/openshift/osdctl/cmd/hive/clusterdeployment/mock/k8s"
+)
+
+// Mock data for ClusterDeployment and ClusterSync that are used across all test cases
+var (
+	cdList = hivev1.ClusterDeploymentList{
```
btoll, line 25: It looks like these global vars are only being used by the `setupMockClient` function.  Can you either inline them or put them in a private function that the `setupMockClient` would call and have the lists returned to it?

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,396 @@
...
+				return
+			}
+
+			assert.NoError(t, err)
+
+		})
+	}
+}
+
+func testListFailingClusterSyncs(t *testing.T, returnErr error, expectedResults []failingClusterSync, isEmpty bool) {
```
btoll, line 209: I still feel this should be inlined in `TestListFailingClusterSyncs`, but if you want to keep it as a helper function could you rename it to something else?  Someone could easily think that this should be a unit test with a lower-case `t` type and then uppercase it.   Having it the same case-insensitive name as the actual unit test is a bit confusing.

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,396 @@
...
+
+	setupMockClient(mockClient, returnErr, isEmpty)
+
+	options := &clusterSyncFailuresOptions{
+		kubeCli: mockClient,
+	}
+
+	result, err := options.listFailingClusterSyncs()
+
+	if returnErr != nil {
```
btoll, line 223: This isn't really testing anything.  Currently, it's checking the value of the inline struct, but it's not a value that has been processed or returned by any package functions.

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,396 @@
...
+	} else {
+		assert.NoError(t, err)
+		if isEmpty {
+			// If isEmpty is true, the result should be an empty slice
+			assert.Len(t, result, 0)
+		} else {
+			assert.NotNil(t, result)
+			assert.Len(t, result, len(expectedResults))
+			for i, expected := range expectedResults {
+				assert.Equal(t, expected.Name, result[i].Name)
```
btoll, line 235: The `expectedResult` is only ever going to be length 1, so can we either add more results to the slice or remove this `for` loop?

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,396 @@
...
+					Status: v1alpha1.ClusterSyncStatus{
+						SyncSets: []v1alpha1.SyncStatus{{
+							Name:           "sync-failure",
+							Result:         "Failure",
+							FailureMessage: "Some error occurred",
+						}},
+					},
+				}},
+			},
+			expectError: false,
```
btoll, line 163: The verbiage on this test is confusing, b/c it's a failure state but `expectError` is set to `false`.

Having looked at the package code, it's unfortunate that the function being tested isn't returning the `FailureMessage` so that we can check it.

Because of that, perhaps we shouldn't have a `Result: "Failure"` because it's confusing when reading this (and not looking at the code) why we wouldn't be expecting an error.

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,396 @@
...
+
+			o := &clusterSyncFailuresOptions{
+				kubeCli:   client,
+				clusterID: "test",
+			}
+
+			err := o.printFailingCluster()
+			if tt.expectError {
+				assert.Error(t, err)
+				return
```
btoll, line 200: @christophermancini  There's nothing wrong with `else` clauses depending upon the context.  The pattern that I've seen in other (most?) packages is that the reason that we are checking for the  `expectError` field along with the returned error from the function being tested is because there are states where `err` is `nil` but there are error messages that need to be checked and other cases where there needs to be tighter control over different error states.

It reads better and it is easier to grok the pattern and importantly its intent when there is also an `else` clause.

cmd/hive/clustersync_test.go
drakshakhan: Renamed the helper function to avoid confusion.

cmd/hive/clustersync_test.go
drakshakhan: Removed the for loop.

cmd/hive/clustersync_test.go
drakshakhan: I'm keeping this test to ensure the `printFailingCluster ` method correctly prints failure messages. While it doesn’t return errors, the test verifies that the appropriate messages are printed for failure scenarios. Added a comment that clarifies that failures are handled by printing, not returning errors.

cmd/hive/clustersync_test.go
drakshakhan: The test uses mock data because `listFailingClusterSyncs` makes an actual API call, which would return nil data in a test environment. To avoid dependencies on third-party APIs and ensure consistent results, I used a mock client to simulate expected responses. The test covers the function under test and provides good coverage, verifying that it correctly processes the mock data and transforms it into the expected output.

btoll: The test is fine, I'm talking about `returnErr`.

cmd/hive/clustersync_test.go
btoll: @drakshakhan Please don't resolve these, it's difficult to find them again :)

You're correct, but this test isn't able to verify the printed messages.  Because of that, it's confusing that it's purportedly testing an error but `expectErr` is `false`.

drakshakhan: I understand your concern. I’ve added a comment clarifying that the function prints errors instead of returning them. Would that be sufficient, or do you think it would be better to remove this test case altogether?

cmd/hive/clustersync_test.go
drakshakhan: I see your point about the check not validating a processed value. I’m wondering if the following condition would be a better approach since err and result are both returned by the package function.
This ensures we are verifying the actual function output instead of an inline struct. Let me know your thoughts.

drakshakhan:
@btoll, @joshbranham, @dustman9000 
Just a reminder. Please approve if no other comments. 
CC - @rpodishe

cmd/hive/clustersync_test.go
```diff
@@ -0,0 +1,377 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	"github.com/openshift/hive/apis/hiveinternal/v1alpha1"
+	mockk8s "github.com/openshift/osdctl/cmd/hive/clusterdeployment/mock/k8s"
+)
+
+func TestPrintFailingCluster(t *testing.T) {
```
petrkotas, line 23: This one is great, it covers all the errors that can arise. May I ask you to further improve it and make it check the expected output on the stdout? 

To give you a hint. One way you can do this is to look at https://pkg.go.dev/k8s.io/cli-runtime@v0.32.1/pkg/genericclioptions#IOStreams which is already used in clustersync.go you can extend the fmt.printLn to look something like 

```go
fmt.Fprintln(o.IOStreams.Out, "Cluster Name:", clusterDeployment.ObjectMeta.Name)
```

This uses the existing setup of steams alreay in place. 

Furthermore you can extend the testing clusterSyncFailureOptions with the streams like 

```
o := &clusterSyncFailuresOptions{
	kubeCli:   client,
	clusterID: "test",
	IOStreams: genericiooptions.IOStreams{
		Out:    stdout,
		ErrOut: stdout,
		In:     nil,
	},
}
```

with this, you can assert the outputs.

petrkotas (changes requested):
Hi @drakshakhan thank you for the PR, the code is sound. I appreciate you keep the style the same as the rest of the codebase. 

I have one ask and that is to extend the print test to check the stdout. I provided a hint how this can be achieved using the mechanism already in place. Thank you

petrkotas (approved):
@drakshakhan thank you, this is exactly what I was thinking.  /lgtm

PR #696, opened 2025-03-19
Add unit tests for servicelog in post_test.go

lambasanchit:
This PR significantly enhances the test coverage for the post.go file in Servicelog, increasing the overall coverage by approximately 50%. To achieve this, additional unit tests were added for various functions in the post file, written in post_test.go. These new tests ensure the correct functionality of the methods defined in post.go. 

User Story -->   https://issues.redhat.com/browse/OSD-28412

~/git/osdctl/cmd/servicelog$   go test -coverprofile=coverage.out
go tool cover -html=coverage.out

As a result, the code coverage for the servicelog package has increased to 48%. All tests have passed successfully, confirming that the newly added tests function as expected and contribute to improved overall test coverage.

### DIFF
--- a/cmd/servicelog/post_test.go
+++ b/cmd/servicelog/post_test.go
@@ -284,7 +284,7 @@ func TestParseUserParameters(t *testing.T) {
 		expectValues []string
 	}{
 		{
-			name:         "Valid Parameters",
+			name:         "valid_parameters",
 			input:        []string{"FOO=BAR", "BAZ=QUX"},
 			expectNames:  []string{"${FOO}", "${BAZ}"},
 			expectValues: []string{"BAR", "QUX"},
@@ -311,7 +311,7 @@ func TestPrintTemplate(t *testing.T) {
 		expected error
 	}{
 		{
-			name: "Valid input with no errors",
+			name: "valid_input_with_no_errors",
 			input: &PostCmdOptions{
 				Message: servicelog.Message{
 					Severity:      "info",
@@ -342,7 +342,7 @@ func TestPrintClusters(t *testing.T) {
 		expected error
 	}{
 		{
-			name: "Valid clusters input with no errors",
+			name: "valid_clusters_input_with_no_errors",
 			input: func() []*v1.Cluster {
 				clusters := []*v1.Cluster{}
 				cluster1, _ := v1.NewCluster().ExternalID("cluster-1").Build()
@@ -395,12 +395,12 @@ func TestParseClustersFile(t *testing.T) {
 		expected error
 	}{
 		{
-			name:     "Valid JSON input with no errors",
+			name:     "valid_JSON_input_with_no_errors",
 			input:    []byte(`{"clusters":["cluster-1","cluster-2"]}`),
 			expected: nil,
 		},
 		{
-			name:     "Invalid JSON input with errors",
+			name:     "invalid_JSON_input_with_errors",
 			input:    []byte(`{"clusters":["cluster-1","cluster-2"`),
 			expected: assert.AnError,
 		},
@@ -431,7 +431,7 @@ func TestReplaceFlags(t *testing.T) {
 		expectedFilter string
 	}{
 		{
-			name: "Valid Flag Replacement in Message",
+			name: "valid_flag_replacement_in_Message",
 			inputOptions: PostCmdOptions{
 				Message: servicelog.Message{Summary: "This is a FILTERREPLACE test"},
 			},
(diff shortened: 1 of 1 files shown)


cmd/servicelog/post_test.go
```diff
@@ -107,4 +114,589 @@ var _ = Describe("Test posting service logs", func() {
...
+			tmpfile, err := os.CreateTemp("", "example")
+			Expect(err).ShouldNot(HaveOccurred())
+
+			defer os.Remove(tmpfile.Name())
+
+			_, err = tmpfile.Write(content)
+			Expect(err).ShouldNot(HaveOccurred())
+
+			err = tmpfile.Close()
+			Expect(err).ShouldNot(HaveOccurred())
```
Amarthya-v, line 188: Overall it looks good, just Instead of explicitly closing the temporary file at the end, use defer tmpfile.Close() immediately after creation. This ensures the file is always closed, even if an assertion fails before reaching tmpfile.Close().

Amarthya-v: You're right, @christophermancini! my bad, Using defer tmpfile.Close()  we wouldn't be able to verify that Close() didn't return an error.

cmd/servicelog/post_test.go
```diff
@@ -107,4 +114,589 @@ var _ = Describe("Test posting service logs", func() {
...
+		inputOptions PostCmdOptions
+	}{
+		{
+			name: "No clusters",
+			inputOptions: PostCmdOptions{
+				successfulClusters: map[string]string{},
+				failedClusters:     map[string]string{},
+			},
+		},
+		/*		{
```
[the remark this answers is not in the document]
lambasanchit: Ok ...  only intention was to if in future fatal is changed to error or panic, we could use commented cases. 
As of now I have removed the comments.

cmd/servicelog/post_test.go
```diff
@@ -107,4 +114,589 @@ var _ = Describe("Test posting service logs", func() {
...
+			name: "No filter files specified",
+			options: PostCmdOptions{
+				filterFiles: []string{}, // No filter files
+			},
+			expectedFilter: "",
+		},
+		{
+			name: "One filter file",
+			options: PostCmdOptions{
+				filterFiles: []string{"filter1.txt"},
```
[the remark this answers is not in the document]
lambasanchit: I have updated the test according to your comment .

cmd/servicelog/post_test.go
```diff
@@ -107,4 +114,589 @@ var _ = Describe("Test posting service logs", func() {
...
+}
+
+func TestListMessagedClusters(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    map[string]string
+		expected error
+	}{
+		{
+			name:     "Valid clusters input with no errors",
```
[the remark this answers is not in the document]
lambasanchit: i have made the changes as per suggestion please review

cmd/servicelog/post_test.go
```diff
@@ -107,4 +113,564 @@ var _ = Describe("Test posting service logs", func() {
...
+		})
+	})
+
+	Context("newPostCmd function", func() {
+		It("creates a new post command", func() {
+			cmd := newPostCmd()
+			Expect(cmd).NotTo(BeNil())
+			Expect(cmd.Use).To(Equal("post CLUSTER_ID"))
+			Expect(cmd.Short).To(Equal("Post a service log to a cluster or list of clusters"))
+		})
```
joshbranham, line 158: I am dubious on these kinds of tests being needed here. Do we think there is value in validating a string we set is actually set in Cobra? We need to weigh the cost/benefit of these kinds of tests. For example, any change to the usage text requires updating in tests, which is maybe fine, but it's not really protecting us from any regression or user error.

lambasanchit: yes, we can remove the test as its only testing if values are getting set in cobra.command, it was earlier written just to increase coverage.

joshbranham: Right, our goal should not be "increase coverage". Our goal should be to write meaningful tests that are easy to maintain and provide value. Increasing test coverage just to do it leads to all sorts of problems. I should be able to open the test file for a given package or file and see the top business cases being tested, not have to skim through dozens of tests.

lambasanchit: I agree with your feedback and have removed the test function as mentioned in the last comment

joshbranham (commented):
/lgtm

PR #697, opened 2025-03-19
OSD-28325: Added test cases for access.go

drakshakhan:
This PR significantly enhances test coverage for the access.go file in the access package, increasing overall coverage by 54%. Additional unit tests have been introduced to validate the functionality of various methods, ensuring robust and reliable behavior.

Verification:
All tests pass successfully, confirming the correctness of the newly added test cases.

Run the following commands to verify coverage:
**go test -coverprofile=coverage.out ./...** to run the tests and generate coverage information.
**go tool cover -html=coverage.out** to visualize the coverage and ensure completeness.

This improvement strengthens the reliability of the access package and contributes to better maintainability.

Jira story - https://issues.redhat.com/browse/OSD-28325


cmd/cluster/access/access_test.go
```diff
@@ -361,3 +362,156 @@ func generateKubeconfigSecretObjectForTesting(name, namespace, key, serverURL st
...
+
+	tests := []struct {
+		Name          string
+		Namespace     corev1.Namespace
+		SecretList    []corev1.Secret
+		ExpectedError bool
+		ExpectedName  string
+	}{
+		{
+			Name: "Successfully retrieve kubeconfig secret",
```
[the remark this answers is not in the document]
drakshakhan: Done the changes.

cmd/cluster/access/access_test.go
```diff
@@ -361,3 +362,156 @@ func generateKubeconfigSecretObjectForTesting(name, namespace, key, serverURL st
...
+}
+
+func TestNewCmdAccess(t *testing.T) {
+	tests := []struct {
+		name        string
+		args        []string
+		expectError bool
+	}{
+		{
+			name:        "invalid number of arguments",
```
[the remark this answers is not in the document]
drakshakhan: Done the changes.

cmd/cluster/access/access_test.go
```diff
@@ -361,3 +362,156 @@ func generateKubeconfigSecretObjectForTesting(name, namespace, key, serverURL st
...
+}
+
+func TestAccessCmdComplete(t *testing.T) {
+	tests := []struct {
+		name        string
+		args        []string
+		expectedErr bool
+	}{
+		{
+			name:        "valid cluster identifier",
```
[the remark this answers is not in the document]
drakshakhan: Done the changes.

cmd/cluster/access/access_test.go
```diff
@@ -361,3 +362,156 @@ func generateKubeconfigSecretObjectForTesting(name, namespace, key, serverURL st
...
+			secret, err := access.getKubeConfigSecret(test.Namespace)
+
+			// If error is expected, check if an error occurred
+			if test.ExpectedError {
+				if err == nil {
+					t.Errorf("Expected an error but got none")
+				}
+				if secret.Name != test.ExpectedName {
+					t.Errorf("Expected an empty secret, but got: %s", secret.Name)
+				}
```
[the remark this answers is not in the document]
drakshakhan: Thanks for the suggestion! I tried using return to exit early, but if I remove the else block and use another if, it causes errors because both conditions would be evaluated independently. Keeping the else ensures the logic flows correctly without running into issues. Let me know if you have any other thoughts!

cmd/cluster/access/access_test.go
```diff
@@ -361,3 +362,158 @@ func generateKubeconfigSecretObjectForTesting(name, namespace, key, serverURL st
...
+			err := corev1.AddToScheme(scheme)
+			if err != nil {
+				t.Fatalf("Failed to add corev1 to scheme: %v", err)
+			}
+
+			client := k8s.NewFakeClient(fake.NewClientBuilder().WithScheme(scheme))
+			streams := genericclioptions.IOStreams{In: genericclioptions.NewTestIOStreamsDiscard().In, Out: os.Stdout, ErrOut: os.Stderr}
+			access := newClusterAccessOptions(client, streams)
+
+			secret, err := access.getKubeConfigSecret(test.Namespace)
```
btoll, line 427: As it stands, this method will always return a generic `secret` object and an `error` for each of the tests because there isn't a secret that matches the expected label:
https://github.com/drakshakhan/osdctl/blob/OSD-28325/cmd/cluster/access/access.go#L356-L358
https://github.com/drakshakhan/osdctl/blob/OSD-28325/cmd/cluster/access/access.go#L346

To test this, it would be good to create a secret with the expected label and make sure that it is returned from the `getKubeConfigSecret` method.

In addition, I don't think the current tests are providing much value b/c the `secret.Name` and the `test.ExpectedName` will always be an empty string.

cmd/cluster/access/access_test.go
```diff
@@ -361,3 +362,158 @@ func generateKubeconfigSecretObjectForTesting(name, namespace, key, serverURL st
...
+			name:        "too_many_arguments",
+			args:        []string{"cluster123", "extra-arg"},
+			expectedErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cmd := &cobra.Command{}
+			err := accessCmdComplete(cmd, tt.args)
```
btoll, line 510: We shouldn't be testing the `cobra` API.  I realize that this is testing an existing function in `access.go`, but that test is unnecessary as well because the `cobra` API is explicitly setting the number of args to 1:
https://github.com/drakshakhan/osdctl/blob/OSD-28325/cmd/cluster/access/access.go#L58

cmd/cluster/access/access_test.go
```diff
@@ -361,3 +362,158 @@ func generateKubeconfigSecretObjectForTesting(name, namespace, key, serverURL st
...
+				}
+				if secret.Name != test.ExpectedName {
+					t.Errorf("Expected secret '%s', but got: %s", test.ExpectedName, secret.Name)
+				}
+			}
+		})
+	}
+}
+
+func TestNewCmdAccess(t *testing.T) {
```
btoll, line 452: We shouldn't be testing the `cobra` API.

drakshakhan: Removed the test case for cobra api.

cmd/cluster/access/access_test.go
drakshakhan: Removed the test case for cobra api.

cmd/cluster/access/access_test.go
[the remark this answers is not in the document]
drakshakhan: It was a miss on my part. I had tried something else by placing the return inside the nested conditions, which caused the issue. Moving it outside the nested conditions, as you suggested, works perfectly now. Thanks for pointing that out.

cmd/cluster/access/access_test.go
drakshakhan: Thanks for the feedback. I've updated the test cases to ensure a secret with the expected label is created and verified in `getKubeConfigSecret`. Now, the test includes a properly labeled secret and validates that it is returned correctly.

drakshakhan:
@joshbranham, @Tafhim 
Just a reminder. Please approve if no other comments.
CC - @rpodishe

petrkotas (approved):
Hi, the test does what it promises. The current version is ok. /lgtm

PR #699, opened 2025-03-21
Osd 28496 pkg jumphost

lambasanchit:
This PR significantly enhances the test coverage for the cmd.go file and create.go  in Jumphost, increasing the overall coverage by approximately 50%. 

User Story --> https://issues.redhat.com/browse/OSD-28496

All tests have passed successfully, confirming that the newly added tests function as expected and contribute to improved overall test coverage.


go.mod
```diff
@@ -235,6 +235,7 @@ require (
 	github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966 // indirect
 	github.com/sourcegraph/conc v0.3.0 // indirect
 	github.com/spf13/cast v1.7.1 // indirect
+	github.com/stretchr/objx v0.5.2 // indirect
```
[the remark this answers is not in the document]
lambasanchit: This dependency is for 
github.com/stretchr/testify/mock
github.com/stretchr/objx 

its used here : args := m.Called(ctx, params) Called takes arguments as interfaces, I think m.Called must be using it for equality checks for arguments.

cmd/jumphost/cmd_test.go
```diff
@@ -0,0 +1,97 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := validateCluster(tt.cluster)
+			if tt.expectError {
+				assert.Error(t, err)
+				assert.Equal(t, tt.errorMsg, err.Error())
+			} else {
```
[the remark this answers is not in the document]
lambasanchit: yes, we can do that as its anonymous inline, I have updated it using return rather than if/else condition

cmd/jumphost/cmd_test.go
```diff
@@ -0,0 +1,92 @@
...
+			}(),
+		},
+		{
+			name: "non_aws_cloud_provider",
+			cluster: func() *cmv1.Cluster {
+				cluster1, _ := cmv1.NewCluster().CloudProvider(cmv1.NewCloudProvider().ID("gcp")).AWS(cmv1.NewAWS()).Build()
+				return cluster1
+			}(),
+			errorMsg: "only supports aws, got gcp",
+		},
```
btoll, line 79: Could you add some other tests here like:
```
cluster1, _ := cmv1.NewCluster().CloudProvider(cmv1.NewCloudProvider().ID("gcp")).GCP(cmv1.NewGCP()).Build()
```
and
```
and creating an STS cluster so it will fail?

File: cmd/jumphost/cmd_test.go [hunk 4]
```diff
@@ -0,0 +1,92 @@
...
+			name: "multiple_tags",
+			tags: []types.Tag{
+				{Key: aws.String("App"), Value: aws.String("Backend")},
+				{Key: aws.String("Team"), Value: aws.String("DevOps")},
+			},
+			expected: []types.Filter{
+				{Name: aws.String("tag:App"), Values: []string{"Backend"}},
+				{Name: aws.String("tag:Team"), Values: []string{"DevOps"}},
+			},
+		},
```
btoll, line 43: Can you create some failure tests?

cmd/jumphost/cmd_test.go
lambasanchit: Added cases for provider AWS with STS and GCP

cmd/jumphost/cmd_test.go
lambasanchit: Added with nil values creating nil pointer exception, handled panic in test case using recover.

rpodishe:
@btoll, @iamkirkbater can you please approve this PR, if no other comments?
@lambasanchit

cmd/jumphost/create_test.go
```diff
@@ -0,0 +1,327 @@
...
+				assert.NoError(t, err)
+				assert.Equal(t, tt.expectedAMI, ami)
+			}
+
+			mockClient.AssertExpectations(t)
+		})
+	}
+}
+
+func TestAllowJumphostSshFromIp(t *testing.T) {
```
petrkotas, line 98: This test is sensitive for internet connection, due to the use of ip, err := determinePublicIp(). However due to the nature of the code, it work well.
So I have question for other reviewers. Is exiting without error when not able to connect to internet ok?

petrkotas (approved):
Hi @lambasanchit thank you for the PR. I appreciate that it is consisent with the pre-existing pattern. The tests are structured and do exactly what was asked.

I have two nits related to use of non mockable function: http.get and os.createTemp. These are meant as a food for thought, and not as a blocker. The change would require code refactoring, which is not worth it imho. 

So /lgtm from me.

cmd/jumphost/create_test.go
```diff
@@ -0,0 +1,327 @@
...
+				assert.Contains(t, err.Error(), tt.errorContains)
+			} else {
+				assert.NoError(t, err)
+			}
+			mockClient.AssertExpectations(t)
+		})
+	}
+}
+
+func TestDeterminePublicIpReal(t *testing.T) {
```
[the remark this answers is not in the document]
lambasanchit: I think we can remove it as it making actual calls . According to me it doesnt make sense to wrap this method with interface as its only making get call and determining clients IP

cmd/jumphost/create_test.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.expected, tt.jumphost.assembleNextSteps())
+		})
+	}
+}
+
+func TestCreateKeyPair(t *testing.T) {
```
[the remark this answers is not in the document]
lambasanchit: Lets not refactor the code ,In future , if we feel like changing the approach then we can refactor the test as well. But as of now lets continue with the test or if you think the test can break ci/cd pipeline or any deployment then we can remove it .

typeid (approved):
/lgtm
/approve

PR #706, opened 2025-03-26
OSD-28399 : Added unit test case for context and cluster under org pkg

amitupadh:
Description
Develop unit test cases to ensure the reliability and correctness of the cluster and context functionalities within the org package. The objective is to validate that these components perform as intended under various scenarios, thereby enhancing code quality and maintainability.

Code Refactoring
    Replaced direct function calls with interface-based dependencies
    Implemented mock implementations for external services (OCM, Jira, PagerDuty)
    Added dependency injection support for all external integrations

Coverage Details
Increased unit test coverage for cluster and context commands.

Verification

Jira Ticket
https://issues.redhat.com/browse/OSD-28399


cmd/org/context_test.go
```diff
@@ -0,0 +1,340 @@
...
+	origSearchSubs := searchAllSubscriptionsByOrgFunc
+	origCreateConn := createConnectionFunc
+	origAddPDAlerts := addPDAlertsFunc
+	origAddLimitedSupportReasons := addLimitedSupportReasonsFunc
+	origAddServiceLogs := addServiceLogsFunc
+	origAddJiraIssues := addJiraIssuesFunc
+
+	// Restore original functions after test
+	defer func() {
+		getClusterFunc = origGetCluster
```
Amarthya-v, line 213: just a nit: Consider injecting dependencies via a struct or interface instead of relying on globals like getClusterFunc. This will improve test isolation and help prevent potential flakiness in parallel test runs.

amitupadh: @Amarthya-v 
Fixed as per suggestion
Please review again

cmd/org/clusters_test.go
```diff
@@ -0,0 +1,211 @@
...
+				assert.NotNil(t, result)
+				if tt.validate != nil {
+					tt.validate(t, result)
+				}
+			}
+		})
+	}
+}
+
+func TestPrintClusters_OutputFormat(t *testing.T) {
```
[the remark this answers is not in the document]
amitupadh: @christophermancini 
Fixed as per your suggestion, can you please review it again.

cmd/org/context.go
```diff
@@ -24,303 +26,293 @@ import (
...
 }
 
-func addLimitedSupportReasons(clusterInfo *ClusterInfo, ocmClient *sdk.Connection) error {
-	var limitedSupportReasonsErr error
-	clusterInfo.LimitedSupportReasons, limitedSupportReasonsErr = utils.GetClusterLimitedSupportReasons(ocmClient, clusterInfo.ID)
-	if limitedSupportReasonsErr != nil {
-		return fmt.Errorf("failed to fetch limited support reasons for cluster %v: %w", clusterInfo.ID, limitedSupportReasonsErr)
-	}
-	return nil
+type ClusterInfo struct {
```
bergmannf, line 219: Nitpick - but I think having the new types at the top of the file is generally nicer, but if everyone else is fine having them here, we can also keep them here.

cmd/org/context.go
```diff
@@ -24,303 +26,293 @@ import (
...
+	GetServiceLogs      func(string, time.Time, bool, bool) ([]*v1.LogEntry, error)
+	GetJiraIssues       func(clusterID, externalID, filter string) ([]jira.Issue, error)
+	NewPDClient         func(baseDomain string) (PDClient, error)
 }
 
-func init() {
-	contextCmd.Flags().StringP("output", "o", "", "output format for the results. only supported value currently is 'json'")
+type PDClient interface {
+	GetPDServiceIDs() ([]string, error)
+	GetFiringAlertsForCluster([]string) (map[string][]pd.Incident, error)
```
bergmannf, line 49:
```suggestion
	GetFiringAlertsForClusters([]string) (map[string][]pd.Incident, error)
```

If the function is name `..ForCluster` I would expect the argument to be a single clusterid not multiple.

bergmannf:
Not a comment to fix anything else, but I wanted to call out - nice work on refactoring `context.go` in a way to make it testable - took me a while to go through the file, but I think this looks good.
I think removing the `ctx.Done()` call in the Context fetching should also be fine as we never really used the channel anyhow.

cmd/org/context.go
amitupadh: Hey @bergmannf 

Thanks for the suggestion! I initially renamed GetFiringAlertsForCluster to GetFiringAlertsForClusters to better reflect the parameter being a slice. However, this method is part of the pagerduty.client in an external package.

Since PDClient is implemented by that external type, changing the method name in the interface causes a mismatch and breaks compilation.

To keep things consistent and compilable, I’ll revert the method name in the interface back to GetFiringAlertsForCluster.

cmd/org/context.go
amitupadh: @bergmannf 
Move `type ClusterInfo struct`  to the top of the file, ideally after the import block, before any functions or logic.

amitupadh:
@bergmannf @iamkirkbater @typeid 
I’ve incorporated the suggested changes — thanks again for the feedback!
When you have a moment, could you please review the latest updates?
If everything looks good, I’d appreciate your approval.

amitupadh:
@iamkirkbater @typeid 
please approve if no comments

amitupadh:
@iamkirkbater @typeid
please approve if no comments

joshbranham:
@amitupadh I approved and this merged last week :)

PR #707, opened 2025-03-27
OSD: 27238 - osdctl controlplane resize should not wait for completion

ratnam915:
As a part of https://issues.redhat.com/browse/OSD-27238, the directive was to ensure when osdctl cluster controlplane resize is being used with cpms, a service log is sent and we exit out instead of waiting for the resize to be completed.

Changes have been made in the code to ensure that we send a service log and exit out instead of waiting for the resize to be completed.


joshbranham (commented):
/lgtm

PR #720, opened 2025-04-11
OSD-29082: Test cases for dynatrace and saas for promote

karanjitsingh01:
This PR contains test for promote/dynatrace and promote/saas packages. 

Most of the functions were running git commands, like git checkout or git pull, these can't be mocked as the logic is inside the function. 

we took the approach of executing real git commands inside a temporary directory (t.TempDir()) to simulate a realistic Git environment.
Approach for test: - 
Create a temporary directory with t.TempDir(), Its a temporary directory so should be deleted after the completion of test
Initialize a Git repo using git init and run all required commands for testing in setup(). 
Call function in test.


Makdaam:
cc @typeid - requires approval

cmd/promote/dynatrace/dt_app_interface.go
```diff
@@ -36,27 +37,30 @@ func DefaultAppInterfaceDirectory() string {
 
 func BootstrapOsdCtlForAppInterfaceAndServicePromotions(appInterfaceCheckoutDir string) AppInterface {
 	a := AppInterface{}
+	a.GitExecutor = iexec.Exec{}
 	if appInterfaceCheckoutDir != "" {
 		a.GitDirectory = appInterfaceCheckoutDir
-		err := checkAppInterfaceCheckout(a.GitDirectory)
+		a.GitExecutor = iexec.Exec{}
```
joshbranham, line 43: Why are we setting `a.GitExecutor` twice, once on line 40 and once on 43?

cmd/promote/dynatrace/dt_git_cmd.go
```diff
@@ -14,21 +14,21 @@ var (
...
 		}
-
 		BaseDir = strings.TrimSpace(string(baseDirOutput))
 	})
 
 	return BaseDir, baseDirErr
 }
 
+/*
+// commented out as not used
```
joshbranham, line 31: So we should remove it :)

PR #721, opened 2025-04-11
Osd-29080 Added unit test case for List.go file and Get.go file under Cost Package 

lambasanchit:
This PR significantly enhances the test coverage for the Cost Package around 50%.

User Story --> https://issues.redhat.com/browse/OSD-29080

go test -coverprofile=coverage.out
go tool cover -html=coverage.out

All tests have passed successfully, confirming that the newly added tests function as expected and contribute to improved overall test coverage.


cmd/cost/get_test.go
```diff
@@ -0,0 +1,183 @@
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			start, end := getTimePeriod(&tc.timePtr)
+			require.Equal(t, tc.expectedStart, start)
+			require.Equal(t, tc.expectedEnd, end)
+		})
+	}
+}
+
+func Test_printCostGet(t *testing.T) {
```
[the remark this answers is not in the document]
lambasanchit: okay Chris , actual function is using fmt.Println  i:e i was trying to fetch output from console . As of now i have removed that part and checking only for errors .

cmd/cost/get_test.go
```diff
@@ -0,0 +1,117 @@
...
+			expectedStart: fmt.Sprintf("%d-%02d-%02d", refTime.Year(), 01, 01),
+			expectedEnd:   fmt.Sprintf("%d-%02d-%02d", refTime.Year(), refTime.Month(), refTime.Day()),
+		},
+		{
+			name:          "1Y",
+			timePtr:       "1Y",
+			expectedStart: refTime.AddDate(-1, 0, 0).Format("2006-01-02"),
+			expectedEnd:   fmt.Sprintf("%d-%02d-%02d", refTime.Year(), refTime.Month(), refTime.Day()),
+		},
+	}
```
btoll, line 57: We should also be testing for expected failures, such as:
-  `asdfadsfdd`
- `mtd` (test for case insensitivity.  should the method being tested automatically uppercase the flag?) 
- "3Y"
- etc

cmd/cost/get_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			start, end := getTimePeriod(&tc.timePtr)
```
btoll, line 61: Can you test for all existing cases?

cmd/cost/get_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/organizations/types"
+	"github.com/onsi/gomega"
+	"github.com/shopspring/decimal"
+	"github.com/stretchr/testify/require"
+)
+
+var mockNowFunc = time.Now
```
btoll, line 15: Is this being used anywhere?

cmd/cost/get_test.go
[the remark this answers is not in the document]
btoll: I agree with @christophermancini that it would be great to refactor this to have it return a byteslice so the method can be properly tested.
In light of doing that, I'm not sure there is much value in the current tests, because it's testing if the `json` and `yaml` packages return an error.  Perhaps they should just be removed.

cmd/cost/list_test.go
```diff
@@ -0,0 +1,461 @@
...
+			} else {
+				assert.NoError(t, err)
+				assert.True(t, sum.Equal(tt.expected), "expected %v, got %v", tt.expected, sum)
+			}
+		})
+	}
+}
+
+func TestListCostsUnderOU(t *testing.T) {
+	g := gomega.NewGomegaWithT(t)
```
btoll, line 140: This method is deprecated.

cmd/cost/get_test.go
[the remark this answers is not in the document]
lambasanchit: @btoll Shall i refactor the code and update the test accordingly

cmd/cost/get_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	tests := []struct {
+		name          string
+		timePtr       string
+		expectedStart string
+		expectedEnd   string
+	}{
+		{
+			name:          "Default",
+			timePtr:       "",
```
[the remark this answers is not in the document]
lambasanchit: @btoll this is just to test  the behaviour of  any random value  for the switch case

cmd/cost/list_test.go
lambasanchit: Now i have removed this function

cmd/cost/get_test.go
lambasanchit: removed

cmd/cost/get_test.go
lambasanchit: addded the test cases

cmd/cost/get_test.go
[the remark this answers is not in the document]
btoll: Do you mind renaming it?  The name "Default" sounds like it should map to a default value in the `switch`.

cmd/cost/get_test.go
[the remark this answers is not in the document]
btoll: If you have time.  If not, I'd just remove this test.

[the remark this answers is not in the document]
lambasanchit: I have refactor the function and update the test cases

cmd/cost/get_test.go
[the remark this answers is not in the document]
lambasanchit: @btoll i have changed

cmd/cost/get.go
```diff
@@ -124,11 +124,12 @@ func (o *getOptions) run() error {
 		}
 	}
 
-	err = o.printCostGet(cost, unit, o, OU)
+	print, err := o.printCostGet(cost, unit, o, OU)
```
joshbranham, line 127: One alternative here to make this even cleaner is to pass an `io.Writer` to `printCostGet()`, which means in normal code `os.Stdout` but under test you can pass a `bytes.Buffer` and then inspect the buffer for the values. Then you don't need this indirection to print a string returned from the func. Thoughts?

lambasanchit: @joshbranham  if we introduce io.writer in printCostGet then we also need to pass io.writer to outputflag.PrintResponse(o.output, resp) and i hve seen outputflag.PrintResponse being called at mutiple files.
Shall we do it or as printCostGet  is only used for logging , returning the print statement can also be an option.
Just let me know if we can go ahaed with return or io.writer then i can change it in outputflag.PrintResponse .

joshbranham: If you don't want to change all callers to pass an `io.Writer` then that is fine, but can we at least refactor this a bit to make more sense? As proposed, `printCostGet` returns a string that you then _print_ later. So maybe that should be named something like `getCostOutput()`.

lambasanchit:
@joshbranham    Please Approve the PR if no other comments

joshbranham (commented):
/lgtm

PR #722, opened 2025-04-14
OSD-29157: Added test cases for hypershift_info.go

drakshakhan:
This PR adds test cases for the hypershift_info.go in cluster package. The following changes were made:

Test Case Addition
Added unit tests to thoroughly test the functionality in hypershift_info.go 

Test Frameworks Used

Test Execution

Verification

Jira Story - https://issues.redhat.com/browse/OSD-29157

### DIFF
--- a/cmd/cluster/hypershift_info_test.go
+++ b/cmd/cluster/hypershift_info_test.go
@@ -1,0 +1,1389 @@
+package cluster
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+
+	aws1 "github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/aws/aws-sdk-go-v2/service/elasticloadbalancingv2"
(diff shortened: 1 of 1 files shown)


bergmannf:
I checked the code before actually checking the tests, and I don't think adding the tests is useful at this point in time:

* The code in place is not paginating a lot of calls that (at the time) probably did not return too many objects, but might now do this.
* So the code probably would need to be updated first to paginate all AWS calls.

Then most of these tests seem to mock AWS calls, which I think we are fine to not test - this is an external library, let's except this to be ok.
So the only thing to actually test would be checking that the right results are returned: e.g. the right VPCs and the right VPCEs in the customer account for thsoe - are found for a cluster, but there it would also make more sense to update the code in-place first, as it's not filtering well on those calls making them way less efficient than they could be.

While I see a lot of effort was put into writing these tests, I think it makes no sense to merge those - the code has not been touched in a long time and is likely not even in a working state right now so we would test a state that is probably broken anyhow.

drakshakhan:
@bergmannf ,
Thank you for your feedback on the PR. After reviewing your comments, I agree that adding the tests right now doesn’t seem productive given the current state of the code. I’ll go ahead and close the PR as per your suggestion, and we can revisit this once the code is updated to handle pagination and efficiency improvements.
Appreciate your insights!

PR #723, opened 2025-04-15
OSD-29084: Tc pkg pko and git

karanjitsingh01:
Added test cases for promote/git and promote/pko package 

Overall coverage for git package is  54%. 
Overall coverage for pko package is  50%. 

Most of the functions were running git commands, like git checkout or git pull, these can't be mocked as the logic is inside the function.

we took the approach of executing real git commands inside a temporary directory (t.TempDir()) to simulate a realistic Git environment.
Approach for test: -
Create a temporary directory with t.TempDir(), Its a temporary directory so should be deleted after the completion of test
Initialize a Git repo using git init and run all required commands for testing in setup().
Call function in test.


Makdaam:
Hi, thanks for writing the tests. They cover the right cases, but at the same time they are testing a lot of environment specific dependencies (os executing the git commands). Could you refactor the app_interface code so that it doesn't call os.exec and allows for proper mocking of the git add/remote/checkout etc. function calls? The original code wasn't written in a way that makes it properly testable in separation.

Similarly the pko tests go through the whole pipeline calling appInterface and all the dependencies inside, which in turn calls os.exec git commands. I think it could be a mocked appInterface passed into the PromotePackage function call.

Makdaam:
Thanks for adding this abstraction, I haven't ran the code yet but the idea looks good for adding the tests without introducing major changes to the original code.

/ok-to-test

karanjitsingh01:
Hi @Makdaam , I have updated the package and tests as well, I have removed the tests from pko package as they call methods from git and saas. 
These methods are tested individually, hence I guess we can skip them here.

Makdaam:
Sorry, didn't notice the project OWNERS wasn't updated yet.
CC @clcollins 
Could you approve?

PR #727, opened 2025-04-18
Fix panic and other misc code issues for 'osdctl cluster support post'

joshbranham:
When running `osdctl cluster support post` with no flags this would panic. I fixed the panic in `setup()` and made a few other changes to simplify things and handle errors better.

### DIFF
--- a/cmd/cluster/support/post.go
+++ b/cmd/cluster/support/post.go
@@ -49,18 +49,18 @@ type Post struct {
 }
 
 type TemplateFile struct {
-	Severity       string             `json:"severity"`
-	Summary        string             `json:"summary"`
-	Log_type       string             `json:"log_type"`
-	Details        string             `json:"details"`
-	Detection_type cmv1.DetectionType `json:"detection_type"`
+	Severity      string             `json:"severity"`
+	Summary       string             `json:"summary"`
+	LogType       string             `json:"log_type"`
+	Details       string             `json:"details"`
+	DetectionType cmv1.DetectionType `json:"detection_type"`
 }
 
 var (
 	userParameterNames, userParameterValues []string
 )
 
-func newCmdpost() *cobra.Command {
+func newCmdPost() *cobra.Command {
 	p := &Post{}
 
 	postCmd := &cobra.Command{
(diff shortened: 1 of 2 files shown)


cmd/cluster/support/post.go
```diff
@@ -79,34 +79,24 @@ The cluster has a second failing ingress controller, which is not supported and
 		Args:              cobra.NoArgs,
 		DisableAutoGenTag: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if err := p.Run(p.ClusterID); err != nil {
-				return fmt.Errorf("error posting limited support reason: %w", err)
```
joshbranham, line 83: Extraneous error handling, this would log twice on top of `cobra`s builtin handling of failures in a `RunE` func def.

cmd/cluster/support/post.go
```diff
@@ -79,34 +79,24 @@ The cluster has a second failing ingress controller, which is not supported and
...
-	// Mark cluster-id as required
-	postCmd.MarkFlagRequired("cluster-id")
+	_ = postCmd.MarkFlagRequired("cluster-id")
 
 	return postCmd
 }
 
-func (p *Post) Init() error {
-	userParameterNames = []string{}
-	userParameterValues = []string{}
```
joshbranham, line 106: These are still global variables and need to be factored better but moved this into `setup()`.

cmd/cluster/support/post.go
```diff
@@ -125,15 +115,15 @@ func validateResolutionString(res string) error {
...
 	if p.Template != "" {
 		if p.Problem != "" || p.Resolution != "" || p.Misconfiguration != "" || p.Evidence != "" {
-			return fmt.Errorf("\nIf Template flag is present, --problem, --resolution, --misconfiguration and --evidence flags cannot be used")
+			return fmt.Errorf("\nIf --template flag is used, --problem, --resolution, --misconfiguration and --evidence flags cannot be used")
 		}
 	} else {
+		if p.Problem == "" || p.Resolution == "" || p.Misconfiguration == "" {
+			return fmt.Errorf("\nIn the absence of --template flag, --problem, --resolution and --misconfiguration flags are mandatory")
+		}
 		if err := validateResolutionString(p.Resolution); err != nil {
```
joshbranham, line 124: This was the source of the panic, we were trying to slice parse the `Resolution` flag before checking if we needed it. By moving the `if p.Problem == "" || p.Resolution == "" || p.Misconfiguration == ""` up before this, the user gets a much better help output when missing flags.

cmd/cluster/support/post.go
```diff
@@ -311,20 +303,19 @@ func (p *Post) parseUserParameters() {
...
 
-func (p *Post) readTemplate() *TemplateFile {
-	if p.Template == "" {
-		log.Fatalf("Template file is not provided. Use '-t' to fix this.")
-	} else {
-		templateObj, err := p.accessFile(p.Template)
-		if err != nil { //check the presence of this URL or file and also if this can be accessed
-			log.Fatal(err)
-		}
+func (p *Post) readTemplate() (*TemplateFile, error) {
```
joshbranham, line 306: This func was doing no error handling, so it has been added.

rafael-azevedo (approved):
/lgtm

PR #730, opened 2025-04-21
Refactors client connection to use ocm-common

iamkirkbater:
By refactoring the OCM Client Connection Builder to use ocm-common tooling we no longer have to maintain our own configuration loading, also we now get os keychain support "for free".

I tested this by building the binary and running a few read-only commands against some CICD clusters. Additional testing will be required to validate.

Additionally, did a `ripgrep` search for `ocm-sdk-go\"` and saw that all imports of this were referring to this as `sdk` (i.e.: `import sdk "github.com/openshift-online/ocm-sdk-go"` - so I checked that all instances of `rg sdk.Connection` were not attempting to create their own OCM connection but to reuse the main connection. All instances that I could find of `sdk.Connection` from the ripgrep were this type being used as a parameter to a function.


pkg/utils/ocm.go
```diff
@@ -165,146 +150,34 @@ func GenerateQuery(clusterIdentifier string) string {
...
-		return nil, errors.New(ocmConfigError)
+		return nil, fmt.Errorf("unable to load OCM config. %w", err)
 	}
 
-	connectionBuilder.Tokens(config.AccessToken, config.RefreshToken)
-
-	if config.URL == "" {
-		return nil, errors.New(ocmConfigError)
-	}
+	// TODO - Would be nice to put the osdctl version in here dynamically as the agent
```
joshbranham, line 173: You can get that from https://github.com/openshift/osdctl/blob/8aff4f8c7efa3f8eb972199281cc3d0b55b01dcf/pkg/utils/version.go#L24 which is set via an `ldflag` at build time

iamkirkbater: That was significantly easier than I thought it would be. Thanks!

joshbranham:
> I tested this by building the binary and running a few read-only commands against some CICD clusters. Additional testing will be required to validate.

Are you saying you have more testing to do or need reviewer to do so?

iamkirkbater:
Thanks for the reminder on this. I'll be testing this over my weekend on-call shift and I'll get back to you on Monday.

joshbranham (commented):
/lgtm

PR #731, opened 2025-04-23
Fix verify docs

joshbranham:
https://github.com/openshift/release/pull/63482 is attempting to make docs generation run in CI. It would fail as the make target was broken, and some docs needed updating.


typeid (approved):
/lgtm
/approve

PR #732, opened 2025-04-23
OSD-29210 - only allow supported instance types when running osdctl cluster resize X

dakotalongRH:
[OSD-29210](https://issues.redhat.com/browse/OSD-29210)
Update osdctl such that when specifying a machine-type during an infra or control-plane node resize, only supported machines are allowed.


cmd/cluster/resize/controlplane_node.go
```diff
@@ -94,6 +95,11 @@ func (o *controlPlane) New() error {
 		return err
 	}
 
+	supportedInstanceTypes := []string{"m5.4xlarge", "m5.8xlarge", "m5.12xlarge", "m5.16xlarge", "m5.24xlarge"}
```
joshbranham, line 98: What do you think about defining this as a `var` in `cmd/cluster/resize/cmd.go` with a map for `infra` and `control-plane`, then this validation at the top of `New()` as the first thing we check to fail as early as possible if the values are bad?

Sample map
```go
var supportedInstanceTypes = map[string][]string{
	"controlplane": {"m5.4xlarge", "m5.8xlarge", "m5.12xlarge", "m5.16xlarge", "m5.24xlarge"},
	"infra":        {"r5.4xlarge", "r5.8xlarge", "r5.12xlarge", "r5.16xlarge", "r5.24xlarge"},
}
```

joshbranham: And then add/extend the tests to show it failing validation when the input values are not valid :)

joshbranham (commented):
/lgtm
/approve

PR #741, opened 2025-05-09
OSD-29210: Add gcp instance types to cluster resize command

zmird-r:
Added everything from https://docs.redhat.com/en/documentation/openshift_dedicated/4/html/planning_your_environment/osd-limits-scalability#node-sizing-during-installation_osd-limits-scalability


joshbranham (commented):
/lgtm

PR #743, opened 2025-05-09
Ensure Go 1.23 Everywhere + Fix CVEs

joshbranham:
This ensures we are using Go 1.23 in CI and in the (unused?) Dockerfile.

Also bumped some dependencies to fix CVEs.


Makefile
```diff
@@ -75,5 +75,3 @@ test:
 
 lint:
 	golangci-lint run
-
-setup: install-hooks
```
joshbranham, line 79: Unused, from a previous PR

PR #745, opened 2025-05-12
Fix `servicelog list` flag handling

joshbranham:
There was an issue with how we were using flags for `osdctl servicelog list`, such that they weren't being parsed properly and therefore our filtering when calling the OCM API was off.

This fixes those issues, and a little bit of cleanup.


cmd/servicelog/list.go
```diff
@@ -45,15 +37,14 @@ osdctl servicelog list --cluster-id=my-cluster-id --all-messages --internal
...
 			return listServiceLogs(opts.clusterID, opts)
 		},
 	}
 
-	cmd.Flags().BoolP(AllMessagesFlag, AllMessagesShortFlag, opts.allMessages, "Toggle if we should see all of the messages or only SRE-P specific ones")
-	cmd.Flags().BoolP(InternalFlag, InternalShortFlag, opts.internal, "Toggle if we should see internal messages")
-	cmd.Flags().StringVar(&opts.clusterID, ListclusterIDFlag, "", "Internal Cluster identifier (required)")
-	cmd.MarkFlagRequired(ListclusterIDFlag)
+	cmd.Flags().BoolVarP(&opts.allMessages, "all-messages", "A", false, "Toggle if we should see all of the messages or only SRE-P specific ones")
+	cmd.Flags().BoolVarP(&opts.internal, "internal", "i", false, "Toggle if we should see internal messages")
```
joshbranham, line 45: The main bugfix is here, using `BoolVarP` instead of `BoolVar`, to ensure it sets the value in our options struct.

cmd/servicelog/list.go
```diff
@@ -13,14 +13,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const (
-	AllMessagesFlag      = "all-messages"
-	AllMessagesShortFlag = "A"
-	InternalFlag         = "internal"
-	InternalShortFlag    = "i"
-	ListclusterIDFlag    = "cluster-id"
-)
```
joshbranham, line 22: I found this distracting/hard to flop back and forth from where the flags are defined so I removed them for ease of reading.

cmd/servicelog/common.go
```diff
@@ -112,8 +112,8 @@ func FetchServiceLogs(clusterID string, allMessages bool, internalOnly bool) (*v
 
 func sendClusterLogsListRequest(ocmClient *sdk.Connection, cluster *cmv1.Cluster, allMessages bool, internalMessages bool) (*v1.ClustersClusterLogsListResponse, error) {
 	request := ocmClient.ServiceLogs().V1().Clusters().ClusterLogs().List().
-		Parameter("cluster_id", cluster.ID()).
-		Parameter("cluster_uuid", cluster.ExternalID()).
+		ClusterID(cluster.ID()).
+		ClusterUUID(cluster.ExternalID()).
```
joshbranham, line 116: These are just helpers in the SDK to make querying more explicit

PR #748, opened 2025-05-13
Fix the linting config and a selection of failures

joshbranham:
After reviewing our prow config I noticed that we are never actually running linting, which means this project has gotten really behind in linting failures. https://github.com/openshift/release/blob/cdb6d2d55b95d6b8e84ea547dc9ad9b6337a6614/ci-operator/config/openshift/osdctl/openshift-osdctl-master.yaml#L26-L30

I upgraded to `golangci-lint` v2, including the config file we use in `boilerplate`, and configured it to ignore any failures prior to the latest commit for the project.

The reason I did that is I spent 2 hours fixing failures and there are still so many. This way we can fix them in batches, while ensuring any new PRs start getting linting.

Once this merges I will open a PR on `openshift/release` to fix the lint target.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,4 +25,3 @@ osdctl
 
 # goReleaser 
 dist/
-bin/
--- a/cmd/account/mgmt/account-assign.go
+++ b/cmd/account/mgmt/account-assign.go
@@ -332,7 +332,6 @@ var (
 	ErrAwsAccountLimitExceeded = fmt.Errorf("ErrAwsAccountLimitExceeded")
 	ErrEmailAlreadyExist       = fmt.Errorf("ErrEmailAlreadyExist")
 	ErrAwsInternalFailure      = fmt.Errorf("ErrAwsInternalFailure")
-	ErrAwsTooManyRequests      = fmt.Errorf("ErrAwsTooManyRequests")
 	ErrAwsFailedCreateAccount  = fmt.Errorf("ErrAwsFailedCreateAccount")
 )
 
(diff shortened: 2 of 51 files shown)


pkg/envConfig/config.go
```diff
@@ -52,41 +51,19 @@ func LoadYaml(paramFilePath string) Config {
 	return config
 }
 
-func LoadPDConfig(paramFilePath string) PDConfig {
```
joshbranham, line 55: Unused

.gitignore
```diff
@@ -25,4 +25,3 @@ osdctl
 
 # goReleaser 
 dist/
-bin/
```
joshbranham, line 28: Duplicate

cmd/account/mgmt/account-assign.go
```diff
@@ -332,7 +332,6 @@ var (
 	ErrAwsAccountLimitExceeded = fmt.Errorf("ErrAwsAccountLimitExceeded")
 	ErrEmailAlreadyExist       = fmt.Errorf("ErrEmailAlreadyExist")
 	ErrAwsInternalFailure      = fmt.Errorf("ErrAwsInternalFailure")
-	ErrAwsTooManyRequests      = fmt.Errorf("ErrAwsTooManyRequests")
```
joshbranham, line 335: Unused

cmd/org/clusters.go
```diff
@@ -153,20 +154,14 @@ func formatClustersOutput(items []*accountsv1.Subscription) ([]byte, error) {
...
-		table.Flush()
+		err := table.Flush()
+		if err != nil {
+			return nil, err
+		}
 		return buf.Bytes(), nil
 	}
 }
 
-func printClusters(items []*accountsv1.Subscription) {
```
joshbranham, line 161: Unused

PR #750, opened 2025-05-14
OSD-30081: Fix issues with `osdctl promote saas`

joshbranham:
In https://github.com/openshift/osdctl/pull/723 a refactor was done to make testing easier for this package. However, it turns out that was not tested functionally and there were a few issues. This fixes those issues.

### DIFF
--- a/cmd/promote/git/git_cmd.go
+++ b/cmd/promote/git/git_cmd.go
@@ -22,50 +22,8 @@ func getBaseDir(exec iexec.IExec) (string, error) {
 			return
 		}
 
-		BaseDir = strings.TrimSpace(string(baseDirOutput))
+		BaseDir = strings.TrimSpace(baseDirOutput)
 	})
 
 	return BaseDir, baseDirErr
 }
-
-/*
-// Not in use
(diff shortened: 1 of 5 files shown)


cmd/promote/git/git_cmd.go
```diff
@@ -22,50 +22,8 @@ func getBaseDir(exec iexec.IExec) (string, error) {
...
 
-		BaseDir = strings.TrimSpace(string(baseDirOutput))
+		BaseDir = strings.TrimSpace(baseDirOutput)
 	})
 
 	return BaseDir, baseDirErr
 }
-
-/*
-// Not in use
```
typeid, line 32: hm

typeid (approved):
/lgtm
/approve

cmd/promote/git/git_cmd.go
joshbranham: Unsure why this was commented out as "not in use" versus just being removed.

joshbranham:
Btw, I ran this locally with the fixes and it produced a valid branch and pushed it 👍

PR #758, opened 2025-05-22
Refactors version check logic

iamkirkbater:
Additionally, ignores `-next` version differences for a better developer experience. Will still throw errors if you're not working from the same version without `-next` builds, or if you use `go build .` instead of `make build`.


fahlmant:
Is there a specific ticket for this? And can you explain the better developer experience? The version check very intentionally makes sure that any locally compiled changes are explicitly acked by the user.

PR #764, opened 2025-06-06
OSD-28506: feat(Add Handover Announcements to Cluster Context Command)

devppratik:
We are looking to migrate from the Handover Doc to a JIRA based Handover Board. This would allow to use JIRA queries to directly display related handover announcements at the time of login for the cluster based on ID, Customer Name, Version and Product.

This PR adds the functionality to query JIRA and print the related announcements for the cluster.

Related Issue [OSD-28505](https://issues.redhat.com//browse/OSD-28505)


chamalabey:
@devppratik I have started testing your changes and noticed that cluster context does not show the announcement for staging cluster `2ja1c3di93h0jk8mm0h12hg7jg83nap2` with the announcement https://issues.redhat.com/browse/SREPHOA-9 
Please have a look and let me know if I am missing something .

devppratik:
@joshbranham Sure I can add some test cases, but the jira.go file would need to be refactored to make it mockable and testable. Will take up in a separate PR

PR #766, opened 2025-06-13
OSD-30482: feat(Add initial command for creating handover announcements)

devppratik:
Adds a command to help create handover announcements directly via osdctl
Also adds test cases for the command.

The command expects both interactive input and flags.
Example of card created via the command - https://issues.redhat.com/browse/SREPHOA-18


cmd/jira/cmd.go
```diff
@@ -10,4 +15,19 @@ var Cmd = &cobra.Command{
...
 func init() {
 	Cmd.AddCommand(quickTaskCmd)
+	Cmd.AddCommand(createHandoverAnnouncmentCmd)
+
+	createHandoverAnnouncmentCmd.Flags().String("summary", "", "Enter Summary/Title for the Announcment")
+	createHandoverAnnouncmentCmd.Flags().String("description", "", "Enter Description for the Announcment")
+	createHandoverAnnouncmentCmd.Flags().String("products", "", "Comma-separated list of products (e.g. 'Product A,Product B')")
+	createHandoverAnnouncmentCmd.Flags().String("customer", "", "Customer name")
+	createHandoverAnnouncmentCmd.Flags().String("cluster", "", "Cluster ID")
+	createHandoverAnnouncmentCmd.Flags().String("version", "", "Affects version")
```
joshbranham, line 25: Is this just a nice to have or do you see this as easier/better than just using the Jira UI?

If you feel strongly about this command existing, how about making it interactive which would be much easier than having to `--thing foo --other-thing bar`?

devppratik: It is both interactive and also has the flags set up. Incase the user misses any flags, it would be prompted to user to input the same

devppratik: > Is this just a nice to have or do you see this as easier/better than just using the Jira UI? 
I would say this would be nice to have but not necessarily a requirement. It would be a bit easier to do via the CLI, the UI needs to navigate between different fields

go.mod
```diff
@@ -61,6 +61,7 @@ require (
 	golang.org/x/term v0.31.0
 	google.golang.org/api v0.220.0
 	google.golang.org/genproto v0.0.0-20250207221924-e9438ea467c6
+	gopkg.in/AlecAivazis/survey.v1 v1.8.8
```
joshbranham, line 64: This package has been deprecated, we should not be extending use of it https://github.com/AlecAivazis/survey/tree/v1.8.8

PR #767, opened 2025-06-13
chore(Add Test Cases for the utils/jira)

devppratik:
Adds test cases for utils/jira.go file
Also updates the file to make it mockable and easy to test

Follow up from https://github.com/openshift/osdctl/pull/764

JIRA https://issues.redhat.com/browse/OSD-28506


pkg/utils/mocks/mock_jira.go
```diff
@@ -0,0 +1,122 @@
+//go:generate mockgen -source=./pkg/utils/jira.go -destination=./mocks/pkg/utils/jira_mock.go -package=utils
```
joshbranham, line 1: This should be defined where the actual implementation is, not in the mocks file.

PR #769, opened 2025-06-27
Add more supported instance types for down scale

feichashao:
This PR adds r5.xlarge and r5.2xlarge to the infra node supported type, and adds m5.2xlarge as the supported control plane instance type.

The use case is mainly for down size. Eg, when SRE up size the infra node to get some room to investigate a high resource usage issue, after SRE fixes the issue, SRE will need to scale it back to the original size.

As per the document https://docs.redhat.com/en/documentation/red_hat_openshift_service_on_aws/4/html/prepare_your_environment/rosa-sts-aws-prereqs#rosa-aws-policy-provisioned_rosa-sts-aws-prereqs , the minimum size for control plane is m5.2xlarge, minimum for infra is r5.xlarge.

### DIFF
--- a/cmd/cluster/resize/cmd.go
+++ b/cmd/cluster/resize/cmd.go
@@ -6,6 +6,7 @@ import (
 
 var supportedInstanceTypes = map[string][]string{
 	"controlplane": {
+		"m5.2xlarge",
 		"m5.4xlarge",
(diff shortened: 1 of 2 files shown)


joshbranham (commented):
/lgtm
/approve

PR #772, opened 2025-07-08
SREP-1010: Added "osdctl dt" check for vault cli

drakshakhan:
This PR introduces a check to verify the availability of the Vault CLI before performing Vault operations. If the CLI is missing, an appropriate error message is returned to inform the user that Vault CLI is not installed. If the Vault CLI is available, it will display the current version of the Vault CLI being used, which will assist in debugging Vault-related issues.


iamkirkbater:
/approve

I'll give a chance for someone else to review as well.

typeid (approved):
/lgtm
/approve

PR #779, opened 2025-07-25
Add a RPM spec file

bergmannf:
This PR adds a spec file to the repository that can be used with a webhook in COPR to automatically build a new version of osdctl on tag pushes.

The spec file and shell-script are mostly for reference, as the shell-script actually needs to be added to the custom build in COPR to actually prepare the spec file to have the correct version information during buildtime.


joshbranham:
Maybe not related to what you're doing, but long term I would suggest using goreleaser in GitHub Actions to release the project on tags 👌

bergmannf:
I think the goreleaser route might be useful for Mac OS (and providing  binaries for whoever does not want to use COPR) - my main goal of using a COPR flow would be allowing users to use their normal DNF to keep our tools up-to-date without needing to do it themselves or some other extra tool.

bergmannf:
I've added a github action to trigger the COPR build now as well - I decided to keep the COPR_URL as a secret, so not everyone can just  trigger builds.
I can add the secret before the merge already.

PR #780, opened 2025-07-29
SREP-962: Add absolute timestamp support to fetching DT logs (--from and --to flags)

MateSaary:
This PR adds support for specification of absolute (datetime) timestamps for the `osdctl logs` command through the newly added `--from` and `--to` flags. 

They are to be used in conjunction with one-another, of course cannot be used with `--since`, and the timestamps are also validated to ensure `from` is not after `to`. The PR also bumps the "spf13/pflag" dependency (to add `time.Time` flag support).

### DIFF
--- a/cmd/cluster/context.go
+++ b/cmd/cluster/context.go
@@ -462,18 +462,20 @@ func (o *contextOptions) generateContextData() (*contextData, []error) {
 				data.DyntraceEnvURL = "Failed to fetch Dynatrace URL"
 			}
 			return
-		} else {
-			query, err := dynatrace.GetQuery(hcpCluster)
-			if err != nil {
-				errors = append(errors, fmt.Errorf("failed to build query for Dynatrace %v", err))
-			}
-			queryTxt := query.Build()
-			data.DyntraceEnvURL = hcpCluster.DynatraceURL
-			data.DyntraceLogsURL, err = dynatrace.GetLinkToWebConsole(hcpCluster.DynatraceURL, 10, queryTxt)
-			if err != nil {
-				errors = append(errors, fmt.Errorf("failed to get url: %v", err))
-			}
 		}
+		query, err := dynatrace.GetQuery(hcpCluster, time.Time{}, time.Time{}, 1) // passing nil from/to values to use --since behaviour
+		if err != nil {
+			errors = append(errors, fmt.Errorf("failed to build query for Dynatrace %v", err))
+			data.DyntraceEnvURL = fmt.Sprintf("Failed to build Dynatrace query: %v", err)
+			return
+		}
+		queryTxt := query.Build()
+		data.DyntraceEnvURL = hcpCluster.DynatraceURL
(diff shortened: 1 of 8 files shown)


cmd/cluster/context.go
```diff
@@ -463,13 +463,13 @@ func (o *contextOptions) generateContextData() (*contextData, []error) {
 			}
 			return
 		} else {
-			query, err := dynatrace.GetQuery(hcpCluster)
+			query, err := dynatrace.GetQuery(hcpCluster, time.Time{}, time.Time{}, 10) // passing nil from/to values to use --since behaviour
 			if err != nil {
 				errors = append(errors, fmt.Errorf("failed to build query for Dynatrace %v", err))
 			}
 			queryTxt := query.Build()
```
rolandmkunkel, line 470: If dynatrace.GetQuery were to fail, wouldn't this call result in a panic? Should we return early in that case?

cmd/dynatrace/dtquery.go
```diff
@@ -18,6 +19,17 @@ func (q *DTQuery) InitLogs(hours int) *DTQuery {
 	return q
 }
 
+func (q *DTQuery) InitLogsWithTimeRange(from time.Time, to time.Time) *DTQuery {
+	q.fragments = []string{}
+
+	fromStr := from.Format("2006-01-02T15:04:05Z")
```
rolandmkunkel, line 25: Would it make sense to define the format string only once as a constant and reference that?

cmd/dynatrace/logsCmd.go
```diff
@@ -100,13 +111,13 @@ func NewCmdLogs() *cobra.Command {
 	return logsCmd
 }
 
-func GetLinkToWebConsole(dtURL string, since int, finalQuery string) (string, error) {
+func GetLinkToWebConsole(dtURL string, from interface{}, to interface{}, finalQuery string) (string, error) {
```
rolandmkunkel, line 114: Can we also use string here instead? Just looking at the signature of this function, it is not clear that it only takes valid Dynatrace from and to values, mainly now() or some variation of that in our code. Having interface here might lead to passing in timestamps in time.Time format

cmd/cluster/context.go
MateSaary: Great point, I've added the below line so it will now return if there is any fails with the query.
```
data.DyntraceEnvURL = fmt.Sprintf("Failed to build Dynatrace query: %v", err)
```

MateSaary:
Thank you @rolandmkunkel for the review :slightly_smiling_face: I've made the suggested changes, let me know if there is anything further of concern. Thanks!

MateSaary:
It indeed was validated locally. `logCmd.go` as it stands doesn't have its testing counterpart from what I can see (maybe creating a card for this might make sense), but I have added a simple test case for the newly added `InitLogsWithTimeRange()` function in `dtquery.go` in the meantime.

PR #787, opened 2025-08-06
add dead link checking to osdctl servicelog post

eth1030:


cmd/servicelog/link_validator.go
```diff
@@ -0,0 +1,69 @@
...
+type LinkValidator struct {
+	timeout       time.Duration
+	skipLinkCheck bool
+	httpClient    *http.Client
+}
+
+// NewLinkValidator creates a new LinkValidator with default settings
+func NewLinkValidator(skipLinkCheck bool) *LinkValidator {
+	return &LinkValidator{
+		timeout:       time.Second * 5,
```
joshbranham, line 21: Make this a const, and reference the timeout in the http client

cmd/servicelog/link_validator.go
```diff
@@ -0,0 +1,69 @@
...
+// Check if URL is active
+func (lv *LinkValidator) checkURL(url string) error {
+	resp, err := lv.httpClient.Head(url)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode >= 400 {
+		return fmt.Errorf("HTTP %d", resp.StatusCode)
+	}
```
joshbranham, line 51: If we get a 403 unauthorized, do we want to count it as a dead link? Also, if we get a 500, we probably don't want to count it, and maybe retry.

joshbranham (commented):
Overall this looks great! Left a few comments and suggestions, let me know if you have questions.

pkg/linkValidator/linkValidator_test.go
```diff
@@ -0,0 +1,270 @@
...
+	server410 := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusGone)
+	}))
+	defer server410.Close()
+
+	// Create a test server for warning status (500 internal server error)
+	server500 := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusInternalServerError)
+	}))
+	defer server500.Close()
```
abyrne55, line 152: Great idea!

pkg/linkValidator/linkValidator.go
```diff
@@ -0,0 +1,82 @@
+package linkValidator
```
joshbranham, line 1: The canonical package name format in Go is snake case, so this should be `package link_validator` and then update imports with the new name. The folder structure should also be `pkg/link_validator`

pkg/linkValidator/linkValidator.go
```diff
@@ -0,0 +1,82 @@
...
+// NewLinkValidator creates a new LinkValidator with default settings
+func NewLinkValidator() *LinkValidator {
+	return &LinkValidator{
+		timeout:    Timeout,
+		httpClient: &http.Client{Timeout: Timeout},
+	}
+}
+
+// Extract URLs from service log
+func extractURLs(text string) []string {
```
joshbranham, line 36: No need to comment private functions, but if you still want to, the format should be `// extractURLs <text>`

pkg/linkValidator/linkValidator_test.go
```diff
@@ -0,0 +1,270 @@
...
+		})
+	}
+}
+
+func TestLinkValidator_ValidateLinks(t *testing.T) {
+	// Create a test server for valid URLs (200 OK)
+	serverOK := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer serverOK.Close()
```
joshbranham, line 134: There's a lot of duplication here. It doesn't matter too much, but one option here is to use a private function to generate test servers for a given http status. For example (pseudo code)
```go
fun newTestServer(status http.Status) ... { }
```

PR #788, opened 2025-08-11
https://issues.redhat.com/browse/SREP-861

dakotalongRH:
Add support running network-verifier in pod with a service account. 
Add a flag to disable sending a service log.

This code was generated using Cursor.  There will be a followup to refine

### DIFF
--- a/cmd/network/verification.go
+++ b/cmd/network/verification.go
@@ -102,6 +102,8 @@ type EgressVerification struct {
 	KubeConfig string
 	// Namespace is the Kubernetes namespace to run verification pods in
 	Namespace string
+	// NoServiceLog disables automatic service log sending on verification failures
+	NoServiceLog bool
 }
 
 func NewCmdValidateEgress() *cobra.Command {
@@ -125,6 +127,12 @@ func NewCmdValidateEgress() *cobra.Command {
      public subnets (in non-privatelink clusters), since they have an internet gateway and no NAT gateway.
   2. Pod mode (--pod-mode): runs verification as Kubernetes Jobs within the target cluster. This mode requires
      cluster admin access but provides more accurate results as it tests from within the actual cluster environment.
+     
+     Pod mode uses the following Kubernetes client configuration priority:
+     1. In-cluster configuration (when ServiceAccount token exists)
+     2. Backplane credentials (when --cluster-id is provided)
+     3. User-provided kubeconfig (when --kubeconfig is specified)
+     4. Default kubeconfig (from ~/.kube/config)
 
   Docs: https://docs.openshift.com/rosa/rosa_install_access_delete_clusters/rosa_getting_started_iam/rosa-aws-prereqs.html#osd-aws-privatelink-firewall-prerequisites_prerequisites`,
 		Example: `
@@ -147,8 +155,14 @@ func NewCmdValidateEgress() *cobra.Command {
   # Run in pod mode using Kubernetes jobs (requires cluster access)
   osdctl network verify-egress --cluster-id my-rosa-cluster --pod-mode
 
+  # Run in pod mode using ServiceAccount (when running inside a Kubernetes Pod)
+  osdctl network verify-egress --pod-mode --region us-east-1 --namespace my-namespace
+
   # Run in pod mode with custom namespace and kubeconfig
   osdctl network verify-egress --pod-mode --region us-east-1 --namespace my-namespace --kubeconfig ~/.kube/config
+
+  # Run network verification without sending service logs on failure
+  osdctl network verify-egress --cluster-id my-rosa-cluster --no-service-log
 
   # (Not recommended) Run against a specific VPC, without specifying cluster-id
   <export environment variables like AWS_ACCESS_KEY_ID or use aws configure>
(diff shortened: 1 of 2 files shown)


cmd/network/verification.go
```diff
@@ -179,6 +193,7 @@ func NewCmdValidateEgress() *cobra.Command {
 	validateEgressCmd.Flags().BoolVar(&e.PodMode, "pod-mode", false, "(optional) run verification using Kubernetes pods instead of cloud instances")
 	validateEgressCmd.Flags().StringVar(&e.KubeConfig, "kubeconfig", "", "(optional) path to kubeconfig file for pod mode (uses default kubeconfig if not specified)")
 	validateEgressCmd.Flags().StringVar(&e.Namespace, "namespace", "openshift-network-diagnostics", "(optional) Kubernetes namespace to run verification pods in")
+	validateEgressCmd.Flags().BoolVar(&e.NoServiceLog, "no-service-log", false, "(optional) disable automatic service log sending when verification fails")
```
joshbranham, line 196: I think calling this something like `--skip-service-log` makes more sense

cmd/network/verification.go
```diff
@@ -580,22 +601,29 @@ func (e *EgressVerification) setupForPodMode(ctx context.Context) (*onvKubeClien
 	var restConfig *rest.Config
 	var err error
 
-	// Prefer backplane credentials when cluster ID is available
-	if e.ClusterId != "" {
+	if _, err := os.Stat("/var/run/secrets/kubernetes.io/serviceaccount/token"); err == nil {
```
joshbranham, line 604: Make this path a constant, then you can reference it in your tests as well.

cmd/network/verification.go
```diff
@@ -257,17 +272,23 @@ func (e *EgressVerification) Run(ctx context.Context) {
...
-			postCmd := generateServiceLog(out, e.ClusterId)
-			blockedUrl := strings.Join(postCmd.TemplateParams, ",")
-			if (strings.Contains(blockedUrl, "deadmanssnitch") || strings.Contains(blockedUrl, "pagerduty")) && e.cluster.State() == "ready" {
-				fmt.Println("PagerDuty and/or DMS outgoing traffic is blocked, resulting in a loss of observability. As a result, Red Hat can no longer guarantee SLAs and the cluster should be put in limited support")
-				pCmd := lsupport.Post{Template: limitedSupportTemplate}
-				if err := pCmd.Run(e.ClusterId); err != nil {
-					fmt.Printf("failed to post limited support reason: %v", err)
+
+			// Only send service logs if not disabled by flag
+			if !e.NoServiceLog {
```
geowa4, line 277: Could this be `if e.NoServiceLog { fmt.Printlin; return }` so we don't need to nest the main code?

cmd/network/verification.go
```diff
@@ -580,22 +601,29 @@ func (e *EgressVerification) setupForPodMode(ctx context.Context) (*onvKubeClien
...
 	var err error
 
-	// Prefer backplane credentials when cluster ID is available
-	if e.ClusterId != "" {
+	if _, err := os.Stat("/var/run/secrets/kubernetes.io/serviceaccount/token"); err == nil {
+		restConfig, err = rest.InClusterConfig()
+		if err == nil {
+			e.log.Info(ctx, "Pod mode using in-cluster configuration with ServiceAccount")
+		} else {
+			e.log.Info(ctx, "ServiceAccount token found but in-cluster config failed, falling back to other methods")
```
joshbranham, line 609: I think the priority should be a bit different here:

1. If I pass `--kubeconfig`, I explicitly want to use that always
2. If I pass `--cluster-id`, I want to use backplane
3. If I don't pass `--kubeconfig` or `--cluster-id`, try in-cluster
4. If in cluster fails, fall back to default kubeconfig

I would suggest moving this decision logic into a simple function so it reads easier within `setupForPodMode`.

cmd/network/verification.go
```diff
@@ -125,6 +127,12 @@ func NewCmdValidateEgress() *cobra.Command {
      public subnets (in non-privatelink clusters), since they have an internet gateway and no NAT gateway.
   2. Pod mode (--pod-mode): runs verification as Kubernetes Jobs within the target cluster. This mode requires
      cluster admin access but provides more accurate results as it tests from within the actual cluster environment.
+     
+     Pod mode uses the following Kubernetes client configuration priority:
+     1. In-cluster configuration (when ServiceAccount token exists)
+     2. Backplane credentials (when --cluster-id is provided)
+     3. User-provided kubeconfig (when --kubeconfig is specified)
+     4. Default kubeconfig (from ~/.kube/config)
```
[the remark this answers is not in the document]
joshbranham: The docs are autogenerated with `make verify-docs` so no manual management.

cmd/network/verification.go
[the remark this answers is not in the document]
joshbranham: This is the canonical path for a serviceaccount token when you run the SDK within a Kubernetes cluster, which is our goal

joshbranham (commented):
Some of the logic is a bit hard to follow in a diff, but having it nested in it's own func helps so let's get this work wrapped.

/lgtm
/approve

PR #790, opened 2025-08-18
SREP-1138 - Print env vars for a cluster.

gvnnn:
Adds a subcommand - osdtcl cluster get-env-vars - to print environment variables for a cluster's name, ID/UUID, and HCP/HC/Klusterlet namespaces. The command defaults to plain text key: value format, but can print export-ready environment lines as well as JSON.


MateSaary (approved):
/lgtm
/approve

PR #791, opened 2025-08-21
SREP-1191: add SDN-to-OVN migration check to osdctl cluster context

eth1030:
This PR increases visibility into the progress or health of an SDN-to-OVN migration by adding a `Migration Status` row to the output of `osdctl cluster context -C $CLUSTER_ID`. It uses the OCM API to get cluster specific migration status.

### DIFF
--- a/cmd/cluster/context.go
+++ b/cmd/cluster/context.go
@@ -925,17 +925,14 @@ func (data *contextData) printClusterHeader(w io.Writer) {
 	fmt.Fprintln(w, strings.Repeat("=", len(clusterHeader)))
 }
 
-func printMigrationStatus(data *contextData, w io.Writer) {
-	var name string = "Migration Status"
+func printSDNtoOVNMigrationStatus(data *contextData, w io.Writer) {
+	var name string = "SDN to OVN Migration Status"
 	fmt.Fprintln(w, "\n"+delimiter+name)
-
-	if data.SdnToOvnMigration == nil {
-		fmt.Fprintln(w, "No active SDN to OVN migrations")
-		return
-	}
 
 	if data.SdnToOvnMigration != nil && data.MigrationStateValue == cmv1.ClusterMigrationStateValueInProgress {
 		fmt.Fprintln(w, "SDN to OVN migration is in progress")
+		return
 	}
 
+	fmt.Fprintln(w, "No active SDN to OVN migrations")
 }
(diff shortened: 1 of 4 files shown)


cmd/cluster/context.go
```diff
@@ -276,6 +280,9 @@ func (o *contextOptions) printLongOutput(data *contextData, w io.Writer) {
 
 	// Print User Banned Details
 	printUserBannedStatus(data, w)
+
+	// Print Migration Status
+	printMigrationStatus(data, w)
```
joshbranham, line 285: Let's make this more explicit, you could imagine other migrations in the future.
```suggestion
	// Print SDNtoOVN Migration Status
	printSDNtoOVNMigrationStatus(data, w)
```

cmd/cluster/context.go
```diff
@@ -896,3 +924,18 @@ func (data *contextData) printClusterHeader(w io.Writer) {
 	fmt.Fprintln(w, clusterHeader)
 	fmt.Fprintln(w, strings.Repeat("=", len(clusterHeader)))
 }
+
+func printMigrationStatus(data *contextData, w io.Writer) {
+	var name string = "Migration Status"
+	fmt.Fprintln(w, "\n"+delimiter+name)
+
+	if data.SdnToOvnMigration == nil {
+		fmt.Fprintln(w, "No active migrations")
```
joshbranham, line 933: Similar to above, `No active migration`, targeting this code to _just_ the SDN to OVN migration

pkg/utils/utils_test.go
```diff
@@ -167,4 +167,5 @@ func TestIsValidMatch(t *testing.T) {
 			t.Errorf("Test %d failed: isValidMatch() = %v; want %v", i, got, tt.shouldPass)
 		}
 	}
+
```
joshbranham, line 170:
```suggestion
```

cmd/cluster/context_test.go
```diff
@@ -592,3 +593,41 @@ func TestPrintUserBannedStatus(t *testing.T) {
 		})
 	}
 }
+func TestPrintMigrationStatus(t *testing.T) {
+	tests := []struct {
+		name           string
+		data           contextData
+		expectedOutput string
+	}{
+		{
```
joshbranham, line 602: I don't think you _have_ to do this here if you don't want to, but you could mock the OCM client API responses and then test that given certain responses, you print certain output. That would also help you validate that whole code flow without having an SDN to OVN migration currently in flight in prod/stage.

cmd/cluster/context.go
abyrne55: Can confirm that OCM team has more migration types in their future plans

joshbranham (commented):
/lgtm
/approve

PR #795, opened 2025-09-04
SREP-966: reimplement pull-secret email validation without backplane elevation

MateSaary:
This PR reimplements `osdctl cluster validate-pull-secret` to improve usability by eliminating the default requirement for backplane elevation.

  Key Changes

  Security Enhancement:
  - Primary method now uses backplane-api to execute the [get-pull-secret-email](https://github.com/openshift/managed-scripts/tree/main/scripts/security/get-pull-secret-email) managedscript
  - Removes default elevation requirement, improving security posture
  - Elevation only occurs as fallback when managedscript execution fails

  Improved User Experience:
  - `--cluster-id` is now optional when logged into a cluster (works with both methods)
  - `--reason` is now optional:
    - Not used if managedscript succeeds (no elevation needed)
    - User will be prompted if elevation fallback is required and a reason was not given
  - New `--elevate` flag forces traditional elevation method when desired

  Execution Flow:
  1. Attempts managedscript execution via backplane-api (no elevation)
  2. Falls back to elevation method if managedscript fails
  3. User prompted for elevation reason only when needed
  4. `--elevate` flag bypasses managedscript and uses elevation directly

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/cmd/cluster/validatepullsecret.go
+++ b/cmd/cluster/validatepullsecret.go
@@ -3,33 +3,30 @@ package cluster
 import (
 	"context"
 	"fmt"
-	"io"
 	"os"
 	"regexp"
-	"time"
 
 	"github.com/AlecAivazis/survey/v2"
 	v1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
-	backplaneapi "github.com/openshift/backplane-api/pkg/client"
 	bpapi "github.com/openshift/backplane-cli/pkg/backplaneapi"
 	bpconfig "github.com/openshift/backplane-cli/pkg/cli/config"
 	bputils "github.com/openshift/backplane-cli/pkg/utils"
 	"github.com/openshift/osdctl/cmd/servicelog"
+	"github.com/openshift/osdctl/pkg/backplane"
 	"github.com/openshift/osdctl/pkg/k8s"
 	"github.com/openshift/osdctl/pkg/utils"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


cmd/cluster/validatepullsecret.go
```diff
@@ -86,6 +133,147 @@ func (o *validatePullSecretOptions) run() error {
...
+
+	client, err := bpapi.DefaultClientUtils.MakeRawBackplaneAPIClient(bp.URL)
+	if err != nil {
+		return "", fmt.Errorf("failed to create backplane API client: %w", err), false
+	}
+
+	return o.runManagedJobWithClient(client)
+}
+
+func (o *validatePullSecretOptions) runManagedJobWithClient(client backplaneapi.ClientInterface) (email string, err error, sentSL bool) {
```
rolandmkunkel, line 150: do we ever return sentSL=true?

rolandmkunkel (approved):
lgtm

pkg/backplane/client.go
joshbranham, line 1: Can we add some unit tests for this new package?

cmd/cluster/validatepullsecret.go
```diff
@@ -55,6 +62,15 @@ func newValidatePullSecretOptions() *validatePullSecretOptions {
 }
 
 func (o *validatePullSecretOptions) run() error {
```
joshbranham, line 64: Would be good to have some test coverage here of at least your new path

PR #802, opened 2025-09-21
SREP-1831 - Fix watch command label

tnierman:
https://issues.redhat.com/browse/SREP-1831

---

Modifies the label used by the suggested `watch` command to print all (current and new) control-plane instances when watching control-plane scale-up/down processes.

### DIFF
--- a/cmd/cluster/resize/controlplane_node.go
+++ b/cmd/cluster/resize/controlplane_node.go
@@ -541,7 +541,7 @@ func promptGenerateResizeSL(clusterID string, newMachineType string) error {
 
 	fmt.Println("Service log sent successfully. Use the following command to track progress of the resize:")
 	fmt.Println()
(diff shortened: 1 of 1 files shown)


cmd/cluster/resize/controlplane_node.go
```diff
@@ -541,7 +541,7 @@ func promptGenerateResizeSL(clusterID string, newMachineType string) error {
 
 	fmt.Println("Service log sent successfully. Use the following command to track progress of the resize:")
 	fmt.Println()
-	fmt.Println(`watch -d 'oc get machines -n openshift-machine-api -l machine.openshift.io/cluster-api-machine-role=master && oc get nodes -l node-role.kubernetes.io/control-plane'`)
+	fmt.Println(`watch -d 'oc get machines -n openshift-machine-api -l machine.openshift.io/cluster-api-machine-role=master && oc get nodes -l node-role.kubernetes.io/master'`)
```
joshbranham, line 544: Didn't CPMS switch to "control-plane" at some point? Won't this not work for "newer" instances?

joshbranham: Gotcha! So the new nodes have _both_ labels? If so, this looks good to me.

/lgtm

PR #887, opened 2026-05-11
Disable self-upgrade and version check for packaged installation

gvnnn:
When osdctl is installed via COPR or Homebrew, the self-upgrade command would overwrite a file owned by the package manager and the version-check nag is unhelpful. This adds a build-time InstallMethod variable (set via ldflags) that disables both behaviors and directs users to their package manager instead.


MateSaary (approved):
/lgtm

PR #890, opened 2026-05-13
[SREP-4791]: Add sre-agent integration to osdctl AI commands

TheUndeadKing:
## Summary
  Integrates the sre-agent Python tool as a new `osdctl ai sre-agent` command with automatic installation, configuration, and validation capabilities.

   ### Files Added
  - `cmd/ai/sre_agent/sre_agent.go` - Main command and orchestration
  - `cmd/ai/sre_agent/sre_agent_install.go` - Installation logic
  - `cmd/ai/sre_agent/sre_agent_config.go` - Configuration management
  - `cmd/ai/sre_agent/helper.go` - Shared utilities (git, venv, prompts)

##

## Summary by CodeRabbit

* **New Features**
  * Added an "ai" command with a new "sre-agent" subcommand to run SRE incident investigations.
  * sre-agent supports --pd-url (required), --auto-execute (optional), and --output (optional), and can run in interactive or fully automated modes.
  * Interactive setup assists with locating or installing the agent environment and required SOP assets.

* **Documentation**
  * Added user docs and command reference for the ai command and sre-agent usage/examples.

### DIFF
--- a/cmd/ai/sre_agent/helper.go
+++ b/cmd/ai/sre_agent/helper.go
@@ -23,8 +23,11 @@ func copyRepository(sourcePath, destPath string) error {
 }
 
 // promptUserInput reads a line of user input from stdin
-func promptUserInput() string {
+func promptUserInput() (string, error) {
 	reader := bufio.NewReader(os.Stdin)
-	input, _ := reader.ReadString('\n')
-	return strings.ToLower(strings.TrimSpace(input))
+	input, err := reader.ReadString('\n')
+	if err != nil {
+		return "", fmt.Errorf("failed to read input: %w", err)
+	}
+	return strings.ToLower(strings.TrimSpace(input)), nil
 }
--- a/cmd/ai/sre_agent/helper.go
+++ b/cmd/ai/sre_agent/helper.go
@@ -1,0 +1,30 @@
+package sreagent
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// copyRepository copies a directory recursively
+func copyRepository(sourcePath, destPath string) error {
+	fmt.Fprintf(os.Stderr, "Copying repository to %s...\n", destPath)
+	cmd := exec.Command("cp", "-r", sourcePath, destPath)
+	cmd.Stdout = os.Stderr
+	cmd.Stderr = os.Stderr
+
+	if err := cmd.Run(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// promptUserInput reads a line of user input from stdin
+func promptUserInput() string {
+	reader := bufio.NewReader(os.Stdin)
+	input, _ := reader.ReadString('\n')
+	return strings.ToLower(strings.TrimSpace(input))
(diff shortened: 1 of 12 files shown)


cmd/ai/sre_agent/helper.go
```diff
@@ -0,0 +1,140 @@
...
+	if urlEnd == -1 {
+		return "", fmt.Errorf("failed to parse wheel URL end")
+	}
+
+	wheelURL := outputStr[urlStart : urlStart+urlEnd]
+	return wheelURL, nil
+}
+
+// createVenvAndInstall creates a Python virtual environment and installs sre-agent from GitLab release
+func createVenvAndInstall(venvPath string) error {
```
bergmannf: I think adding installation steps into osdctl is not a good idea.
We currently work at providing homebrew recipes for most of our tools or RPM packages for Fedora.
I see this is from gitlab - so would it be an option to package this up instead of making osdctl suddenly become an installer?
We already had https://github.com/openshift/backplane-tools/ which barely got maintained, so I would like to *not* have even more random places pretending to be installers.
At most it should check that it's installed and fail early if it's not.

cmd/ai/sre_agent/validate_sre_agent.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/openshift/osdctl/internal/utils"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+// validateSreAgent checks if sre-agent is installed
+func validateSreAgent(homeDir string) bool {
```
bergmannf: I guess this should just be replaced with a call to a `sre-agent` binary and if this doesn't succeed, prompt the user to install it and make the binary available?
This would then leave it up to the user, if they want to use venvs + a script, or a aliased docker container.

cmd/ai/sre_agent/validate_sre_agent_config.go
```diff
@@ -0,0 +1,102 @@
...
+	"path/filepath"
+
+	"github.com/openshift/osdctl/internal/utils"
+	"gopkg.in/yaml.v3"
+)
+
+// checkSreAgentConfig validates config.yaml and updates ops-sop path if needed
+func checkSreAgentConfig(homeDir string) bool {
+	baseDir := filepath.Join(homeDir, ".local/share/sre-agent")
+	configPath := filepath.Join(homeDir, ".config/sre-agent/config.yaml")
```
bergmannf: If we really depend on this - on Linux at least - we should not reinvent the wheel and use https://pkg.go.dev/github.com/adrg/xdg to use the XDG directories to find config and data files.

geowa4 (changes requested):
IMO this kind of thing shouldn't be in osdctl. It makes another layer that is brittle by its nature. If sre-agent changes at all, this breaks. It also doesn't solve logistics like installation or versioning so it's a fraught abstraction. However, if this orchestrated some kind of workflow, I'd be open to hearing an argument for it.

PR #904, opened 2026-05-29
SREP-347 & SREP-4896: factorizing code, fixing some potential race conditions in StreamLogs & making sure 'osdctl rhobs metrics' can work at a given time or with a time range

Nikokolas3270:
Follow up of this PR:
https://github.com/openshift/osdctl/pull/876

- Factorizing some code duplicate introduced by https://github.com/openshift/osdctl/pull/892
- Fixing some potential race conditions in `osdctl rhobs logs --follow`
- Making sure 'osdctl rhobs metrics' can work at a given time (new `--time` option) or with a time range (new `--start-time` and `--since` options)

## Summary by CodeRabbit

* **New Features**
  * Metrics: choose instant vs range evaluation with --time / --start-time / --end-time / --since / --step; metrics tool returns standardized results + count
  * Logs: clarified LogQL wording, expanded follow-mode exclusivity, and refined streaming behavior

* **Bug Fixes**
  * Cell command: now fails if either metrics or logs retrieval errors

* **Documentation**
  * Updated CLI help and docs for rhobs metrics and logs usage and flags


cmd/rhobs/requests.go
```diff
@@ -826,25 +1009,45 @@ func (q *RhobsFetcher) PrintLogs(lokiExpr string, startTime, endTime time.Time,
...
+	return nil
+}
+
+type streamLogsContext struct {
+	isLooping bool
+	webSocket *websocket.Conn
+}
+
+func (f *RhobsFetcher) StreamLogs(lokiExpr string, format LogsFormat, isPrintingTimeValue bool, fieldNames []string) error {
+	startTime := time.Now().Add(-5 * time.Minute)
```
dustman9000, line 1037: `startTime` is still shared across goroutines without mutex protection. `closeWebsocket` writes it (`startTime = time.Now()`) from the signal goroutine, while `createWebsocket` reads it from the main loop. Moving it into `streamLogsContext` and guarding it with the same mutex would close the race.

Nikokolas3270: Thanks for pointing that out.
`startTime` is now read and updated by the same goroutine, all the time.

PR #910, opened 2026-06-03
feat: dynamic vault OIDC callback port and in-process token capture

clcollins:
## Summary

- Read dynamically-assigned host port from `/tmp/vault_callback_port` (written by ocm-container's ports feature) and pass it as `callbackport` to `vault login -method=oidc`
- When re-authenticating inside a container, first try writing `~/.vault-token` normally; if that fails (read-only bind mount or atomic rename issues), fall back to capturing the token in-process via `VAULT_TOKEN` env var
- Enables multiple ocm-container instances to run vault OIDC authentication simultaneously without port conflicts
- Updates error messages to reference the ports feature instead of the deprecated `launch-opts` approach

## Companion PR

- ocm-container: https://github.com/openshift/ocm-container/pull/531

## Jira

[ROSAENG-39831](https://redhat.atlassian.net/browse/ROSAENG-39831)

## Test plan

- [x] Unit tests pass: `go test ./pkg/utils/ -v`
- [x] Inside ocm-container with existing valid `~/.vault-token`: `osdctl rhobs logs` uses it directly
- [x] Inside ocm-container with expired token: `osdctl rhobs logs` triggers OIDC re-auth
- [x] With `/tmp/vault_callback_port` present: uses dynamic callback port
- [x] Without `/tmp/vault_callback_port`: falls back to default port 8250
- [x] Token file write failure: falls back to in-process `VAULT_TOKEN` capture

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[ROSAENG-39831]: https://redhat.atlassian.net/browse/ROSAENG-39831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Refactor**
  * Improved Vault token setup: dynamic callback-port detection, more reliable container-mode login flow, and a fallback that captures tokens when container file writes fail.

* **Tests**
  * Reworked tests to target helper behavior: added checks for port reading and login-argument composition; removed higher-level integration tests for full token setup.

### DIFF
--- a/pkg/utils/vault.go
+++ b/pkg/utils/vault.go
@@ -2,11 +2,13 @@ package utils
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
 	"os/exec"
 	"strings"
+	"time"
 
 	ocmutils "github.com/openshift/ocm-container/pkg/utils"
 
@@ -17,6 +19,7 @@ import (
 const (
 	vaultCallbackPortFile = "/tmp/vault_callback_port"
 	defaultVaultOIDCPort  = "8250"
+	vaultLoginTimeout     = 5 * time.Minute
 )
 
 const (
@@ -106,11 +109,17 @@ func setupVaultToken(vaultAddr string) error {
 
 // setupVaultTokenLocal handles vault OIDC login outside of a container.
 func setupVaultTokenLocal() error {
-	loginCmd := exec.Command("vault", "login", "-method=oidc", "-no-print")
+	ctx, cancel := context.WithTimeout(context.Background(), vaultLoginTimeout)
+	defer cancel()
+
+	loginCmd := exec.CommandContext(ctx, "vault", "login", "-method=oidc", "-no-print")
 	loginCmd.Stdout = nil
 	loginCmd.Stderr = nil
 
 	if err := loginCmd.Run(); err != nil {
+		if ctx.Err() == context.DeadlineExceeded {
+			return fmt.Errorf("vault login timed out after %s", vaultLoginTimeout)
+		}
 		return fmt.Errorf("error running 'vault login': %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


clcollins:

🤖 Claude <claude@anthropic.com> commenting on behalf of @clcollins

pkg/utils/vault.go
```diff
@@ -62,51 +94,139 @@ func setupVaultToken(vaultAddr string) error {
...
+	loginCmd.Stderr = os.Stderr
+
+	if err := loginCmd.Run(); err == nil {
 		log.Infoln("Acquired vault token")
+		return nil
+	} else if ctx.Err() == context.DeadlineExceeded {
+		return fmt.Errorf("vault login timed out after %s", vaultLoginTimeout)
+	} else if !isTokenFileError(err) {
+		return fmt.Errorf("vault login failed: %v\n\n"+
+			"If authentication timed out or the callback failed:\n"+
```
dustman9000, line 188: Can you make this less ocm-container focused? Seems like it would be a general UX issue?

pkg/utils/vault.go
```diff
@@ -62,51 +94,139 @@ func setupVaultToken(vaultAddr string) error {
...
+	loginCmd.Stdout = &tokenBuf
+	loginCmd.Stderr = os.Stderr
+
+	if err := loginCmd.Run(); err != nil {
+		if ctx2.Err() == context.DeadlineExceeded {
+			return fmt.Errorf("vault login timed out after %s", vaultLoginTimeout)
+		}
+		return fmt.Errorf("vault login failed: %v\n\n"+
+			"If authentication timed out or the callback failed:\n"+
+			"  1. Ensure ocm-container has the vault port enabled in the ports feature\n"+
```
dustman9000, line 215: Same UX issue here, please make more generic.

clcollins: @dustman9000 this is an ocm-container only problem, and this function is not reachable unless you are in ocm-container.  Normal `osdctl` users outside of ocm-container won't ever see this message.

pkg/utils/vault.go
clcollins: @dustman9000 Same as the other one: this is an ocm-container only problem & this code isn't reachable by normal `osdctl` users outside of ocm-container.

PR #929, opened 2026-07-06
fix: support PSC cluster cleanup in break-glass 

RaphaelBut:
break-glass cleanup was not detecting GCP Private Service Connect (PSC) clusters, causing it to fall through to the local KUBECONFIG path instead of cleaning up jump pods on hive. 
Also adds --hive-ocm-url flag to the cleanup subcommand so it can connect to the correct hive shard when a non-default OCM environment was used during break-glass. ( needed for testing in stage or if we would actually want to breakglass in stage at some point )

https://redhat.atlassian.net/browse/ROSAENG-16042

## Summary by CodeRabbit

* **New Features**
  * Added support for cleaning up both PrivateLink and PSC clusters.
  * Added an option to point cleanup and access commands at a specific Hive OCM URL.

* **Bug Fixes**
  * Improved error messages with a helpful hint when Hive shard access or login fails.
  * Validates cluster IDs and Hive URL input before cleanup continues.

* **Tests**
  * Expanded cleanup coverage for PSC clusters and Hive URL validation.

### DIFF
--- a/cmd/cluster/access/cleanup.go
+++ b/cmd/cluster/access/cleanup.go
@@ -7,6 +7,7 @@ import (
 	fpath "path/filepath"
 	"strings"
 
+	sdk "github.com/openshift-online/ocm-sdk-go"
 	clustersmgmtv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	"github.com/openshift/osdctl/pkg/k8s"
 	osdctlutil "github.com/openshift/osdctl/pkg/utils"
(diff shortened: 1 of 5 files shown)


cmd/cluster/access/cleanup.go
```diff
@@ -106,26 +121,51 @@ func (c *cleanupAccessOptions) Run(cmd *cobra.Command) error {
 		return err
 	}
 	c.Println(fmt.Sprintf("Dropping access to cluster '%s'", cluster.Name()))
-	if cluster.AWS().PrivateLink() {
-		return c.dropPrivateLinkAccess(cluster)
+	isPscCluster := cluster.GCP().PrivateServiceConnect().ServiceAttachmentSubnet() != ""
+	if cluster.AWS().PrivateLink() || isPscCluster {
```
RaphaelBut, line 125: this was the missing part to do the cleanup
the rest of the PR adds the multi env hive-ocm-url support to test/make this possible in stage

PR #940, opened 2026-07-30
Fix flaky link_validator test by using .invalid TLD

samanthajayasinghe:
## Summary
- The `TestLinkValidator_ValidateLinks/message_with_network_error_URL` test was failing because the fake `.com` domain it relied on started resolving
- Switched the test URL from `http://this-domain-should-not-exist-12345.com` to `http://this-domain-should-not-exist-12345.invalid`, using the IANA-reserved `.invalid` TLD (RFC 2606) which is guaranteed to never resolve

## Test plan
- [x] `go test ./pkg/link_validator/ -v` passes
- [x] All other tests unaffected (`go test ./...`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Updated link validation test data to use a guaranteed-invalid domain for network error scenarios.

### DIFF
--- a/pkg/link_validator/link_validator_test.go
+++ b/pkg/link_validator/link_validator_test.go
@@ -185,7 +185,7 @@ func TestLinkValidator_ValidateLinks(t *testing.T) {
 		},
 		{
 			name:             "message with network error URL",
-			message:          "Check http://this-domain-should-not-exist-12345.com",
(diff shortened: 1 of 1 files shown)


geowa4 (approved):
/lgtm

geowa4:
Example.com was reserved and intended for this use case right?

PR #948, opened 2026-08-14
ROSAENG-65404: remove COPR from release automation

dustman9000:
While testing the [ROSAENG-65404](https://redhat.atlassian.net/browse/ROSAENG-65404) release automation end-to-end (#946, #947), the COPR trigger step failed with a `404` from the `COPR_URL` webhook secret.

Turns out osdctl doesn't actually need COPR:
- The maintainer's real release process has always been tag + `make release`, which only ever published GitHub release assets. COPR was a separate, parallel automation nobody was relying on day to day.
- `backplane-tools` installs osdctl straight from GitHub release tarballs (`t.Source.FetchLatestRelease()` + `sha256sum.txt`), not an RPM.
- No other repo in the `openshift` org references an osdctl RPM/spec (checked via code search).

Rather than fix a webhook nobody needs, this removes COPR entirely:
- drops the COPR trigger step from `release-on-version-bump.yaml`
- deletes `trigger_copr.yaml` (fired on manual tag pushes)
- deletes `hack/copr.sh` and the checked-in `hack/osdctl.spec`
- updates README/Makefile references

Jira: https://redhat.atlassian.net/browse/ROSAENG-65404

[ROSAENG-65404]: https://redhat.atlassian.net/browse/ROSAENG-65404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Documentation**
  * Updated release documentation to describe GitHub-based tagging and release publication.
  * Removed obsolete references to Fedora COPR builds, configuration, and packaging.

* **Chores**
  * Removed the automated COPR build workflow and related packaging scripts.
  * Simplified release workflow validation and comments.
  * Removed COPR-specific installation and upgrade guidance.


dustman9000:
Good catch on the dead `copr` case in `UpgradeInstruction()`. Removed it along with the now-unreachable `copr` test cases in `version_test.go` and `upgrade_test.go`. Pushed as `de262c0`.

RaphaelBut:
/hold 
i believe this is used by some folks :)
IIRC backplane-tools is or was to be put into Maintenance mode and we are looking into these alternative install methods?

Is there a need to remove it?

petrkotas:
We want to keep COPR, this was added for folks that want to rely on Fedora packaging and not backplane-tools pulling binary. If this is failing, we need to collaborate and fix the failure.

bergmannf:
Ah - this was my bad. When I initially wanted to add the COPR builds I tried it with github action - which at the time - was not supported on our orgs.
I must have left the old COPR_URL in the secrets and forgot to update it when we moved the build repository.
I've updated it now and tested the command from the action locally and it works now.

geowa4 (approved):
/lgtm

bergmannf:
If we merge #951 and the COPR builds works after there is no need to remove it.

PR #974, opened 2026-09-23
Remove hardcoded Jira config from triage harness

RaphaelBut:
The custom triage harness unconditionally injected Jira credentials and a Jira-specific sandbox policy, which broke /fs-triage on GitHub issues (JIRA_USER_EMAIL not set on the native dispatch path).

The upstream base harness at 7ab05564 already handles forge-specific credentials via provider-backed overlays (ADR 0065), selecting GitHub or Jira config based on event.source.system. The custom env overrides and sandbox policy were a pre-overlay approach that is no longer needed.

Drop the env block, the policy reference, and the policies/triage.yaml file. The harness now inherits the upstream base policy (filesystem and process restrictions only, no hardcoded network rules) and lets the provider layer handle Jira and GitHub egress conditionally.

## Summary by CodeRabbit

* **Configuration Changes**
  * Triage now selects GitHub or Jira configuration based on the event source and relies on the upstream base for forge credentials and network access.
  * Triage no longer applies its dedicated sandbox policy or passes Jira credentials and settings through its own runner configuration.


RaphaelBut:
@coderabbitai review

bergmannf:
Looks and sounds good to me, but I don't know enough about fullsend and how it works at this level of detail to drop a final LGTM on this.
I'll leave this up to someone more versed in the fullsend ecosystem, but I'm fine adding an LGTM once that's done.

